PR #800, opened 2015-01-07
Change Backbone.js 001 to Backbone.js 101

chrisdziewa:
I don't believe 001 was the intended in the original title.


hgducharme:
@eriktrautman You want to check this out? I'm not sure if you meant `001` when you typed this up, but I can't imagine you accidentally hit zero when you meant to hit one, because they're on opposite sides of the keyboard lol.

PR #815, opened 2015-01-09
Add Tommy Noe's Private Events solution

thomasjnoe:


hgducharme:
Your link is broken.

Change the `private_events` in the url to `private-events`

PR #822, opened 2015-01-11
Added solution to html/css

calvinbh:


hgducharme:
Thanks for contributing your project, but please put your solution in chronological order so I can merge this in

PR #836, opened 2015-01-12
Update project_web.md with solution

craftykate:


hgducharme:
You forgot to add a leading bracket in your markup link

PR #1069, opened 2015-02-19
Add my solution to Connect Four

pattisdr:


craftykate:
Please add your solution at the bottom, under the other submissions.

PR #1127, opened 2015-02-25
Adding my solutuon to student solutions

csavage1994:


craftykate:
Please add your solution at the bottom of the other students. Right above "add your solution above this line". Also, you'll need a space between your asterisk and your bracket.

PR #1135, opened 2015-02-27
Update project_js_jquery.md

MichellePepe:


craftykate:
Please add your solution above the line "Add your solution above this line!" You will also need an asterisk before your link. For more help you can watch this short video: http://bit.ly/odin-solutions

PR #1139, opened 2015-02-28
Fix typos and switch to Ruby 1.9 hash syntax

JulsFelic:


rails/form_basics.md
```diff
@@ -8,7 +8,7 @@ Up until now, we've been thinking about Models in Rails on sort of a one-off bas
...
 
-Most forms won't be that long or complicated for you, but it's useful to appreciate all the things you can (and one day will) do with them.  It's incredibly easy to make a basic form so the first thing we'll do is make sure you've got an intimate understanding of how forms are created in HTML and then how Rails offers you some helpers to make your life easier.  We'll cover the way data is structured and sent to the controller until you feel pretty comfortable with that.  Then a later lesson will deal with how to take that basic understanding and make forms handle some more firepower. 
+Most forms won't be that long or complicated for you, but it's useful to appreciate all the things you can (and one day will) do with them.  It's incredibly easy to make a basic form so the first thing we'll do is make sure you've got an intimate understanding of how forms are created in HTML and then how Rails offers you some helpers to make your life easier.  We'll cover the way data is structured and sent to the controller until you feel pretty comfortable with that.  Then a later lesson will deal with how to take that basic understanding and make forms handle some more firepower.
```
hgducharme: What was changed here? I can't find what was edited.

rails/form_basics.md
```diff
@@ -180,7 +180,7 @@ And the generated HTML is:
 
 Note that this helper nests the Article's attributes (the hard brackets in the `name` attribute should be the dead giveaway).
 
-The best part about `form_for` is that if you just pass it a model object like `@article` in the example above, Rails will check for you if the object has been saved yet.  If it's a new object, it will send the form to your `#create` action.  If the object has been saved before, so we know that we're **edit**ing an existing object, it will send the object to your `#update` action instead.  This is done by automatically generating the correct URL when the form is created.  Magic!
+The best part about `form_for` is that if you just pass it a model object like `@article` in the example above, Rails will check for you if the object has been saved yet.  If it's a new object, it will send the form to your `#create` action.  If the object has been saved before, so we know that we're **edit**ing an existing object, it will send the form to your `#update` action instead.  This is done by automatically generating the correct URL when the form is created.  Magic!
```
hgducharme: @craftykate Thoughts on this edit? You know more rails that I do

rails/form_basics.md
JulsFelic: I don't remember editing anything specific here

craftykate:
Offhand I wouldn't change the word "object" on line 183 to "form". The paragraph is talking about the object itself and it will send the object, not the form, to the `#update` action. The form doesn't get sent to the `#update` action, the object does.

craftykate:
As for the syntax, I'll have to take a closer look.

rails/form_basics.md
onebree: Went from `firepower. `, to `firepower.` -- notice the deleted space at the end.

onebree:
I agree with @craftykate -- the form may change based on the controller action, be it "new", "edit", etc. The paragraph should keep the word `object` as that is what is affected by the controller actions.

hgducharme:
So @JulsFelic, if you would like to fix that edit from `form` back to `object` then we will be able to merge this in. Thanks

JulsFelic:
No problem!

PR #1164, opened 2015-03-06
Update project_advanced_building_blocks.md

samfreeosk:
added my solutions


craftykate:
It looks like you added the same link (bubble sort) twice, and not the enumerables link.

craftykate:
I saw you added the bubble sort link in a following pull request so I'll close this request, but if you have an enumerable link to add you can just open a new pull request when you're ready.

Issue #493, opened 2014-08-29
string's inside '#' makes it  display as comment.


riahtire:
in code examples at http://www.theodinproject.com/ruby-on-rails/routing 
it seems like everything behind '#' has been treated as comment, even when the '#' is in a quoted string. 
this problem maybe also exists in other page.

onebree:
@N19270 -- @eriktrautman committed a change to their syntax highlighterm which in turn affects all pages. As you said, any page that had the issue you described is now fixed.

https://github.com/TheOdinProject/theodinproject/commit/32bfe58b4ec7d6d84685b6ba8ea8f9fda704e0ac

N19270:
:+1:

PR #1169, opened 2015-03-07
Rails/Routing - Fix syntax highlighting for Ruby code

onebree:
In response to issue #493

changed `language-ruby` to simply `ruby`, as suggested by Github-Flavored Markdown docs. In the document preview, it appears top work now. Prior to the commit, the code blocks were not highlighted correctly on the site, and not at all in Github.


hgducharme:
@eriktrautman What do you think? I won't be able to run this on my server so I'm not sure.

onebree:
@eriktrautman

eriktrautman:
The issue is that we're not actually using pure Github Flavored Markdown on the project, we're using RedCarpet for it (see `md` in https://github.com/TheOdinProject/theodinproject/blob/master/app/helpers/application_helper.rb) and it needs `language-ruby` to work.  I updated Prism.js, our syntax highlighter, to account for comments better https://github.com/TheOdinProject/theodinproject/commit/32bfe58b4ec7d6d84685b6ba8ea8f9fda704e0ac

onebree:
@eriktrautman Thank you for the clarification, and the fix as well. I did not look through the application itself for helpers. I now see that all `#` symbols are highlighted properly, and that your commit affects all pages.

PR #1182, opened 2015-03-10
Add my solution's links for the given projects

antoniosb:
I've just finnished these lesson, and I would be very thankfull if you could merge my solutions.
Any review observations are welcome!
Thanks in advance!


craftykate:
Thanks Antonio!

PR #1198, opened 2015-03-13
Add HTML/CSS project solutions

gb69010p:
- Add Gb69010p's easy and difficult project solution GitHub links
- Change HTTPS to HTTP for Chris Peers' HTMLPreview link


craftykate:
Am I missing something? Chris' preview link worked just fine has http instead of https. Plenty of people use http for their preview. Just wondering if there's something I don't know.

gb69010p:
If it was working for you it's probably fine, but for some reason I'm not able to view any users projects showing HTTPS. All I see is a white page that says "Loading..." and never goes anywhere, but for some reason HTTP worked. I found this while trying to validate that my link worked and noticed most people used HTTP which gave me no problems at all.

Must be something on my end then. Sorry about that! I have changed this back so that only my solution is added.

craftykate:
Thanks for clarifying - I was just wondering. I'd never seen it make a difference and I was curious.

PR #1243, opened 2015-03-19
Hello... Please add my solution for Rails project: Blogger

rickstewart:


hgducharme:
Hi rick, you placed your solution at the end of another student's solution. Please put yours on its own line and we will be able to merge this in

craftykate:
@rickstewart I think if you just add a space between your asterisk and your bracket it should work out. Without the space there, markdown doesn't know to make it a new list item.

rickstewart:
Hey Cathy, Hunter,
Sorry about that - snow shoveler's exhaustion is my excuse - and I'm sticking to that! I added a space between  my submission and the previous one.  I had not noticed there is a preview tab before now, else I would have probably caught my mistake.

Rick

PR #1251, opened 2015-03-20

omokoro:
Add my Cipher solution.


craftykate:
I think you put your caesar cipher at the end of the substring section. Can you please move it to the appropriate section of the page? Then I'll add it. Thanks!

omokoro:
My mistake.  I placed my link in the correct spot.  Thanks for the notice.

PR #1259, opened 2015-03-21

omokoro:
Added my substrings solution.


craftykate:
@omokoro - Can you show me where the code is that you're linking to? All I can find is an empty file. Thanks!

omokoro:
I guess I failed to push the last commit.  The link should be good now.  Thanks!

PR #1276, opened 2015-03-23
Remove dead links from Additional Resources

prashanthcr:
The link to the interactive CSS vocabulary tool doesn't work. I looked for it in the domain http://foolproof.me/posts/vvv/css-vocabulary, it seems to have been taken down.

LearnStreet shut down almost a year ago, and doesn't seem to be coming back.


craftykate:
@eriktrautman I'm merging this as they are, indeed, dead links, just a head's up if you want to find alternative resources.

PR #1298, opened 2015-03-25
Tested on different resolution.

jlampe1985:
Realized that the footer wasn't positioned properly so I removed solution.


craftykate:
Do you want to leave it in anyway? As an example for others? It doesn't have to be perfect. You can leave your solution in and work on it. Up to you, but it's totally okay if it's not pixel-perfect. My vote is to leave it in, but you tell me what you want to do.

jlampe1985:
OK you convened me. Please leave the solution in thanks!

craftykate:
Okay, good. =) Everyone here is learning, everyone is at different stages. I think you did a really good job, regardless of where your footer is exactly. I'll close this pull request. Keep going, you're doing great!

PR #1306, opened 2015-03-26
Another solution added

sleaz0id:


hgducharme:
Hi @sleaz0id, please put your solution at the end of the list and we'll be able to merge this in. Thanks

sleaz0id:
Thanks for response. Now it should be fine :smile: .

PR #1338, opened 2015-03-30
Update_project.md

Fully34:
Needed to fix case problems to make the preview work <link href="stylesheet.css ... , instead of <link href="Stylesheet.css ...  The preview was also missing one of the img's I put in for the same reason.  Used "Notification_dark.png" instead of "notification_dark.png".   Both are fixed, should be good to go!  Will use lowercase for every file and link from now on!


craftykate:
@Fully34 Good job, looks great!

PR #1367, opened 2015-04-06
added student solution to Mastermind

mixophrygian:
With a simplified Knuth-algorithm AI and bitchin' ASCII-style titles.


craftykate:
@mixophrygian The space after the asterisk before "Add your solution above this line" needs to be added back or it won't render correctly.  So change it to: `* Add your solution above this line!` and then I can merge. Thanks!

mixophrygian:
I think I fixed it?  Does it look right now?

craftykate:
Yep, perfect!

PR #1395, opened 2015-04-10
Add optional reading about freelancing

nhquiroz:


craftykate:
@eriktrautman Thoughts on this addition?

PR #1396, opened 2015-04-10
Add optional reading for front/back/fullstack comparison

nhquiroz:


craftykate:
@eriktrautman Thoughts?

PR #1399, opened 2015-04-10
Add optional reading about resumes

nhquiroz:
I think it's valuable info and this is the best place to put it.


craftykate:
@eriktrautman Thoughts?

PR #1413, opened 2015-04-14
Fix Student Solutions List

sami:
If you go to the page http://www.theodinproject.com/web-development-101/javascript-and-jquery you'll see the list of Student Solutions is broken. I think it is because, there was no empty line added between the paragraph and the list.


craftykate:
Thanks!

PR #1480, opened 2015-05-01
add steven zerneris solution

sjzerneri:


craftykate:
@sjzerneri It looks like you have your caesar cipher in the substrings section. Move your link and then I can add. Thanks!

PR #1500, opened 2015-05-07
Update project_html_css.md

NickFS:


craftykate:
@NickFS Your link needs to be formatted correctly. It's `[link text](http://actual/link)`. Note the parentheses around the link. If you need more help here's a short video on adding links to the odin project: http://bit.ly/odin-solutions.

PR #1505, opened 2015-05-07
Ruby 101 assignment suggestions

nhquiroz:
Hi! I was learning Ruby for the last 2 months and I think the assignment section can be improved (based on my own experience). 
My suggestions:

- *tryruby* doesn't add any value, like any other **try-x** thing if you want to learn properly. The Fundamental Ruby course from Jumpstart Lab cover the same topics (and more) in a project-based context, with better details and explanations.
- The *Fundamental Ruby* course has the *Ruby in 100 minutes* included.

### DIFF
--- a/web_development_101/ruby_basics.md
+++ b/web_development_101/ruby_basics.md
@@ -30,10 +30,9 @@ A final note -- you'll be learning a bunch of new terminology and concepts here 
 * What is a gem?
 
 ## Assignment:
-1. Go to [tryruby.org](http://tryruby.org) and do the quick exercises there to get your feet wet.  Don't worry about the stuff where they're reading from a file, that's getting deeper than you need to right now.
-2. Read through the [Ruby in 100 Minutes](http://tutorials.jumpstartlab.com/projects/ruby_in_100_minutes.html) project from Jumpstart Lab.  If you can't get IRB running, check out the [Installations Project](/web-development-101/installations), which you should have done already.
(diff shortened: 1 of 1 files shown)


craftykate:
@eriktrautman Thoughts on this?

eriktrautman:
I agree partially but not 100%.   

Try Ruby is there precisely because it's not stressing anyone out -- it's just a mechanism to gain some exposure and familiarity.  It's useful to have a slightly automated hand-holding introduction for newbies even if it's not teaching much.  Keep that. 

The changed link to Jumpstarts' day 1 is basically just saying to do Ruby in 100 minutes plus the event manager, which would be better as two separate steps.  So best would be to leave the 100 minutes as-is and add in the event manager right after Chris Pine's book.

muzfuz:
Speaking from personal experience - I found Try Ruby very useful.  The 101 course isn't a deep dive by any stretch, and coming at it from a "I know nothing about this" perspective, Try Ruby holds your hand in a way that is useful at this early stage.

nhquiroz:
> The changed link to Jumpstarts' day 1 is basically just saying to do Ruby in 100 minutes plus the event manager, which would be better as two separate steps. So best would be to leave the 100 minutes as-is and add in the event manager right after Chris Pine's book.

Ok @eriktrautman, I agree with that. I can't say the same about the Try Ruby thing but it's my very own personal view against that type of tutorials, it's ok if you leave it.

PR #1506, opened 2015-05-07
Add Neo Ruby Koans to the assignment

nhquiroz:
I didn't see the koans anywhere in the curriculum and think it's a must. It covers beginner and some intermediate/advanced topics with the **learn-by-debugging** approach and teaches TDD principles too. This seems a good place to put it.


craftykate:
@eriktrautman Thoughts?

eriktrautman:
I'd put the Koans below the main assignment in an `## (Optional) Extra Practice` section right about "test yourself".  The Koans can be overkill for some folks and useful for others.

nhquiroz:
Actually, I found the Ruby Koans to be easier than Test-First: it focuses on features of the language and "the ruby way", issues you can solve reading the docs and googling (and that'll save precious time of developing/debugging later), while Test-First it's more algorithmic oriented.

muzfuz:
I think I agree with @eriktrautman on this one.  Looking at it from the perspective of my own learning process, I would have been pretty overwhelmed by this stuff at this stage.

PR #1537, opened 2015-05-14
Add Computer Science in Modern Everyday Life link

nhquiroz:


muzfuz:
Interesting.  I certainly can see some value in these, and I think it probably points a direction toward how this section can be improved.  But I'm not sure if these particular resources are the best ones to go with.  What do @eriktrautman and @craftykate and @hgducharme think?

eriktrautman:
This would fit better in the last mini-course on finding a job.

nhquiroz:
I moved the CS concepts links and Google guide to this PR

https://github.com/TheOdinProject/curriculum/pull/1564

PR #1564, opened 2015-05-19
Add CS concepts link and Google guide

nhquiroz:


muzfuz:
That google resource is pretty neat!

PR #1618, opened 2015-05-26
Please add my solution to HTML/CSS google project

gerardjorgensen:


craftykate:
@gerardjorgensen You have a formatting issue with your link so it won't render properly. Looks like a double asterisk. Please fix it and I will merge. Thanks!

PR #1622, opened 2015-05-26
This are my solutions.

ponkbrown:


craftykate:
@ponkbrown You need to add your solution to the bottom of the list of other solutions. Please change and I can add. Thanks! (For full instructions on how to add a solution to odin project you can check out this video: http://bit.ly/odin-solutions)

PR #1645, opened 2015-05-29
Added my solution

magicmamba:
Solution for the first HTML/CSS project, recreating the google homepage.


magicmamba:
Solution for the first HTML/CSS project, recreating the google homepage.

PR #1651, opened 2015-05-30
add new youtube solution

dstodolny:


craftykate:
@dstodolny @muzfuz I fixed this one by hand, but the added link was **below** the line that says "Add your link above this line", not **above** it. Something to watch out for. =)

muzfuz:
@craftykate Whoops sorry.  I usually catch those, but I think my mind was elsewhere on this one.

PR #1662, opened 2015-06-01
add my solution

alexgh123:
add my solution


craftykate:
@alexgh123 Can you give your second link a name? Like this: `[View in browser](your link)`

alexgh123:
I think I just updated the second link name. Let me know if I didn't do it
right.

Thanks,
Alex

craftykate:
@alexgh123 That worked! I changed it to say "View in browser" to be consistent across all project pages, but you did it right. Let me know if you need github help with anything.

alexgh123:
Awesome, thanks!

PR #1702, opened 2015-06-10
Added Adrian Clark's solution to Student Solutions

AdrianClark:


AdrianClark:
I'm not sure where the two commits are coming from. Both should be the exact same fork though. All I added was my solution to the Google Homepage project.

PR #1892, opened 2015-07-06
Changed third link and text on JS Basics.

Hutbytheton:
The current link to JQuery's Javascript 101 returns a 404 error, so it would be nice to remove that link and add a new one or simply remove that 3rd part if it's not necessary. 

Here, I substituted the Mozilla JS Basics page, which a lot of people direct beginners to. It covers things a little differently than the original link, but provides a good review of the Codecademy course while showing some of the cool ways that you can use JS on the web. As a beginner, I've been itching to see some basic use of JS on a webpage and this site scratched that itch.

### DIFF
--- a/javascript/js_basics.md
+++ b/javascript/js_basics.md
@@ -47,9 +47,9 @@ This is a fairly long list of things to pay attention to but it should be mostly
 
 ## Your Assignment
 
-2. Read [Javascript is Super Sexy](http://javascriptissexy.com/javascript-is-super-sexy/), which is a good post for getting fired up to learn Javascript.
-1. Finish [the entire Codecademy Javascript track](http://www.codecademy.com/tracks/javascript) (you already did up to lesson 5).
-2. Glance through [Javascript 101 from learn.jquery.com](http://learn.jquery.com/javascript-101/) to fill in gaps in the Codecademy course.  The stuff on scopes and closures will be covered in the later section in more depth.
+1. Read [Javascript is Super Sexy](http://javascriptissexy.com/javascript-is-super-sexy/), which is a good post for getting fired up to learn Javascript.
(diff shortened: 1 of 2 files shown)


KevinMulhern:
Looks good to me, anybody else have any thoughts? @eriktrautman, @craftykate, @muzfuz

eriktrautman:
Yeah it’s frustrating that they killed Javascript 101 so this may be the best we can do…

                                                           
Erik S. Trautman
Founder,  Viking Code School
The premier online program for learning professional web development

muzfuz:
My only thought on this is that, seeing as NodeJS provides an OS runtime, we have JavaScript separated from the browser kind of dominating the scene right now.  So would it make sense to think about a lesson plan where learning JavaScript isn't just about DOM manipulation?

The only reason I mention this is that personally it took me a while to learn to separate between functions and objects while I'm reading pure JS, and frankly learning jQuery adds layers of complexity that makes it difficult for someone who is just learning to differentiate the two.

If I'm totally honest, at the beginning I didn't know or understand where jQuery started and ended.

So I guess what I'm saying is would it make sense to try to find materials or write up a lesson that involves learning the basics of JavaScript by running scripts within a Node environment?  I feel like I would be OK with telling people exactly how to set up node to run simple scripts and interface with the file system... and then you could essentially do the same things as you've been doing in Ruby.

Once the student learns that stuff, adding on the layers of "this is how you structure a simple HTTP server in Node" and "this is how you call a client side library like jQuery" becomes way more straightforward.  (At least in theory).

eriktrautman:
I’d love to see more deep JS knowledge in the curriculum.  Specifically, it would be interesting to pull together some Node resources.  I’d welcome it.

                                                           
Erik S. Trautman
Founder,  Viking Code School
The premier online program for learning professional web development

PR #2056, opened 2015-07-26
Added my soluton

adduran17:
I added my solution and fixed dchen71's solution showing the actual link instead of a hyperlink


adduran17:
just updated to current version

PR #2341, opened 2015-09-10
Updating/deleting broken links for Simon Kraus' solution

simonkrausgit:


simonkrausgit:
I also added some more solutions for the ruby projects

PR #2446, opened 2015-09-29
Update project_js_jquery.md

fcarlosdev:
Implementation of exercise Sketchpad.


kriox26:
Hi @fcarlosdev, I see that the solution you uploaded points to the `index.html` file, could update that to point to the root directory, that link would be: https://github.com/fcarlosdev/sketchpad
Also, If you want others to see your solution live, you can go to [this webpage](http://htmlpreview.github.io/) paste in the link to your `index.html` file and generate the live demonstration, and then add it here like the others. This is only optional, but please update the url of the link to lead to the root directory of the project

fcarlosdev:
Ok, I going do that.

PR #2460, opened 2015-10-02
Update project_html_css.md

akpark93:
Added my solution!


kriox26:
Hi @akpark93! The link to your solution leads to the `index.html` file, why don't you make it link to root directory of the project, that would be this one: https://github.com/akpark93/the_odin_project/tree/master/google-homepage
So that anybody can go and navigate to the file they want :+1:

akpark93:
Got it, will do!

PR #2558, opened 2015-10-16
Update project_positioning.md

nomankarim:


kriox26:
@nomankarim You've already uploaded this solution and it has been merged, here is the PR: #2562

PR #2589, opened 2015-10-21
Added my project solution

patmallee:
Added my own project solution on line 38 of the file


kriox26:
Hi @patmallee , the link for the online solution is pointing to the wrong file. You should use this url to generate it: https://github.com/patmallee/mintForm/blob/master/index.html . Please fix that and resubmit :+1:

patmallee:
Thanks! I've fixed the link and re-submitted.

Where would I have found the correct link? is there an easy way to get to
it?

kriox26:
Actually there is, head to http://htmlpreview.github.io/ and input the **full** url to your html file, which in this case is: https://github.com/patmallee/mintForm/blob/master/index.html . That will give you the url for previewing your project :+1:

PR #2644, opened 2015-10-26
Student solution of sketchpad added

johnsmalley:


kriox26:
Hi @johnsmalley , on the online solution, I don't see the grid where I should be painting. I think it has some problems loading the `.js` files. Do you see it working in the url you submitted?

johnsmalley:
Hi @kriox26 thanks for catching this as I didn't check Chrome using the htmlpreview.github.io however it works fine in Fire Fox. It also works in both browsers when working off of my hard drive. I'm pretty new at this so I'll have trouble shoot why this is only happening in Chrome when using htmlpreview.github.io

Below is the message that I get in Chrome using the Developer Tool under the Console section

?https://github.com/johnsmalley/sketchpad/blob/master/index.html:1 
Refused to execute script from 
'https://raw.githubusercontent.com/johnsmalley/sketchpad/master/javascript/jquery-2.1.4.min.js' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.

?https://github.com/johnsmalley/sketchpad/blob/master/index.html:1 
Refused to execute script from 
'https://raw.githubusercontent.com/johnsmalley/sketchpad/master/javascript/script.js' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.

kriox26:
Yeah, that's actually an issue of how chrome interprets the raw files that are on github, with `raw.githubusercontent` does not work, but with `rawgithub` does. Don't worry, dealing with this stuff is annoying

PR #2665, opened 2015-11-01
Git in the real world project

arialblack14:


kriox26:
Hi @arialblack14, it seems that you are trying to merge the merge in here the `theodinproject` repo, I'm closing this pr since is not valid, what where the steps you did to do this?? So we can prevent this in the future, maybe we have some unclear info in the Git in the real world project. :+1:

arialblack14:
I was just following this:
 http://www.theodinproject.com/ruby-programming/using-git-in-the-real-world?ref=lnav
I realized there was something wrong, as soon as i created a pull-request.

kriox26:
Great, you probably got confused with this: `Let's say you want to contribute to this website. How do you do that?` . That is actually refering to the odin project web app, not the curriculum. A few paragraphs after, when talking about remotes we can see that it mentions `TheOdinProject/theodinproject` instead of `TheOdinProject/curriculum`. We'll update the currículum to make it less confusing, since we've had a few other prs like this one

arialblack14:
You are absolutely right...! That did confuse me a bit.

kriox26:
Okay then, we'll fix it right away. Thanks for the feedback :+1:

kriox26:
We've added a solution, you can see it here #2669. It'll be live soon :+1:

PR #2698, opened 2015-11-07
Update project_js_jquery.md

ad-johnson:
Added my project into the list


kriox26:
Hi @ad-johnson , the link for the online solution is wrong. Use this one instead: `http://htmlpreview.github.io/?https://github.com/ad-johnson/griddle/blob/master/griddle.html` and resubmit please

PR #2785, opened 2015-11-22
Update project_ruby.md

fcarlosdev:


fcarlosdev:
Solution's to project learn ruby.

PR #3167, opened 2016-02-04
Ruby stock picker project solution

myjoytou:
Please add my solution to Ruby stock picker project


muzfuz:
@myjoytou I'm happy to merge - but I noticed you are using a `for` loop, which is considered un-Ruby like.  (It actually uses `Array#each` internally).  Perhaps you want to try to re-implement using a different iteration method?  It's totally up to you.

PR #3170, opened 2016-02-04
Update project_html_css.md with my solution

SawyerMerchant:


muzfuz:
Nice work.  Do you want to include this solution?

SawyerMerchant:
Yes I do. Thank you.

muzfuz:
Ok awesome, I'll add your solution just gotta do a bit of fiddling to make it happen now :+1: I will sort it out no probs - and in future just keep your issue open and we will get it sorted!

SawyerMerchant:
Thank you!

PR #3229, opened 2016-02-13
suggested grammatical edit for clarity

jeanmerlet:
My suggested fix may not be the best, but the sentence as-is is confusing. "a proc or a block" implies either a block or a proc, but the second phrase then states only if BOTH are supplied, which contradicts the either-or statement in the first phrase.


muzfuz:
That sounds better to me.  @eriktrautman pinging you since you wrote it.

PR #3733, opened 2016-05-04
Solution submission.

BryanGigs:
```
Thank you,
The Odin Project team.
```

Thank you!


cdouglass:
@BryanGigs Your solution works for me locally (and it looks great), but when I preview it on https://htmlpreview.github.io none of the jQuery code runs (`ReferenceError: $ is not defined`). Can you change the 'http' in line 4 of your HTML document to 'https' and see if that fixes it?

What I think is happening here: your HTML has `<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js"></script>` but that URL is a redirect to the HTTPS version `https://ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js`. Browsers are more restrictive about this kind of mismatch when you're viewing a site that's not hosted on your own computer.

cdouglass:
@BryanGigs Merged :)

BryanGigs:
Thanks for the recommendation Cathy.  I made the change you suggested and its working now!
Cheers,
Bryan

PR #4022, opened 2016-06-20
Add Project Javascript and jQuery Solution

lah4life:
```
Thank you,
The Odin Project team.
```


KevinMulhern:
Hey @lah4life, your solution is fine 😄. I was Just wondering why you update .gitignore every time?

lah4life:
Thanks for asking. Guess I'm still a little timid with regard to asking
questions. I use an Ubuntu machine that seems to create a shadow copy of
any file I update. It adds a ~ to the end of the filename. As you pointed
out to me on a previous pull request I submitted, those shadow copies get
pushed to my github repo whenever I commit. Only solution I have found is
to add *~ to my .gitignore so they don't get pushed to my repo. I then
remove it directly in github so my .gitignore matches. If you have a better
suggestion I would love it. I'm sure this is an amateur solution I'm using.

KevinMulhern:
Thanks for explaining it @lah4life, what text editor are you using?

lah4life:
Bluefish

KevinMulhern:
Interesting choice of editor 😄 I haven't heard of that one before. In the settings in bluefish there will be options for backing up files. Turning it off should stop the extra files being made

lah4life:
Awesome, appreciate that suggestion. What is your text editor of choice?
Still early enough for me that switching to a more well known option won't
mess with my mojo.

Thank you for your help!

KevinMulhern:
are you on our slack chat @lah4life, you can gets lots of help quickly in there when you encounter problems like this 😄 

The most popular editors at the moment are sublime text 3 and atom.io

PR #4302, opened 2016-07-29
Removed non-working link to old Coursera course

tutordelphia:
The link to the old Coursera course did not work and it does not appear (after a quick search) to be anywhere else on the Coursera site.


codyloyd:
Nice catch.

PR #4305, opened 2016-07-29
Update project_html_css.md

fcarlosdev:
```
Updating the reference of my solution


fcarlosdev:
Updating the reference of my solution.

PR #4513, opened 2016-08-23
fix minor typo

bwitka:
```
Thank you,
The Odin Project team.
```

Line 75: typo -> letter 'a' appears twice when it should only appear once.
Change:
"...is a very good explanation of what a a virtual machine is." 
to 
"...is a very good explanation of what a virtual machine is."


csrail:
A good eye for detail @bwitka.
Thanks for this!

PR #5081, opened 2016-11-02
Add Git Real to additional resourses

OlehSliusar:


KevinMulhern:
Hey @OlehSliusar, we want to keep the additional resources section of new lessons to 3 - 4 extra helpful resources to avoid link walls and to keep students from being overloaded. We are already at that limit here so we can replace one of the other resources with Git real if its better than an another one or leave it as it is. 

Imo It's not as good as the other resources there due to it being a paid course. But of course I would love to get more opinions on this before we make any final decisions.

Which of the resources do you think it should replace and why? @OlehSliusar 😄 

@csrail @codyloyd @CouchofTomato @leosoaivan @105ron 
can we get input from you guys too please 😄

OlehSliusar:
@KevinMulhern I didn't know about a restriction of 3 - 4 items. We can don't include git real thus TryGit, which is in lesson's assignments, suggest continuing learning git with Git Real. So, the student will be aware of it.

codyloyd:
I generally dislike Code School, and since there are plenty of free resources out there, I'd say leave this one out.

KevinMulhern:
@OlehSliusar no worries you couldn't have known, its something we decided on recently. I need to draw up a contributing guide that will outline all of these kind of things.

Sorry I didn't catch that last part, you think it should be in the assignments?

OlehSliusar:
@KevinMulhern No, I didn't mean it should be in assignments. 
TryGit is in assignments, and when you finish it, it suggests you to continue learning with Git Real.
screenshot from 2016-11-02 17-34-58

KevinMulhern:
ah gotcha, yeah I agree at least the student will be aware of it 😄

PR #5254, opened 2016-11-24
Change YouTube link to begin at the correct time.

ninoM:
Berkeley SAAS course YouTube link now starts at minute 36, when link is clicked, like it is instructed.


105ron:
Thanks @ninoM . Changes added to the page. :shipit:

PR #5287, opened 2016-11-28
Some grammar fixes and useless symbols removal

zmux:


105ron:
Thanks for the updated @zmux :)

PR #5747, opened 2017-02-04
format fix + new student solution link

janus0:
```
Thank you,
The Odin Project team.

```


105ron:
Hi @janus0,
If you include a space between `*` and the `[` at the start of the first line the bug your trying to fix with this commit won't happen. It's just an oddity in mark down language. Other than that all good 💯

PR #5803, opened 2017-02-12
Patty's Google Homepage Added

pattydi:
```
Thank you,
The Odin Project team.

```


pattydi:
Adding Google homepage solution.

PR #5836, opened 2017-02-17
Orlando's solution

orlandodan14:
Fibonacci and merge_sort solution

```
Thank you,
The Odin Project team.

```


105ron:
@orlandodan14 in the future we need to keep `* Add your solution below this line!` on the page. Your solution should go below this line, not removed and solution added on the line. Thanks, Rhys.

PR #5923, opened 2017-03-03
Fixed some typos, grammar, punctuation, and a link

TheDarkula:
The Pomodoro Technique "great article" link isn't working, so I updated it with a similar one from the same site.

### DIFF
--- a/web_development_101/gearing_up.md
+++ b/web_development_101/gearing_up.md
@@ -1,4 +1,4 @@
-# Gearing up for success
+# Gearing Up For Success
 Learning how to program takes work. Like any skill worth knowing it takes time to acquire and it can't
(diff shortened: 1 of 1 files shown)


OlehSliusar:
Are you sure coordinating conjunctions should be capitalized? (and, but, or, for, nor, etc.)

TheDarkula:
Yeah, it's part of a title, not "sentence case"

105ron:
I think that would depend on what style guide you are using for headings. I don't know if we need to be that particular with the curriculum. We would like to encourage more people to contribute to lessons like in the Ruby and JavaScript overhaul we are trying to complete at the moment. Picking on whether a conjunction in a title should be capitalized probably makes us look elitist and discourages new users from contributing. We have many users whose primary language isn't English, we would like to create an environment that welcomes content from them to improve the overall course. Just my 2 cents worth, Rhys.

OlehSliusar:
@105ron you're right

TheDarkula:
Fantastic. Thanks!

PR #5976, opened 2017-03-12
Add holdercp's solution for the difficult HTML and CSS project.

holdercp:
Fix previous submitter's link to Github repo.

```
Thank you,
The Odin Project team.

```


105ron:
Thanks for fixing that for us @holdercp

PR #6062, opened 2017-03-24
Add my solutions

theghall:
```
Thank you,
The Odin Project team.

```


arun1595:
@theghall Looks like you've added the gitl url rather than the repository url. Can you fix it and update the PR?

PR #6088, opened 2017-03-28
Orlando's solution

orlandodan14:
My Toy app

```
Thank you,
The Odin Project team.

```


arun1595:
@orlandodan14 I would suggest to try deploying the application to Heroku, to get comfortable with deployment.

PR #6238, opened 2017-04-17
Link dead - line 32

Blimeys:
[Node.js vs Ruby on Rails by Will Nathan](http://willnathan.com/nodejs-vs-ruby-on-rails) on line 32 is dead
replace with a similar article: 
http://blog.techmagic.co/nodejs-vs-ruby-on-rails-comparison-2017-which-is-the-best-for-web-development/


arun1595:
@Blimeys Thanks! 

@codyloyd What do you think?

PR #6401, opened 2017-05-08
add solution to project js/jquery and missing ) on javasscript basics…

mojotron:
… solution

```
Thank you,
The Odin Project team.

```


105ron:
Thanks for fixing that @mojotron .

PR #6457, opened 2017-05-17
Add Bolcik's solution to html/css project

SeanKohlbrenner:
```
Thank you,
The Odin Project team.

```


arun1595:
Good job with the google homepage clone :+1:  Try centering the position of the buttons below the search textbox.

PR #6489, opened 2017-05-22
Use Better Specs original url

lujanfernaud:
[betterspecs.org](http://www.betterspecs.org) is working again :smile:


arun1595:
@lujanfernaud Thanks for fixing it again ;)

PR #6730, opened 2017-06-23
Update assignment principles of good programming

anistor86:
Hi guys, the first link of the assignments is broken! It points to a 404 page not found!!


KevinMulhern:
Hi @anistor86, thank you for reporting this. This lesson will probably be removed from the curriculum soon. In the mean time we can remove that broken assignment entirely and just have one assignment in the lesson. Do you want to create a new PR with it removed?

anistor86:
Yep! No worries! I'll do one also for another dead link!
Thanks guys!

PR #6734, opened 2017-06-23
Added iArunava's Solution for Euler Problem 1

iArunava:
This is the solution for Project Euler's Problem 1


105ron:
Thanks for fixing the previous solution and adding the extra line in @iArunava.

PR #6932, opened 2017-07-25
fixed markup

Khanthulhu:
The wrong colons were used around <head> and it wasn't appearing on the page.

```
Thank you,
The Odin Project team.

```


arun1595:
@Khanthulhu Thanks for fixing it!

PR #6945, opened 2017-07-28
Finished projects!

AndyDeNike:
Thank you for your time and effort!


105ron:
Thank you for adding your solution :smile:

PR #7135, opened 2017-09-01
Add article about associations to resources

lujanfernaud:
This article explains very well Active Record associations.


mindovermiles262:
Nice Find :sunglasses: Thanks for adding this!

PR #7203, opened 2017-09-15
Add danhofer's ruby testing solutions

danhofer:
```
Thank you,
The Odin Project team.

```


mindovermiles262:
Looks good 👍 

Just a note about Implicit Returns in ruby: The result of the last evaluated line is returned in a function

def add num1, num2
  sum = num1 + num2
  sum
end

can be rewritten as

def add num1, num2
  num1 + num2
end

danhofer:
Thanks for looking through my code and thanks for the tips, Andy! I revised my solutions to incorporate what you taught me.

PR #7375, opened 2017-10-18
Update rspec resources

CouchofTomato:
The tutorials point rspec guide is a pretty good beginners resource so it should be included at least as an additional resource.


mindovermiles262:
👍

PR #7497, opened 2017-11-11
Add questions, tasks, additional reading

theghall:
@codyloyd Take a look.  Ethics may deserve it's own section, but see what you think.


mindovermiles262:
LGTM :+1: 

Thanks for adding these resources, @theghall

codyloyd:
IMO the ethics stuff doesn't _really_ apply to a11y.  If people were to come here looking for content on a11y they'd be looking more for "how to make my site accessible" and maybe a little bit on "why a11y is important".  The ethics stuff touches a tiny bit on that last point, but seems to be more about security issues/selling peoples info/serving ads/ etc.

theghall:
@codyloyd Sounds good.

JonathanYiv:
Closing for now unless discussion is reopened.

PR #7574, opened 2017-12-06
added solution to list for html/css exercise

gingermostly:
```
Thank you,
The Odin Project team.

```


JonathanYiv:
Thanks for adding your solution!
For future submissions, be careful to not leave a space for markdown links
like so
[link] (something.com)
doesn't work, but
`[link](something.com)`
will work!

PR #7581, opened 2017-12-09
My Solution For The HTML/CSS Exerise

slimane2018:
```
Thank you,
The Odin Project team.

```


JonathanYiv:
Looks like you already made an earlier PR with the project submission, so I'm going to close this one! Looking good.

PR #7605, opened 2017-12-21
add mention of 3rd party source for codecademy to assignment

tansaku:
was working with someone who might have been helped if this had been highlighted


leosoaivan:
Helpful. Thanks!

PR #7607, opened 2017-12-22
Add Jeremy-D mint form solution

Jeremy-D:
```
Thank you,
The Odin Project team.

```


Jeremy-D:
I've tried to contribute before, but I think this is the first time I did it right! Please let me know if there is anything else I need to do. Thanks!

105ron:
Well, you did your solution correctly but there's a lot of other changes to the file that shouldn't be there. There should be only one line changed with this -> `* [Jeremy-D's solution](https://github.com/Jeremy-D/Mint-form) - [Live](https://jeremy-d.github.io/Mint-form/)`

Jeremy-D:
hmm that's strange, i'll try to do the whole thing over again!

PR #7866, opened 2018-04-15
Added Tobias' solution

viviers:
```
Thank you,
The Odin Project team.

```


ryanford-frontend:
Looks like your links didn't render correctly. Make sure you provide the following format:
`[link text](https://path.to.page)` with no space between the `[]` and the `()`. Also be sure to include the `http://` at the start of your urls.

PR #7873, opened 2018-04-19
Fix broken "Dive into HTML5" link

brxck:


arun1595:
@brxck Thanks for the fix!

PR #8012, opened 2018-06-17
Added a note on Line 39

lerenah:
Line 39, references alternatives to using GitHub

Thank you,

The Odin Project team.

### DIFF
--- a/web_development_101/project_html_css.md
+++ b/web_development_101/project_html_css.md
@@ -36,6 +36,8 @@ If you are having trouble, refer to:
 
 Note: All Git commands need to be run from inside your project's folder (did you forget to `cd` into the `google-homepage` folder?)
 
+(Add another note here saying about how github recently lost 1/4 million users and add an option to use gitlab or bitbucket)
+
 *When you're building your project, you will probably end up doing several `git add` + `git commit` cycles before being ready to push it up to Github with `git push origin master`.*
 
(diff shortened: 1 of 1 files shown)


KevinMulhern:
Wow I wasn't aware Github had lost that amount of users so far 😮 

Do you have a source for that I could read? (not disputing its accuracy, just want to learn more 🙂)

I think this might fit better in the [Introduction to Git Lesson](https://www.theodinproject.com/courses/web-development-101/lessons/introduction-to-git)

Its the first time we expose users to Git properly and just seems like the more appropriate place for something like this to me.

Whats your thoughts?

lerenah:
I’ve been trying to hunt down the original link for it, maybe I should have included it in the note. But basically it has something to do with Microsoft buying github and a solid of number of users lost trust in the stability of github afterwards. It’s just a suggestion. 

I think the Introduction to Github link is quite effective! Thank you for the help!

Fate be changed 
Look inside 
Mend the bond
Broken by pride

KevinMulhern:
Thanks @lerenah, I think mentioning that there are alternatives for Github would be great 👍
 
But we would like to remain impartial to the Github situation. We'd prefer to wait and see what MS does with Github before taking any stances. I personally don't think MS now owning Github needs to be highlighted in any of our lessons.

But I'm open to hearing your thoughts on why we should highlight it 🙂

lerenah:
That is  a very sound stance!  Thank you for sharing :) !!

PR #8085, opened 2018-07-14
Change download link for Virtual Box

dvik1950:


dvik1950:
I've split the downloading process (Step 1) and the installation (Step 2) into separate steps as instructed. 

#### Step 2.1: Installing VirtualBox
```
The installation of VirtualBox is a very straight forward process. It doesn't require any technical knowledge and is the same as installing any other computer program on your Windows computer. Double-clicking the downloaded file is sufficient to start the installation process. Any additional options prompted by the installation are left for the user to decide (such as creating a desktop icon and so on). After the installation is finished (the progress bar might get stuck for a few minutes, just wait for it to finish) search for your newly installed Virtual Box program and run it.
```

PR #8236, opened 2018-08-27
rework installfest

mindovermiles262:

Thank you,

The Odin Project team.

### DIFF
--- a/web_development_101/installfest_prerequisites.md
+++ b/web_development_101/installfest_prerequisites.md
@@ -1,5 +1,5 @@
 ### Introduction
-Before we can install ruby and get ready to launch a rails application, we have to set up the operating system to be compatible. If you are using a `MacOS` or `Linux` machine already, you can safely skip this section. Otherwise, click on the dropdown of the method you would like to use below.
+Before we can install ruby and get ready to launch a rails application, we have to set up the operating system to be compatible. If you are using a `MacOS` or `Linux` already you can skip this section. Otherwise, click on the dropdown of the method you would like to use below.
 
 <details markdown="block">
 <summary class="dropDown-header">Linux Dual Boot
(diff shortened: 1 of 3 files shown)


web_development_101/installfest_prerequisites.md
```diff
@@ -201,25 +171,19 @@ After completing the last step, click the Create button. Your newly created virt
...
 
-Now you can start Ubuntu by right clicking on the icon in the menu and selecting Start then Normal Start.
+You can start the VM by right clicking on the icon in the menu and selecting Start then Normal Start.
 
-The next thing to do is Install Ubuntu. The process is very simple and most of the default options can be left like that including the Installation type which should be `Erase disk and install Ubuntu`. The setup will ask you to confirm this step because it thinks you're formatting your entire hard disk, but actually you're only formatting the newly created virtual hard disk, which doesn't have any data on it, and installing Ubuntu.
+The next thing to do is Install Xubuntu. The process is very simple and most of the default options can be left like that including the Installation type which should be `Erase disk and install Ubuntu`. You will need to confirm this step. It may sound dangerous, but the VM can only see the "Hard Drive" you created above in step 2.2 #2-5. This is the beauty of VMs - the ability to separate the physical space of your computer across many VMs.
```
I3uckwheat: > in step 2.2 #2-5. Thi

I'm unsure what you mean by `step 2.2 #2-5`

mindovermiles262: It's hard to tell from the diff but just before this line it reads:

    Step 2.2: Setting up Ubuntu

    Now that you have Virtual Box installed, double click the icon and you should see something like this:

    installed_vbox

    Click on the 'New' button to create a virtual operating system. Find your operating system in the dropdown menu (Linux/Ubuntu) and name it as you wish. Continue by pressing next and choose the following options in the next steps:

        1. Memory size - Should be about half of your computers maximum. For example, if you have 8GB of RAM memory, allocate 4GB to your virtual operating system.

        2. Hard disk - Create a virtual hard disk

        3. Hard disk file type - Choose the VDI (VirtualBox Disk Image) option

        4. Storage on physical hard disk - Dynamically allocated

        5. File location and size - We recommend at least 20GB for the virtual hard disk

    After completing the last step, click the Create button. Your newly created virtual OS should be in the menu now. Right click on it and go to Settings. Go to the Storage section and add the Ubuntu iso file you downloaded earlier:

    choose_disc_vbox

    After that, you can go to the System tab and change the amount of hardware the virtual operating system will be using. Generally 50% of RAM and at least 2 processors should be allocated to the virtual OS, but you can always change these settings based on your needs

Perhaps there is a different way I can phrase this? or do you have any suggestions?

I3uckwheat: Ahh, That makes more sense there. Looks good to me!

PR #8263, opened 2018-09-07
removing odd character that breaks the formatting

tansaku:
current problem:


tansaku:
@105ron you're welcome - it looks like it got there from @NathanPearson in https://github.com/TheOdinProject/curriculum/commit/40e498ce671f266b8933f513e5dadd3652c4a9f4

PR #8348, opened 2018-10-02
Edited Complete Ruby Link

uzorjchibuzor:
Hi The Odin Project Team,

I added a link to the complete book (Beginning Ruby) as the one on google books is missing several pages unless the reader buys the book.

Thank you

Chibuzor Efedigbue


CouchofTomato:
Fair question. You can personally use the material you have found if you like. That is your choice to make but we can't show support for that. The author himself has personally said he doesn't mind if people pirate it as the book isn't new anymore (although still good for the basics) but both morally and legally we can't link to it ourselves.

PR #8412, opened 2018-10-09
troubleshooting opening VSCode

jamiegitgit:
I ran into this problem while using the tutorial and added the solution to the instructions in case other users also encounter trouble here


I3uckwheat:
@jamiegitgit What error were you getting? A user shouldn't need to restart the entire machine, just the WSL instance. If it's not a windows error, it might be something else we can fix.

The problem with restarting the machine is the fact people can "lose" their spot in the file-system and have issues navigating back to the project directory on WSL, which happens without a restart. I feel this may cause more people to have this issue.

We should find a way to prevent this error from happening in the first place instead :)

PR #8466, opened 2018-10-15
fix broken link to ruby monk

### LINKED ISSUE #8460: Ruby Monk resources insecure error, then 502
https://www.theodinproject.com/courses/web-development-101/lessons/ruby-basics?ref=lnav

This lesson links to Ruby Monk which seems to be down, I have had a report of this before a while back and it is still down. Here is the page it links to: https://rubymonk.com/learning/books/1-ruby-primer

Possible solution: Find another resource to replace it?

dvik1950:
I replaced the current link with a link to the web archived version of the website, because Ruby Monk seems to be down for a few months now. This fixes #8460 .


mindovermiles262:
--- REVERTED ---

I'm not sure this will work as intended. The wayback portal doesn't provide any feedback on the actual ruby lessons.

Example: the tutorial says to put `self` into IRB

```
irb > self
=> main
irb >
```

But Wayback doesn't provide this feedback. 

Solution - we should find a new resource to substitute.  @dvik1950 have you come across one by chance?  If not I think REPL.IT might have something (they seem to be doing great things lately)

dvik1950:
@mindovermiles262 If I understood correctly, the entire Ruby section in Web 101 is going to be replaced pretty soon. This was meant to be a temporary fix until the [new lessons](https://github.com/TheOdinProject/curriculum/tree/master/new_course) are ready. The web.archive version of Ruby Monk might not work perfectly as the original website (students are still able to read through the lessons and see the solutions), but in my opinion it's better than a broken link.

leosoaivan:
Closing, see https://github.com/TheOdinProject/curriculum/issues/8460.

PR #8467, opened 2018-10-15
Add Jeremy-D's solution

Jeremy-D:

Thank you,

The Odin Project team.


Jeremy-D:
Solution is for the Rails Project: Forms section

PR #8474, opened 2018-10-16
Update rake to rails 5 syntax

brasscapon:


mindovermiles262:
Thanks for updating these, @brasscapon :smile_cat:

PR #8543, opened 2018-10-22
add updated video on queues and stacks

hoangtommy:
The existing 3min video on stacks and queue was lacking detail. Adding this video from the same youtube channel used in other lessons.


hoangtommy:
Including a video on recursion to supplement the additional resources on the recursion lesson

PR #8556, opened 2018-10-23
Victor Jin Solution for SQL Zoo Exercise

Viktorjin:
This solution is missing the Q10 answer for the last chapter Q10 question.  
Could you please guide us what is the standard answer for Q10?

Thank you,

The Odin Project team.


Javier-Machin:
Hi Victor,

You had a dot between the two parts of the link, I fixed it but try to be sure that the format is like the rest before sending the PR

Thanks.

PR #8580, opened 2018-10-26
Update heroku CLI MacOS installation instructions

CouchofTomato:
Heroku seem to have updated their CLI install instructions for Mac. This PR fixes that.


arun1595:
Thanks @CouchofTomato :+1:

PR #8649, opened 2018-11-04
Fixes directions

I3uckwheat:
gem install bundler command was missing


mindovermiles262:
Let's hold off on merging this:  Bundler is a prerequisite of rails, such then when you `gem install rails` bundler is installed as well.

The person that was having this in the Gitter room did not have ruby/rails installed via rbenv.

I3uckwheat:
Looks like this is a non-issue, closing.

PR #8650, opened 2018-11-04
Add the push command

alaajerbi:
After executing the git remote add command, we need to push the local branch to the repository. I added the missing command.


mindovermiles262:
Thanks for updating this @alaajerbi :+1: 

I just made a change to the line above that said "one simple command" to "two simple commands"

Thanks again :smile_cat:

PR #8678, opened 2018-11-07
Add my solution to HTML/CSS exercise

kody59527:

Thank you,

The Odin Project team.


Javier-Machin:
Hi @kody59527 !

I noticed that your `style.css` file isn't being loaded and the reason is that you have it linked with an absolute path, you should change it to a relative path (relative to index.html).
You can do that locally and then push the changes to your google homepage Github repository and it will be fixed 👍  .

PR #8848, opened 2018-12-01
fix typo config.ru to config.rb

hoangtommy:
Double checked that config.ru is indeed suppose to be config.rb. .ru is a russian java file haha


hoangtommy:
haha whoops, sorry for the false alarm! guess my google fu is still developing!

arun1595:
@hoangtommy  No worries.

PR #8856, opened 2018-12-02
Corrected grammatical error

prhps:
Hi, I noticed a potential grammatical error on line 213, where it says:

> The terminal will start it’s work, pausing to verify that you’re okay connection to GitHub

I proposing that this might make more sense: 

> The terminal will start it’s work, pausing to verify your connection to GitHub.

Not super important, but what do you think? 😄


105ron:
If we are going to correct the grammar, I think it should be:
> The terminal will start its work, pausing to verify your connection to GitHub.

prhps:
@105ron yeah can't believe I missed that!

105ron:
Merged, sorry I missed the updated commit.

PR #8859, opened 2018-12-02
Fix formatting of bulleted list

leila-alderman:
The first bulleted list under Motivation isn't showing up correctly on the TOP website. I think it needs a blank line above the list for it show correctly. 
Thanks!


arun1595:
@leila-alderman Thanks for the PR!

PR #8971, opened 2018-12-18
add word to fix typo

hoangtommy:
"that's because it uses the MiniTest testing framework instead of RSpec, which you have **been so** far." --> 
"that's because it uses the MiniTest testing framework instead of RSpec, which you have **been using** so far."


105ron:
If we are going to correct this, let's use `MiniTest` or `minitest`, not both. One must be correct, but I'm not sure which.

hoangtommy:
Rhys, I did a quick search. Seems like both are used. The original repo uses `minitest` so it might make sense for us to follow that.

105ron:
Let's change `MiniTest` to `minitest` then in the paragraph, then I'll merge it. :tada:

hoangtommy:
Done!

105ron:
Thank you :100:

PR #8976, opened 2018-12-19
Fix typo

PhilippSchweizer:


leila-alderman:
The original text here is correct. 
Original: "When stuck on a concept or project, taking a break to refresh and let your subconscious work on making connections more often than not **yields** a solution to your problem."

Changing "yields" to "yield" as you've suggested would be incorrect because the subject ("taking a break") is singular and therefore needs a singular verb ("yields").  

Thanks!

105ron:
I'll side with @leila-alderman on this one. But additional PR's for spelling and grammar are welcome. Thanks.

PR #9036, opened 2018-12-29
Remove apostrophe from it's

loumarven:
Should be "After the terminal finishes _its_ magic"


105ron:
Indeed it should be. And now it's. Thanks for your contribution. :smile:

PR #9067, opened 2019-01-03
Add JMG1995 console solution

JMG1995:

Thank you,

The Odin Project team.


codyloyd:
Couple of issues here:

First off, your markdown syntax changes will break things on our site... Our markdown parser uses a special syntax for some things. 

Second, it is generally a bad idea to bunch changes like this together because we can't accept anything if there's even one issue. 

And finally... Some of your changes are just kind of weird and unnecessary... Why would you change someone else's solution link?

I'm going to close this PR.. please submit another one with ONLY your solution in it.  In the future if you find a grammar mistake or something submit that separately from your solutions.

Don't be discouraged! This stuff does have a bit of a learning curve and the reason we ask you to do it is so that you learn the way your future employer will want you to do these things... So please do submit a corrected PR.

Issue #8347, opened 2018-10-02
Large Portions of Beginning Ruby Book on Google is not Available


uzorjchibuzor:
[https://books.google.com.pk/books?id=MiGpDAAAQBAJ&printsec=frontcover&hl=en#v=onepage&q=Chapter%208&f=false](https://books.google.com.pk/books?id=MiGpDAAAQBAJ&printsec=frontcover&hl=en#v=onepage&q=Chapter%208&f=false)
Large Portions of this books is hidden and asking for purchase before it can be viewed.

uzorjchibuzor:
I was able to download the pdf [here](http://www.allitebooks.in/beginning-ruby-3rd-edition/)

PR #9194, opened 2019-01-14
Update links to Beginning Ruby

leila-alderman:
I was running into the same problems as brought up in this issue: https://github.com/TheOdinProject/curriculum/issues/8347.

I changed the links on this lesson to point to the free pdf download of the Beginning Ruby book instead of the Google Books version, which is very limited.


leila-alderman:
Someone brought to my attention that this is an illegal pirated resource, so I'm closing this PR.

PR #9391, opened 2019-01-29
Add warning about eval() function

leila-alderman:
A common question about the calculator project is whether it's okay to use the eval() function. I added a warning to not use this problematic function in the project introduction.


mindovermiles262:
It's only the 28th day of the year and you've already won the Contributor of the Year Award 🎉  Thanks again for all your work

PR #9425, opened 2019-01-31
Possible typo in the Ruby event-manager project

SilasSpet:
I'm not a native but the sentence "we are able query for the representatives" doesn't sound right to me. I'm pretty sure it is missing a "to".


leosoaivan:
Thanks. Good catch!

PR #9429, opened 2019-01-31
Small writing changes in Ruby event-manager project

SilasSpet:
I changed the break down of the API URL since the link uses "https" and "key" instead of "http" and "apikey". This was probably left over from an old version of the URL.

I also removed a small typo with the double "the".

And last I changed "file#puts" to "IO#puts" with an explanation that the file object inherits from IO. I think it's better that way because throughout the page we have always used the class name followed by the referenced method.


mindovermiles262:
Great finds! Thank you!

PR #9910, opened 2019-03-14
Fixed markdown rendering in assignment section

leonardovega:
Fixed markdown rendering for nested lists in assignment section. When writing nested lists, they must be indented with 4 spaces, not 2.


105ron:
Thanks for the help 👍

PR #9973, opened 2019-03-19
Added solutions

N00bG1rl:
Added my solutions and deleted one link that was not working.


ryanford-frontend:
Thanks for spotting that link

PR #10004, opened 2019-03-22
add alternate version for sqlite3

svn1991:

Thank you,

The Odin Project team.


I3uckwheat:
Can you explain why we need this change?

svn1991:
The original sqlite3 version mentioned errored out when I was trying to set up the installation process on Windows. However, after googling and trial and error, the proposed sqlite3 worked. So I thought I should suggest it in case any other users faced this problem.

leila-alderman:
The Odin Project doesn't officially support development on Windows, so I don't think that this an update we need to make. 

If you'd like to work on creating a more complete documentation for how to do all of the installations on Windows along with corrections to any of the Linux commands that may differ for Windows, we're open to that. We'd like to avoid making those additions piecemeal, though, as that could lead to confusion in which directions to follow and whether or not we're actually able to support Windows.

codyloyd:
also.. why are we linking a fork that is less up-to-date than the original.

PR #10042, opened 2019-03-24
fixed broken link in Web Dev 101 Ruby Basics

benfowler04:
The link to Marc Gayle's article on Ruby was broken as HackHands.com has been shut down (redirects to Pluralsight). I've updated the link to point to the original article source at Smashing Magazine.


leila-alderman:
Thanks for fixing this!

PR #10224, opened 2019-04-10
Removes Broken Link from Gearing Up

Coop56:
I noticed that one of the links on the gearing up page goes to a Reddit post that has been deleted. This PR just removes that link from the page.


105ron:
Thank you for fixing that.

PR #10332, opened 2019-04-18
Added links to text guides as alternative to video

bycdiaz:


105ron:
Should we make these into list items?

bycdiaz:
Ah. That makes sense. Just added that.

PR #10344, opened 2019-04-19
Minor typo corrected

ruthmoog:
small correction on 1 word


leila-alderman:
Thanks for catching that!

PR #10362, opened 2019-04-21
remove unnecessary is

Rabidza:
Removed the word "is" to fix sentence.
From:
The following subsections cover **is** what you'll be using the Developer Tools for 95% of the time.

To:
The following subsections cover what you'll be using the Developer Tools for 95% of the time.


codyloyd:
BOOM

good catch.. thanks

PR #10399, opened 2019-04-24
removes ruby on rails tutor video

bycdiaz:
Video appears as unavailable on previous link: http://rubyonrailstutor.github.io/


105ron:
Why is there a blank line in the middle of this list?

Issue #10508, opened 2019-05-02
sqlite workaround is not needed anymore


I3uckwheat:
![image](https://user-images.githubusercontent.com/4411701/57108816-79585780-6cf9-11e9-97bf-3fbf08cb939f.png)

leosoaivan:
Resolved by f7715175ef20db79b5816b8ab3ceec45d592e561

PR #10510, opened 2019-05-03
Update curriculum text to match the latest version of Ruby

pwiggers:

Thank you,

The Odin Project team.


pwiggers:
Should also be the fix for issue #10508

PR #10540, opened 2019-05-06
Move command_line_basics to correct location.

leonardovega:
Moved the command-line basics lesson to it's new location.


leonardovega:
Np problem! :+1:

PR #10549, opened 2019-05-06
Add development instructions and explain better the messages loop

ste001:
I went through the tutorial, and decided to:
- Add a line to point 4, just to remember to the user to pass the messages array to the template, which wasn't really explained in the previous lessons or in this one.
- Add the deployment instructions for the last part. I checked on the Heroku website for the best practices.


codyloyd:
awesome. thanks so much <3

PR #10561, opened 2019-05-08
moves until loop info just below while loop info

bycdiaz:
Since the Until loop is being contrasted with the while loop, I figure this might be a better spot for this content.


leila-alderman:
I think this change makes a lot of sense. Thanks for making this suggestion @bycdiaz!

PR #10573, opened 2019-05-09
Link video on positioning to additional resources

mlbaquerizo:
This video made CSS positioning click for me after a couple of years of just throwing different position values at elements to see what works. It's so good, it might even make sense to put this in the Assignment section.


codyloyd:
nice find!

PR #10575, opened 2019-05-09
fix mistakes with markup

vladthelittleone:
tor does not recognize markdown inside html.

Снимок экрана 2019-05-10 в 0 03 05


leila-alderman:
Looks good. Thanks for contributing!

PR #10597, opened 2019-05-11
fix bug with markdown markup

vladthelittleone:
Снимок экрана 2019-05-11 в 18 46 48

Bug with markdown list in text.


leila-alderman:
@vladthelittleone Thanks for bringing this formatting issue to our attention! The solution to this problem, though, isn't related to the symbols used for creating bullet points but rather that there needs to be a blank line prior to the start of a bulleted list for it to display correctly. I've gone ahead and made that change in a separate commit, and it appears to be displaying correctly now. 

image

Thanks again for contributing to make TOP better!

PR #10616, opened 2019-05-13

zachmmeyer:
Just a little more user friendly description of what appears to be a semi uncommon issue with adding the public ssh to heroku.


105ron:
Thank you for the added information.

PR #10629, opened 2019-05-14
Update #before_filter to #before_action

suretrust:
Since the #before_filter syntax has been deprecated, I updated accordingly.


leila-alderman:
Thanks for making this update and for contributing!

PR #10634, opened 2019-05-14
Change sections to chapters

suretrust:
Since each section is defined as a chapter on the rails guides webpage as shown on the right sidebar of the image attached, I thought it's better understood when they are regarded as chapters.

rails guides

### DIFF
--- a/rails_programming/advanced_forms_and_activerecord/active_record_queries.md
+++ b/rails_programming/advanced_forms_and_activerecord/active_record_queries.md
@@ -212,16 +212,16 @@ Sometimes, you just can't get ActiveRecord to do what you want it to.  In that c
 <div class="lesson-content__panel" markdown="1">
 
 ### Querying Basics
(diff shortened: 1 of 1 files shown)


leila-alderman:
Hey @suretrust -- I'm curious to know whether you actually found it confusing that our lesson said "sections" instead of "chapters". While the Rails Guides pages clearly use "Chapters" at the top of the Table of Contents, I find that to be a rather unusual and not intuitive term in this context. If folks are actually finding our use of "section" to be confusing, though, we can consider switching it over.

suretrust:
@leila-alderman I didn't find it confusing at first. It became confusing when I got to 'Advanced Querying'. Chapters were used instead of sections. Then, I checked rails guides and realized they used chapters instead of sections. 

I think it'd be nice to have a uniform reference, choosing either chapters or sections all through.

leila-alderman:
@suretrust Thanks for your feedback! That's really helpful to know. 

You're absolutely right that we should be consistent in the terminology that we use throughout. Based on what you've said, though, I think that it's better to change everything over to using "section" instead of "chapter" for now. I've gone ahead and updated the lessons that used "chapter", which should cut down on confusion now. 

Thanks again for bringing this up! I'm going to close this PR for now since I went the "section" route, but please let me know if you have other questions or comments.

PR #10678, opened 2019-05-19
Fix typo in project Etch-a-Sketch

chrisjwelly:
A minor typo I found in Project Etch-a-Sketch. Although "verses" and "versus" are homophones, I believe the intended word here should be "versus"


105ron:
I agree. Thanks for the correction.

PR #10719, opened 2019-05-22
Add a tip for those doing JS exercises in Web Dev 101

chrisjwelly:
When I first did the exercises, I was frustrated by the fact that I was unable to test my JavaScript functions easily in the terminal. I even searched for a JS interpreter online before realising that I could do what I propose in this pull request. The Jasmine framework allowed me to somewhat see what the output of my wrong function was and what it was supposed to be, but often times the terminal gives many irrelevant information that I am not concerned with when using Jasmine, not to mention that I have to write tests on my own if I actually want to try inputting values that is not in the scope of the tests.

I believe many users would also like an easy way to test their functions for debugging purposes. Assuming they did this course sequentially, they should have done something similar to this in the console-only Rock, Paper, Scissors project a few lessons prior. However, it is also possible that they could have forgotten that they could do this, or there could be users who skipped the projects entirely.

Hence, I added this small tip section under the exercises to give users some direction.

### DIFF
--- a/web_development_101/javascript_basics/fundamentals-4.md
+++ b/web_development_101/javascript_basics/fundamentals-4.md
@@ -35,3 +35,19 @@ Complete the following exercises:
 - sumAll
 - leapYears
 - tempConversion
+
+Tip: To help you debug your code, it may be helpful to link your `.js` files to a HTML page and test the functions you have defined on the console.
+
+For example:
+```html
+<!DOCTYPE html>
+<head>
+  <meta charset="UTF-8">
+  <title>A page where I can test my code!</title>
+</head>
+<body>
+  <h1>Open your console to test out your functions!</h1>
+  <!-- Note: Below is the important part where you are linking your code to the HTML page -->
+  <script src="filename.js"></script>
+</body>
+```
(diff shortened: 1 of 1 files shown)


105ron:
I have always used repl.it and used their online nodejs IDE for things like this. Why don't you try that and maybe we can recommend it to students if you agree?

chrisjwelly:
> 
> 
> I have always used repl.it and used their online nodejs IDE for things like this. Why don't you try that and maybe we can recommend it to students if you agree?

I totally forgot about repl.it! I will give it a shot and update this pull request as necessary.

chrisjwelly:
@105ron I have tried repl.it for the exercises and I think it is a nice environment to test out my functions! We can recommend it to the students. 
I would still like to include my original proposal as an alternative recommendation, especially for those who have cloned the exercises and would like to work offline. But I have omitted most details so that the focus will be more on the easier method of using repl.it.

Additionally, I just realised that in [Fundamentals Part 2](https://www.theodinproject.com/courses/web-development-101/lessons/fundamentals-part-2), it says: "In general we believe that it’s best to work on programming on your own computer rather than in an online environment but we’ll get to that soon enough."

I think the current alternative recommendation is in line with the lesson's belief to work on our own computer rather than in an online environment. Are there alternative and simpler ways to test our functions offline? I considered `node filename.js` in the terminal, but it seems to only print out the `console.log`s.

codyloyd:
i think this kind of defeats the purpose of using and running Jasmine... you can just console.log stuff in your JS files anyway.

PR #10734, opened 2019-05-24
Update fundamentals-2.md

dsmchen:
Clarify instructions.


ryanford-frontend:
Reading it as it is now, I think the intention was to say "there are benefits to signing up, but it's not required for TOP"

leila-alderman:
I worked on making this phrasing a bit better overall. 

@dsmchen Please let me know if this looks better to you.

dsmchen:
Sorry, my first comment should've been:

Edit instructions as viewing the lessons without logging in / signing up shows a blank page.

ryanford-frontend:
Just tested with same results. I sent them a notice, but in the mean time we should definitely implement @dsmchen 's rewording so they can access the problems.

leila-alderman:
Thanks for the clarification @dsmchen and for confirming this @ryanford-frontend! I changed the phrasing to make it clear that signing in is required for the lessons and merged in the change.

PR #10780, opened 2019-05-27
Updates text in assignments area: terminal & AWS

bycdiaz:
The lesson was inviting people to skip ahead for installations. Updated text to reflect the fact that installations happen before this lesson.

Removed reference to AWS. Since learners should already have linux installed (or using MacOS), AWS shouldn't be necessary.

Added a step that shows learners how to open a terminal and try a basic command. This isn't explicitly mentioned at the start of the Conquering the Command Line book.


105ron:
It should be Windows subsystem? Then there shouldn't be a line between point 1 and two in the ordered list, and the two line breaks before the list only needs one line break.

105ron:
Thank you.

PR #10819, opened 2019-06-01
Update project_media.md

cgrossi:
Add cgrossi's solution

Thank you,

The Odin Project team.


cgrossi:
Added Carmine's solution

PR #10821, opened 2019-06-01
Update legislators_by_zipcode method

keskiviikko:
Line 952: def legislators_by_zipcode(zip) -> def legislators_by_zipcode(zipcode)

I noticed that "zip" returned the string under rescue instead of the legislators (for all of them). "zipcode" returns each legislator as I suppose it should.

(Great exercise btw!)


keskiviikko:
Never mind, I just noticed my mistake. Should pay more attention.

PR #10825, opened 2019-06-01
Updates project_your_first_rails_app.md

fernandamarr:
Changes "ours" to "yours" on Step 1.3


105ron:
If we change ours to your should we then should we also change we to you on the same line?

leila-alderman:
@fernandamarr Thanks for contributing!

@105ron Good catch!

PR #10917, opened 2019-06-09
fixed grammar

samwillis4216:
Changed "working on a team" to "working in a team"


leila-alderman:
@samwillis4216 I believe that "working on a team" is more commonly used than "working in a team", but since you found it a bit confusing, I changed to "working as part of a team". Hopefully you find that clearer. 

Thanks for contributing!

PR #10920, opened 2019-06-09
Update Express-Mini-Message-Board.md

saifmode:
I changed app.get to router.get because that's how the Express Generator configures the routes. app.get isn't used in either the routes or app.js

Added a helpful hint in bullet 4 where I got stuck trying to send variables to my templates.

Finally,  corrected some copy and fixed the bullets, which had some duplicate numbers.


codyloyd:
nice! thanks for this!

PR #10929, opened 2019-06-10
Adding the solution of Jose Guaura

jguaura:

Thank you,

The Odin Project team.


105ron:
Hi @jguaura,
The mark down in this file is used for our website so it is important it is correct. Your solution should be inside the details tags and be formatted the same as other students. You can enable github pages if you want other students to be able to view your solution or remove the part from your contribution. Thanks.

PR #10983, opened 2019-06-15
added my solution

allan-glasier:

Thank you,

The Odin Project team.


105ron:
Hi, it looks like you've run the file through an auto formatter, which is ok. But line 38, the four extra spaces for the closing `div` don't really make sense, could you please remove them thanks.

allan-glasier:
Sorry about that! I've got prettier that formats on save, changed that to only when I ask!

PR #11007, opened 2019-06-17
Remove empty "Additional resources"

shafrazi:
Removed empty "Additional resources" section

Thank you,

The Odin Project team.


105ron:
Hi @shafrazi,
I would prefer to keep the section there and simply add a point that we need some added, so that students viewing the page will be prompted to add additional resources if they know any when they view the lesson. [This page](https://www.theodinproject.com/courses/getting-hired/lessons/strategy?ref=lnav) would be the example I would follow.

codyloyd:
agree with @105ron

PR #11012, opened 2019-06-18
Update link to Steve Yegge blogpost

Duncan-7:
Previous link was to google plus and no longer worked.


leila-alderman:
Thanks for tracking this down and for contributing!

Closes issue #11019

PR #11047, opened 2019-06-21
Change subheading of "Take Breaks" and edit body

MicoPham:
For the section "Pitfalls to Avoid", the subheading "Take Breaks!" did not flow well with the outline of the article nor did it make sense.  
Example 
Section "Pitfalls to Avoid"  
-   "Procrastination", "Take Breaks!", "Digital Distractions", "Rabbit Holes", "Comparing Yourself to Others".  

"Not Taking Breaks" is the right fit as the subheading for this section.

Also edited body paragraph of "Not Taking Breaks" for better readability.


bycdiaz:
@MicoPham Good call on the edits. Thanks!

PR #11167, opened 2019-07-02
Make proofreading changes to lesson content

coped:
I noticed some small punctuation errors and awkward phrasing within the lesson content of Ruby On Rails: Project: Let's Get Building.

This PR features my suggested changes. All are very small in scale.


leila-alderman:
Thanks for contributing @coped!

PR #11230, opened 2019-07-06
Add keyboard shortcut F12 for dev tools

JoshAubrey:
Hi, I just noticed this was missing the simpler keyboard shortcut.

Thanks,
Josh


leila-alderman:
This is a great addition! Thanks for contributing @JoshAubrey!

PR #11247, opened 2019-07-08

Welteam:
May be an error on my part but I believe this field is meant for the "car maker"

Thank you,

The Odin Project team.


leila-alderman:
@Welteam It's common to refer to the "make and model" of a car (https://www.carinsurance.com/make-model-definition), so this is correct as is. I'm going to close this PR, but thanks for looking into this and for contributing to TOP!

PR #11248, opened 2019-07-08
Update setting_up_git.md

Welteam:
Sometimes (especially if you already have an ssh key and/or used a different path than default), the key isn’t automatically added to ssh. That’s why I think the following steps should be added.

Thank you,

The Odin Project team.

### DIFF
--- a/web_development_101/installations/setting_up_git.md
+++ b/web_development_101/installations/setting_up_git.md
@@ -121,6 +121,26 @@ Now, go back to GitHub in your browser window and paste the key you copied into 
 
 #### Step 2.5 Testing your key
 
+First, verify that your new key is being used:
+
+~~~bash
+ssh-add -l -E md5
+~~~
+
+The output should be:
+
+~~~bash
+2048 MD5:[the signature] [your email] (RSA)
+~~~
+
+If the output is "The agent has no identities.", you need to add it. For Linux/WSL, enter:
+
+~~~bash
+echo 'ssh-add ~/.ssh/id_rsa &>/dev/null' >> ~/.bashrc
+~~~
+
+If you are on MacOS, simply replace "bashrc" with "bash_profile".
+
 https://help.github.com/en/articles/testing-your-ssh-connection
 
 Go to that link and verify everything matches up, if it doesn't try doing these steps again, or coming to the discord chat for help. 


web_development_101/installations/setting_up_git.md
```diff
@@ -121,6 +121,26 @@ Now, go back to GitHub in your browser window and paste the key you copied into
...
+The output should be:
+
+~~~bash
+2048 MD5:[the signature] [your email] (RSA)
+~~~
+
+If the output is "The agent has no identities.", you need to add it. For Linux/WSL, enter:
+
+~~~bash
+echo 'ssh-add ~/.ssh/id_rsa &>/dev/null' >> ~/.bashrc
```
I3uckwheat: I don't think we want to add this to the bashrc/bash_profile. This would attempt to add it to the authentication agent every time someone opened a terminal.  Maybe it would be better to have the ssh-add command be ran when setting up git earlier. see [here]https://help.github.com/en/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agentl) . It looks like I forgot to add that while writing that lesson. Although, `~/.ssh/id_rsa` should be located by the auth agent automagically.

web_development_101/installations/setting_up_git.md
```diff
@@ -121,6 +121,26 @@ Now, go back to GitHub in your browser window and paste the key you copied into
...
+~~~
+
+If you are on MacOS, simply replace "bashrc" with "bash_profile".
```
I3uckwheat: This will have people asking where "bashrc" is

web_development_101/installations/setting_up_git.md
Welteam: The problem with this method is that it only last until you close the terminal for some reasons. That's why my method add it every time you open the terminal

web_development_101/installations/setting_up_git.md
Welteam: True, I clearly don't have the skills to write an efficient lesson. I mainly did this to raise the issue and there definitely are better ways to formulate this.

leosoaivan:
Closing as this is not immediately relevant to new users with default environments.

PR #11323, opened 2019-07-13
Add my solution and fix Leo Holanda markdown typo

nikolasbroman:

Thank you,

The Odin Project team.


105ron:
Thanks for the fix. We should add a comment to the PR template text about spaces within the mark down not being valid.

PR #11359, opened 2019-07-16
Update rake to rails

suretrust:
I updated rake to rails in the conclusion.md file of the rails curriculum.


105ron:
Thanks. That's a change that should have been made some time ago.

PR #11431, opened 2019-07-21
Kenneth

BinarySamuraii:
My solution to assignment Easy https://codepen.io/BinarySamurai/full/gVpEoo


105ron:
Hi Kenneth,
we need your solution between the `details` tags and formatted the same as the other as the other student solutions. Thanks.

PR #11437, opened 2019-07-22
Submit to student links

nilajawill:
Added my own project to the list. 

However, the output on my github page is not what is shown while using VS Code. Aside from that, project is completed.

Thank you very much for the project. It was fun!

Nellie Williams


105ron:
Github pages can be a little buggy at times. Sometimes changing the `index.html` file can resolve the issue. It's not worth your time trying to resolve it at the moment, continue on with the curriculum.

nilajawill:
Okay, thank you very much!! :D

PR #11446, opened 2019-07-22
Adds an example for command generating SSH key.

bycdiaz:
I removed the < > from the SSH generation command and added an example. I've noticed plenty of folks get tripped up on this step.  Hopefully this will help.

### DIFF
--- a/web_development_101/installations/setting_up_git.md
+++ b/web_development_101/installations/setting_up_git.md
@@ -95,7 +95,7 @@ If the message in the console contains `No such file or directory`, then you don
 To create a new SSH key, run the following command inside your terminal. The `-C` flag followed by our email address ensures that GitHub knows who we are.
 
 ~~~bash
-ssh-keygen -C youremail
+ssh-keygen -C <youremail>
 ~~~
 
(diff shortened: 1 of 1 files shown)


leila-alderman:
I kinda like the angle brackets around the variable that you're supposed to replace. I think that's clearer than removing them. Definitely keep the example though!

bycdiaz:
@leila-alderman I don't mind the angle brackets at all. I only removed them because a bunch of people were including them in their command. I'll add them back in and hopefully the example helps.

leila-alderman:
You're right. After a quick glance, it looks like this is the first time we use that convention in the curriculum. It's a common enough convention that I think it's worth keeping, but I think you're right that it's easy to misunderstand. Maybe we should add a quick statement about how angle brackets are used for placeholders in commands throughout?

bycdiaz:
Just added a bit of context. Let me know if that was what you had in mind.

leila-alderman:
Thanks @bycdiaz! I moved the text to above the code snippet and a bit more context. 

Ready to be merged after you look it over!

bycdiaz:
I like it.

PR #11458, opened 2019-07-23
Fix minor typos in sinatra.md

billclaytor:
Thanks for this great doc; it has helped me get started with Sinatra! I first noticed the missing space between these sentences:
"Take a look at the Sinatra README and read through the first eight sections of it ( until section 8, “Helpers”).Almost everything you will need to know about Sinatra is right here on this page so use it as a reference when you are working through the project."

I took a closer look and found a few other minor issues. Hopefully this helps :)


mindovermiles262:
Looks good

Thanks @billclaytor 👍

PR #11524, opened 2019-07-29
Update installing_ruby.md

grdguez:
Most Linux distributions recognize and recommend using apt instead of apt-get. I think to avoid confusion we should follow that pattern here too.

Thank you,

The Odin Project team.

### DIFF
--- a/web_development_101/installations/installing_ruby.md
+++ b/web_development_101/installations/installing_ruby.md
@@ -35,8 +35,8 @@ The rest of the installation will take place inside the terminal window.
 First, we need to make sure your Linux distribution is up to date. Run these commands one by one. Because these commands use `sudo`, you will have to enter your password in order for them to run. When typing your password, you may not get any visual feedback, but rest assured that your password is being entered. Once you're done typing your password, press `enter`.
 
 ~~~bash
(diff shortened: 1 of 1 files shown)


web_development_101/installations/installing_ruby.md
```diff
@@ -70,10 +70,9 @@ Note: Run these commands one by one in sequence. They will not provide any outpu
 ~~~bash
 echo 'export PATH="$HOME/.rbenv/bin:$PATH"' >> ~/.bashrc
 echo 'eval "$(rbenv init -)"' >> ~/.bashrc
-exit
+source ~/.bashrc
```
I3uckwheat: This change doesn't fit with the rest of the changes. Please submit this as another pull request if you think this should be included.

web_development_101/installations/installing_ruby.md
```diff
@@ -70,10 +70,9 @@ Note: Run these commands one by one in sequence. They will not provide any outpu
...
 ~~~
 
-After running the final `exit` command, you will need to re-open the terminal (see Step 1.1 above).
```
I3uckwheat: Same as above comment.

PR #11533, opened 2019-07-29
Update to include new user solution

anncancode:
Adding my etch-a-sketch solution.


bycdiaz:
@anncancode I love your use of the knobs. Merged!

PR #11628, opened 2019-08-04
added new "Web Development 101" directory specifically for front end track. deleted/edited files pertaining to ruby/rails/backend

dekulinkk:
I noticed that the front end track uses the same Web Development 101 course as the full stack track, so I cloned and edited it to be used specifically for the front end track.


105ron:
This will increase the maintenance of the curriculum as the files are all duplicated. When a resource changes we need to edit the two files. For this to work we also need to change the seeds in the Rails app repository. Maybe the structure of the curriculum needs to change but it would require a collaboration with the maintainers about the best approach.

PR #11682, opened 2019-08-08
Update unstable link to use the original source

MauricioRobayo:
On [html_css/design_ux.md](https://github.com/TheOdinProject/curriculum/blob/master/html_css/design_ux.md) there is a link to this resource:

https://blog.dashburst.com/infographic/ux-101-what-is-user-experience/

But that link more often than not produces an `Error establishing a database connection`.

I could access the page using Google cache:

https://webcache.googleusercontent.com/search?q=cache:NvYcYL1eOCEJ:https://blog.dashburst.com/infographic/ux-101-what-is-user-experience/

At the end of the page is the source to the original publication, by homestead.com, which is working fine:

https://www.homestead.com/blog/06/2013/ux-101-what-user-experience-infographic

This pull request updates the broken link to point to the original source of the infographic.

### DIFF
--- a/html_css/design_ux.md
+++ b/html_css/design_ux.md
@@ -25,9 +25,9 @@ Remember that your goal here is to absorb the most important concepts and contin
 
 <div class="lesson-content__panel" markdown="1">
 1. Read [Startups, This is How Design Works](http://startupsthisishowdesignworks.com/)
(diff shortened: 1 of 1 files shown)


leila-alderman:
@MauricioRobayo Thanks for tracking down the updated link and for taking the time to correct it here!

PR #11719, opened 2019-08-11
Added my own solution.

GaiaLancer:
Search boxes and buttons are difficult. :frowning_face:


105ron:
They are but these things get easier with practice. Just keep at it.

PR #11723, opened 2019-08-11
Odin's HTML/CSS Project (Google Page - Easy)

AlexARNcode:
And this is my first contribution to an OpenSource Project by the way. Hope I did it correctly :).


105ron:
Almost. The first link should be https://github.com/alexarncode/google-clone/. Would you like to edit this pull request and fix it? Everything else is good.

PR #11735, opened 2019-08-12
Update Flickr REST API Endpoint

SilasSpet:
The Flickr endpoint changed. As you can see [here](https://www.flickr.com/services/api/request.rest.html) the subdomain is no longer "api" and instead is now the normal "www".


mindovermiles262:
Hey @SilasSpet 👋 

I think the point that the section was trying make was that you will need to use the `https://api.flickr.com` endpoint in your API requests.

According to their [API Overview](https://www.flickr.com/services/api/misc.overview.html) page this is the URL to use.

Thanks for submitting this PR but I am going to close it with no action.

Let me knwo if you have any questions

Thanks

SilasSpet:
@mindovermiles262 I don't know why the API Overview page still uses api.flicker.com because if you visit their API Explorer they use www.flicker.com. Anyways, I still think the lesson should be updated, because there is a specific link being provided and the content of the link no longer matches the content of the lesson. As you can see [here](https://web.archive.org/web/20180310070922/https://www.flickr.com/services/api/request.rest.html) it used to be api.flicker.com but if you visit this page now ([here](https://www.flickr.com/services/api/request.rest.html)) it changed. So I think point 3 definitely needs to change, because it is supposed to reflect the important line of the API documentation and because it's no longer api.flicker.com it fails to do so.

If you follow point 6 and go to the [Search API Explorer](https://www.flickr.com/services/api/explore/flickr.photos.search) and follow the instructions this section provides then you can see at the bottom that it uses www.flicker.com and not api.flicker.com like section 7 says. Further proving my point that these sections need to change because flicker changed the here important parts of their API subdomain.

PR #11766, opened 2019-08-14
Update api_basics.md

gregbast1994:
I think this is worded better?


mindovermiles262:
Thanks @gregbast1994 ! Welcome to TOP and Github!

PR #11782, opened 2019-08-15
Fix another typo

nikolasbroman:
Example code's method `save_thank_you_letters` is referenced in the explaining text as `save_thank_you_letter` (without an 's' in the end).


105ron:
Thank you. I'm sure there's a few more typo's in the curriculum. :)

PR #11878, opened 2019-08-21
added student solution

mjwills-inf:
I'm unsure if I have made an error with partr of this contributing pull request as it did not seem to go as smoothely as the first time when following the video step by step.

Apologies if so


105ron:
It's because you have an outdated version of the repository on your machine because you cloned it a bit before you made the pull request. You need to add this repo as a remote and pull from it to prevent this happening in the future. The issue you had is called a merge conflict. Two people have changed the same line and git doesn't know which change to accept so it gives you the two options between conflict markers FYI.

mjwills-inf:
Thank you for taking the time to explain.

All the best

________________________________

PR #11960, opened 2019-08-26
reorder conditionals flow so that student has knowledge of conditiona…

p0rt3rd:
…ls prior to logical operators

(below is the copy of a previous pull request that I closed. I had to do it this way because this pull request was previously pending, and I needed to create a new pull request. I could not do it because I didn't have branches set up at the time. I'm sure there was probably a better way, but the only way I could figure out how to created 2 pull requests at the same time was to delete the old one, create branches, and resubmit it.)

In the current Fundamentals Part 2 lesson in JavaScript Basics, the student encounters a lesson about logical operators before learning about conditionals (the lesson about conditionals is listed directly below the logical operators lesson.) But the logical operators lesson requires knowledge of conditionals to complete the exercises on the page. (knowledge of [if, else, etc...] is required, but not included in any lesson previously.)

I discovered this when completing the lesson. I was eventually able to learn on my own and finish the final exercise on the logical operators page, but it would have certainly been much less difficult, not to mention confusing, if the lesson that taught me about conditionals took place prior to, and not after, I needed knowledge of them.

I did my best to reorder the lessons in a way that made sense. But feel free to change the wording and structure to whatever works best, but I do believe the lesson about conditionals should be included before the lesson on logical operators.

### DIFF
--- a/web_development_101/javascript_basics/fundamentals-2.md
+++ b/web_development_101/javascript_basics/fundamentals-2.md
@@ -35,11 +35,12 @@ Depending on what kind of work you're doing, you might end up working more with 
 
 Now it's time for the fun stuff...  So far we haven't done much with our programming that you couldn't do with simple math skills.  Sure, we've told our computer how to do the math, so that makes it quicker, but the essence of programming is teaching the computer how to make decisions in order to do more involved things.  Conditionals are how we do that.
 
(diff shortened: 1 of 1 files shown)


105ron:
Since Cody wrote the JS curriculum I'll let him review it and make suggestions. It might take a while but he will get to it. Thanks for your suggestion.

PR #11967, opened 2019-08-27
[Add note in Testing Practice to use import]

rvvergara:
Currently most are using `require` to import modules in a `.test.js` file. The more updated way is to use `ES6` import statements.

### DIFF
--- a/javascript/testing/testing-practice.md
+++ b/javascript/testing/testing-practice.md
@@ -33,6 +33,27 @@ Write tests for the following functions, and then make the tests pass!
    ~~~
 </div>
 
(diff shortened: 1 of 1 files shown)


rvvergara:
Hi @codyloyd. Would love to get your insights related to this change I am proposing :) I figure we need to update the instructions for this project as currently it's still using the old ES5 syntax only (i.e, `import` statements in test files get errors)

105ron:
Let's just merge it and see if he has any suggestions... Thanks for helping us keep the curriculum up to date.

rvvergara:
Thanks @105ron! :)

PR #12012, opened 2019-08-30
Adds link to Ruby Libraries content.

bycdiaz:


leila-alderman:
Thanks for finding this great resource! I think it's definitely a great addition, although I feel like this some pretty heady stuff to be adding alongside "open Ruby using IRB", especially as it gets into Bundler and other fairly Rails-y things. 

Couple of alternative suggestions:
1. Perhaps this would considered more "intermediate" knowledge and could find a home alongside the OOP lesson: https://www.theodinproject.com/courses/ruby-programming/lessons/object-oriented-programming ? 
2. Consider adding it to the Rails section inside Web Dev 101: https://www.theodinproject.com/courses/web-development-101/lessons/ruby-on-rails-basics?ref=lnav

bycdiaz:
@leila-alderman Thanks for taking a look. It does feel a little meaty for this portion now that you pointed out other potential spots for it. I'll move it to the Rails section in web dev 101.

PR #12020, opened 2019-08-31
Adds link to Ruby Libraries content.

bycdiaz:


bycdiaz:
@I3uckwheat the resource I've proposed linking to just briefly introduces the ideas behind, and the use of, the Ruby Standard Library, Rubygems, Bundler, and the Ruby Load Path. Doesn't take a deep dive. In the case of Rubygems, they are used but never explicitly explained in any prior lesson. The resource is more to make a learner aware of the role of these tools rather than a deep dive.

PR #12034, opened 2019-09-01
Updated Xubuntu ISO link in Dual-Boot section

AKey:


105ron:
Thank you for the update.

PR #12036, opened 2019-09-01
Fix typo

shermix:
Replaced "AND" to "and"


105ron:
Should we also replace json with JSON whilst we are at it?

shermix:
Yeah that would be better. As later in the article all json's are expressed in uppercase as JSON. So yeah, good to go. :+1:

CouchofTomato:
@shermix 

It still appears to be `views and json`.

Are you going to submit a commit for it to be JSON?

shermix:
> @shermix
> 
> It still appears to be `views and json`.
> 
> Are you going to submit a commit for it to be JSON?

I fear this is my first contribution and experience with pull requests in general. I can't figure out how I make commit on this patch-2 branch from the forked repo.

105ron:
You managed to fix it on the second commit. Merged, thank you.

PR #12039, opened 2019-09-01
Fix codeblocks rendering as plaintext on website

shermix:
Inserted newlines after paras so that code blocks are rendered not as normal text on the main website. Currently, many code blocks in this article are being rendered as plain text on the website because of the lack of newline between a paragraph and commencing code block codes in the markdown file.


shermix:
> Inserted newlines after paras so that code blocks are rendered not as normal text on the main website. Currently, many code blocks in this article are being rendered as plain text on the website because of the lack of newline between a paragraph and commencing code block codes in the markdown file.

[Here\(https://i.imgur.com/DoHEAQB.png) is how some code blocks are rendered on the website due to absence of newlines in markdown.

105ron:
Thank you for the fix.

PR #12059, opened 2019-09-02
Replace broken links for 'You didn't know JS'

suretrust:
The "You don't know JS" links are broken as shown below because it now has two editions -- the first and second. The second edition is still a work in progress, so I replaced the link with the first edition as is.

one
Two


CouchofTomato:
Hi @suretrust 

Both the links work for me as they are. Perhaps github was having an issue.

105ron:
A PR fixed them the previous day, but the site wasn't updated when this was opened. The site is using the 2nd edition now, I'm not sure how far it is from being finished. I guess people will complain if it's not good enough.

PR #12061, opened 2019-09-02
Added student solution

Etheonor:
This is the first JS thing I'm proud of!


105ron:
Nice work, keep it up and make many more things to be proud of.

PR #12107, opened 2019-09-05
Replacing links to new paid version of the Ruby on Rails tutorial with links to the old,free version

dvik1950:


leila-alderman:
Looks good - thanks for updating these links and for turning it around so quickly!

PR #12144, opened 2019-09-07
Update code listing in 'Databases and SQL'

cloudtemplar:
I guess grouping by User's ID seems more sensible- you can run into situation when user names will be duplicated, which leads to a situation where Posts of different Users with the same name will be treated as coming from just one User, so grouped incorrectly.
There is also an option that User's name in this listing is meant to be unique- if so, I think it's worth to emphasise this fact.


CouchofTomato:
I would guess that when this was originally written user.name may have meant to express a username which would be unique perhaps, or just that it was an example to illustrate the point about grouping rather than diving deeply into the details of SQL queries.

I personally don't think it hurts to change it but I don't think it's an issue keeping it as it is either.

Let's see if @KevinMulhern has an opinion on the matter.

PR #12186, opened 2019-09-10
My solution

Cacurriapower:
Did the random color version but i made it lower each rgb value individually by 10% on each pass so it takes more than 10 passes to reach full black, however it works perfectly. You can also reset it to any grid side that you want however anything past 200 kinda slows down my VM too much.


105ron:
Did you enable Github pages? Your view in browser link doesn't work. You need to enable it in the repository settings...

Cacurriapower:
Yeah, sorry forgot to. Its enabled now.

PR #12259, opened 2019-09-16
Update new repo setup instructions

nearmint:
In the instructions, you're right now telling the user to clone the https link from git (git clone https://github.com/YourUserName/google-homepage). If the user does this, they'll have to enter user/password in their console everytime they make a commit. This happened to me. Therefore I suggest to tell the user to git clone the SSH link instead (see edit).


I3uckwheat:
If we do this, we should probably mention how to fix it if something goes wrong and they don't have their SSH key setup @nearmint

Issue #12086, opened 2019-09-04
Formerly free resource for Git Basics assignment now requires payment


bycdiaz:
The first assignment for the [Git Basics ](https://www.theodinproject.com/courses/web-development-101/lessons/git-basics) lesson is now part of the pro package at codecademy. 

The lesson is the Basic Git Workflow from the [Learn Git course at codecademy](https://www.codecademy.com/learn/learn-git).

We'll need to identify a new resource for this lesson.

jmcarter006:
I’m on mobile so this page isn’t working right. But it looks like this resource may be a viable interactive git learning tool. Especially the “Learn by Doing” section 

https://try.github.io/

Rickeo:
Following on from what @jmcarter006 said, specifically the link on that page to this resource:

https://learngitbranching.js.org/

That seems to be the bulk of the interactive lesson(s).

As someone who has recently just done that Code Academy course, having a brief look at the content on js.org it seems to cover the same type of material.

bycdiaz:
I came across this resource:[git-tower.com](https://www.git-tower.com/learn/git/ebook/en/command-line/introduction#start) It feels a bit more accessible than the others. If I'm not mistaken, Part 1 - The Basics should give learners what Codecademy was covering.

I only have two (small) concerns about this resource. It mentions a set of lessons using a desktop GUI. We would just need to be explicit about folks sticking to the command line version.

Also, during the [Getting Ready lesson](https://www.git-tower.com/learn/git/ebook/en/command-line/basics/getting-ready#start), they don't mention installing in Linux (only Mac OS and Windows). But we'll just need to remind learners that they should have installed git before arriving to this lesson.

bycdiaz:
Resolved by pull request #12291.

PR #12291, opened 2019-09-17
Replaces codecademy's git lesson with Tower's.

bycdiaz:
Resolves issue #12086.

### DIFF
--- a/web_development_101/git_basics/git_basics.md
+++ b/web_development_101/git_basics/git_basics.md
@@ -15,7 +15,7 @@ By the end of this lesson, you should be able to do the following:
 
 <div class="lesson-content__panel" markdown="1">
 
-  1. Complete the first interactive lesson in the Codecademy [Basic Git Workflow](https://www.codecademy.com/learn/learn-git) course. Try to look for a pattern in the commands you are running.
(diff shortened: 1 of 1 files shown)


mindovermiles262:
I think it would suffice to say 

```
[Tower](https://www.git-tower.com) has a sweet lesson on version control with Git. You only need to do Part 1 - The Basics. Visit [Tower's Learn Version Control with Git lesson](https://www.git-tower.com/learn/git/ebook/en/command-line/basics/what-is-version-control#start) to get started.
```

It is more to the point and links directly to the page where they should start

I3uckwheat:
I think there is value stating to not use the GUI version for this, people will invariably try to do the GUI version and miss out on a lot of topics

bycdiaz:
Both links present users with note: 
image

My concern is that users will assume that doing a GUI will be acceptable. Or will result in many learners coming to the chat to ask which approach they should take. Including a statement about using the command line version could give clarity to new learners and save us from fielding the question in the chat.

The [Getting Ready](https://www.git-tower.com/learn/git/ebook/en/command-line/basics/getting-ready#start) lesson assumes the user is using MacOS or Windows and makes no reference to Linux and does not offer a way to check if Git is already installed. I know this is obvious for us. New learners wouldn't have this awareness and not being explicit about this would only confuse them.

While wordy, my sense is that the context is necessary. I will look at it again and see if I can make it a bit more concise. Naturally, since this is a group effort, I'm happy to go with a majority. I'll invite others to weigh in.

PR #12322, opened 2019-09-20
Fix/project_media.md

mvedataydin:
Fixing : student solution not properly added.


105ron:
Thanks for the fix. :+1:

PR #12481, opened 2019-10-02
Replaces Tower Git course with Udacity's

bycdiaz:
Udacity's lessons feel fairly accessible to new learners, which I think is key for lessons early in our curriculum.


bycdiaz:
After some discussion, we've decided to write some material straight into our lesson and passing on the inclusion of the resource initially proposed in this pull request.

PR #12543, opened 2019-10-04
Guest Additions note re emptying CD drive

martink-rsa:
This issue keeps popping up in the Discord servers which usually turns out to be:

i. The host OS needs to be rebooted.
ii. The CD image files need to be unmounted.

A paragraph has been inserted to address this.

Please could you see if it should be simplified for those who are less experienced with computers.


I3uckwheat:
I think this will suffice for now, however, most people will not understand what the steps to resolve this are. @martink-rsa Can you make an issue on this so we don't forget to continue to move forward on this? 

I am thinking adding screenshots to the entirety of the VM installations would be beneficial.

martink-rsa:
No problem, will do later.

PR #12545, opened 2019-10-05
Changed section id="hero" to class="hero"

mkezmkez:
Hello, I just found a tiny error in the description of the project, in point #7

7. Set up the main background image. You can grab that exact image from their webpage by finding the <section id="hero"> tag in your inspector and checking out its background image.

should be <section class="hero">

Thank you,

Luis Puente Farias


105ron:
Thank you Luis.

PR #12567, opened 2019-10-07
add Andy's rock-paper-scissor to solutions

andaylin:

Thank you,

The Odin Project team.


105ron:
Hi, could you please turn off your auto formatter so you are only changing the lines needed to be added for your solution please.

Issue #12553, opened 2019-10-05
Remove mentions of Rails from JS Track


leosoaivan:
## Description
After breaking out JavaScript/Node.js into its own track, we are still referencing Rails in a few lessons, including but not limited to shared lessons, such as WebDev101.

To begin addressing this issue, the following should be addressed:

- [x] Identity and list any JS/Node lessons (with links in the repository) that mention Rails that are **not** in shared tracks, if any.
- [x] If such lessons exist, discuss with mods the course of action.
- [ ] Make edits as necessary

This issue can be considered close when the above tasks have been successfully addressed.

## Related
* https://github.com/TheOdinProject/curriculum/issues/11148

Girlscoding-ng:
Hey @leosoaivan @arku Good day.... will like to work on this issue but will need some little direction on what to do

leosoaivan:
@Nelson-Chinedu thanks for taking a look at this.

The Odin Project's JS curriculum lives in this sub-directory: https://github.com/TheOdinProject/curriculum/tree/master/javascript.

We'd like to first identity where Ruby on Rails is mentioned. The way you search through this directory is up to you. GitHub search, or cloning the curriculum and doing a search on your IDE/text editor is fine, too.

The first step is creating a list of which pages mention Rails. Obviously, the entire section titled "JavaScript and Rails" will need to be addressed. See if you can find any other places.

Let us know if you have further questions, and thanks again for picking this up :) 👍

Nelson-Chinedu:
Ok, so is there any style guide to follow for contributing regarding branching and commit @leosoaivan

leila-alderman:
@Nelson-Chinedu We're not too particular about branching and committing. If you're looking for some guidance, we have [a section on contributing](https://github.com/TheOdinProject/curriculum/blob/master/contributing.md). If you're not sure where to start, I recommend [the first video linked there](https://www.youtube.com/watch?v=V74l_zS1x8E).

Nelson-Chinedu:
Ok I just raised a PR on https://github.com/TheOdinProject/curriculum/pull/12586 regarding the issue on what to do first @leila-alderman, will like to know if am on track with the issue

leila-alderman:
@Nelson-Chinedu Thanks for working on this issue and for creating a list of lessons! That's a great first step. We probably don't need to upload a new file with that list to our curriculum, though. This issue would be a great place to store that list to get started:

- [ ] conclusion.md
- [ ] project_final_js.md
[...]

PR #12572, opened 2019-10-07
Removed ruby and rails course mentions

ribaishtiaq:
Removed ruby and rails course mentions from the first lesson of javascript course.


leosoaivan:
Thank you for taking a stab at this, but this is already addressed and claimed in #12553.

I'm closing this, but feel free to following the aforementioned issue.

PR #12623, opened 2019-10-10
Fix grammar in JS Fundamentals Part 1

disco-coder:
Just a minor grammar fix - changed singular "type" to plural "types".


CouchofTomato:
Thanks @disco-coder

PR #12711, opened 2019-10-17
Fix typo from yahoo to flickr

3starblaze:
Yahoo isn't required to get flickr developer api. I'm fairly certain it was meant to say flickr instead of yahoo.


leila-alderman:
@3starblaze Thanks for contributing to The Odin Project and improving our curriculum!

PR #12714, opened 2019-10-17
Typo fix in installing_ruby.md

Legio1221:
Numbering on the macOS subsection is out of order. Changed from 1.1 to 2.1 (header is under #2).


bycdiaz:
@Legio1221 Nice catch. Thanks for editing.

PR #12749, opened 2019-10-20
remove Cloud9 from pair programming section, replace with Teletype

cecyc:
Addresses [#9426](https://github.com/TheOdinProject/curriculum/issues/9426): remove Cloud9 from pair programming section.

I added [Teletype](https://teletype.atom.io/) as an alternative, but I can also just remove the mention of Cloud9 altogether. LMK!


105ron:
Thanks, that will be enough to close the issue. I think there is a similar plug for VSC too.

PR #12869, opened 2019-10-28
Removed no content link and added new resource

kennguyen01:
The online chapters of the book Conquering the Command Line content cannot be displayed when I accessed it. I removed the outdated link and added an interactive tutorial on command line from Linux Journey.


bycdiaz:
@kennguyen01 - The author fixed the issue with the resource and it is now back online. Thanks for the proposed fix! We'll be keeping the lesson as is for now.

PR #12878, opened 2019-10-29
Fixed formatting

vedantshetty:
Step 6 in project 1(binary search tree) was being displayed as an H1
Screenshot from 2019-10-29 11-27-05


105ron:
Thanks. 👍

PR #12945, opened 2019-11-02
Added fix for error "E: Package 'make' has no installation candidate"

martink-rsa:
**Issue:**

Some individuals encounter an error in point 3 of Step 3: Install and Enable Guest Additions: https://www.theodinproject.com/courses/web-development-101/lessons/prerequisites#step-3-install-and-enable-guest-additions

Point 3 reads as follows:
> 3. Type the following command into the terminal: `sudo apt install gcc make perl` You will be asked to type in the password you setup earlier. As you type your password, you'll notice there is no visual feedback. This is a security measure. Trust that it is taking you input. (Just type it and then push enter). If an error is thrown, reboot the VM and try the steps in this list again.

The error typically encountered states: 
> Reading dependency lists... Done
> Building dependency tree
> Reading state information... Done
> Package make is not available, but is referred to by another package.
> This may mean that the package is missing, has been obsoleted, or is only available from another source
> 
> E: Package 'make' has no installation candidate 

**Fix:**

Running `sudo apt-get update` and `sudo apt-get upgrade` in the respective order should clear the issue, whereafter `sudo apt install gcc make perl` should run successfully.

**Proof/Example of it working:**

Discord #general channel chat logs on 2 November 2019 at 13:08 - 13:13 GMT +2.
image

### DIFF
--- a/web_development_101/installations/prerequisites.md
+++ b/web_development_101/installations/prerequisites.md
@@ -63,7 +63,7 @@ While your VM is running, do the following steps:
 
   1. Click "Devices" -> "Insert Guest additions CD image" in the menu bar
   2. Open a terminal by pushing `ctrl + alt+ t` on the keyboard, if a terminal does not open, click anywhere on the desktop of the VM and try again.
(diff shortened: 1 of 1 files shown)


martink-rsa:
The above was discussed with @bycdiaz who recommended that the fix be moved to the steps.  Changed accordingly.

bycdiaz:
Made a couple of adjustments to the language.

I3uckwheat:
Thanks for this :+1:

PR #13019, opened 2019-11-06
Fixed incorrect link in Javascript Fundamentals Part 3

Kineara:
Corrected Javascript Fundamentals Part 3 Section 5 link to Function Expressions and Arrow Functions article - the original article appears to have been split into two separate ones, and the original link is no longer valid.


leila-alderman:
@Kineara Thanks so much for taking the time to correct this broken link!

PR #13082, opened 2019-11-11
Change order of steps in Binary Search Tree Project

leavism:
In the current instruction order, it isn't clear that the `#build_tree` method was supposed to be under the `Tree` class. Swapping the steps (2 and 3) and clarifying that the `#build_tree` method will be written in the next step should be easier to understand.


105ron:
An older PR has resolved this issue I believe, now closed.

PR #13098, opened 2019-11-11
Remove codecademy

I3uckwheat:
Removes codecademy from curriculum

Removing due to Codecademy changing their business model, there is no longer a free section due to the onset of the "premium" membership


bycdiaz:
I think this looks solid. :+1:

PR #13126, opened 2019-11-13
Should clarify that tip doesn't guarantee a balanced tree

leavism:
Spent 2 hours wondering why my level-ordered array couldn't produce a balanced tree, as suggested by the tip in step 9. Turns out that it's not a guaranteed way to rebalance a tree. Kept referring back to my notes and debugging until I realized it wasn't a guarantee. Found this [GeekForGeek](https://www.geeksforgeeks.org/sorted-array-to-balanced-bst/) page about rebalancing.

(Example array that doesn't work: [12, 5 ,15, 3, 7, 13, 17, 1, 4, 9, 14, 20, 8, 11, 8])

I'd suggest **at least** mentioning that the tip isn't a guaranteed way to rebalance a tree and that the reader should research their own way.

### DIFF
--- a/ruby_programming/computer_science/project_data_structures_algorithms.md
+++ b/ruby_programming/computer_science/project_data_structures_algorithms.md
@@ -29,7 +29,7 @@ You'll build a simple binary search tree in this assignment. In this lesson, our
   8. Write a `#balanced?` method which checks if the tree is balanced. A balanced tree is one where the difference between heights of left subtree and right subtree is not more than 1.
 
   9. Write a `#rebalance!` method which rebalances an unbalanced tree.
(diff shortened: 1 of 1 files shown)


I3uckwheat:
@leavism Because the directions say that "build_tree" is supposed to make a balanced tree, I do not think this is valid. 

> Write a method #build_tree which takes an array of data (e.g. [1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]) and turns it into a balanced binary tree full of Node objects appropriately placed (don't forget to sort and remove duplicates!). The #build_tree method should return the level-1 root node.

PR #13237, opened 2019-11-20
Change a sentence to make it more clear

olav35:
When I did this project a couple a couple of days ago I found this formulation a bit confusing and someone in TOP's discord channel suggested I send in a pull request to correct it.

The reason I found it confusing was that it used the word still. Earlier in the assignment it explicitly said to use return as opposed to console.log(). To me it initially seemed like "you should still just be using `console.log()`" suggested that we either where supposed to use console.log() earlier or return now.


105ron:
Looks good, thanks for the fix.

PR #13238, opened 2019-11-20
Clarifying instructions in ruby_programming/cs/project_data_structures

jjvaz25:
I ran into some confusion when doing assignment 1 about the initial steps to creating a balanced binary search tree. I asked for clarification on DiscordApp and realized that many people shared my confusion, resulting in many interpretations of the project instructions.

It isn't clear that we should be sorting the array before building the BST and using recursion to set the root(s). If we don't sort the array and use recursion, the tree will not be balanced. 

Thanks,
Jeff


105ron:
There's a significant amount of new lines here - It won't render very well on the web site.

I3uckwheat:
Closing for inactivity

PR #13283, opened 2019-11-23
Ruby Rogues Podcast URL Update

TimothyMalahy:
Ruby Rogues podcast added in another layer in the URL. I went through the podcasts on the site and located the original one.

Updated the URL link.


bycdiaz:
@TimothyMalahy Thanks your catching and correcting this.

PR #13290, opened 2019-11-24
fixing the problem with the nvm install

Jxtr0:
current solution does not work anymore, so I propose a new working solution. Hope it's usefull


I3uckwheat:
Closing as current instructions seem to work as expected.

PR #13365, opened 2019-11-28
Added clarification for which page to start from at 'Tower's Learn Ver…

skylermartyn:
There is no persistent display for where you are in the course without clicking 'Table of Contents.' The instructions for where to go are informative for their explanations of why you're skipping what you're skipping, but this process is inherently confusing. This edit makes it clear where you ought to start the lesson from.


I3uckwheat:
Thank you for the PR, but another has covered this. Appreciate the help!

PR #13475, opened 2019-12-05
move student solution back inside details dropdown

cgardn:
The last student solution for basic ruby project #2 (stock picker) was added outside the dropdown. This change moves it back inside.


mindovermiles262:
Thanks @cgardn 💯

PR #13518, opened 2019-12-08
Removing redundant paragraph from Introduction

Hahlh:
I chose the second occurrence, as the first one seems to fit better into the surrounding context.

The issue is easily visible here:

image


105ron:
Thanks for spotting and fixing this.

PR #13547, opened 2019-12-10
Update rbenv bash instructions

CouchofTomato:
The current instructions assume a bash shell. As the new MacOS uses zsh this update includes instructions for appending to the zshrc.


mindovermiles262:
LGTM 💯

PR #13730, opened 2019-12-20
Adding a short description to contribute.md

nearmint:
I personally am checking back sometimes how to contribute in a proper way, but I don't always want to rewatch the long video. For people like me, I am adding a short description of the proper way of contributing.


nearmint:
I have also added a minor wording update to javascript/objects-and-object-constructors to make it more easily readable and better to understand

nearmint:
I have added a little info to the installations section. Back when I was going through it, I thought we need to install vanilla Ubuntu. But I had a lot of problems with vanilla Ubuntu, since it used up so much RAM. It's possible to save up to 1GB in RAM usage by selecting another desktop environment such as Xubuntu or Kubuntu Minimal.

web_development_101/installations/installation_overview.md
```diff
@@ -36,7 +36,7 @@ If you're using a Mac, you're in great shape. Most professional developers rely
...
+[Linux](https://en.wikipedia.org/wiki/Linux) is a free and open-source operating system that works well with all programming languages. Most development tools are written to work natively with Linux. Your tools will likely be updated more often, have more information available for troubleshooting, and just plain run better on Linux. We'll be using Ubuntu, one of the most popular and user-friendly versions available. **If you don't use a Mac, we recommend that you use Linux.** It's that simple. If you opt to use Linux on a computer with 4gb RAM or less, it's best to use a light-weight version of Ubuntu to avoid lag or even freezing of the system. For more information on light-weight options, refer to this [table](https://www.reddit.com/r/Ubuntu/comments/8fz0dv/disk_and_ram_usage_of_ubuntu_kubuntu_lubuntu/dy7iye5/).
```
leila-alderman: We already recommend Xubuntu both for VMs (https://github.com/TheOdinProject/curriculum/blob/3a51b15fe0e15fbad7524b96536307117c4484a2/web_development_101/installations/prerequisites.md#step-12-download-linux) and for dual boots on older computers (https://github.com/TheOdinProject/curriculum/blob/3a51b15fe0e15fbad7524b96536307117c4484a2/web_development_101/installations/prerequisites.md#step-1-download-linux). 

Perhaps that would be a better location to add the specific recommendation for computers with 4 GB of RAM of less?
If you'd like to also add something here, then I'd suggest keeping it simple, e.g., "We'll be using Ubuntu, one of the most popular and user-friendly versions available, or the lighter weight alternative Xubuntu."

javascript/organizing-js/objects-constructors.md
```diff
@@ -28,9 +28,9 @@ Which method you use will depend on context. Dot notation is cleaner and is usua
 ~~~javascript
 const variable = 'property'
 
-myObject.variable // this gives us 'undefined' because it's literally looking for a property named 'variable' in our object
+myObject.variable // this gives us 'undefined' because it's looking for a property named 'variable' in our object
 
-myObject[variable] // 'Value!'
+myObject[variable] // same as myObject['property'] and gives us 'Value!'
```
leila-alderman:
```suggestion
myObject[variable] // this is equivalent to myObject['property'] and returns 'Value!'
```

contributing.md
```diff
@@ -7,3 +7,18 @@
 
 ### 2 The more involved but proper way
 * [Video](https://www.youtube.com/watch?v=mENDYhfxH-o)
+
+#### Short description of the proper way
```
leila-alderman: I definitely agree that there should be a written reference offered here, but I think it would be more useful to link to existing resources for this work flow. 

Could you add a link to the written workflow in The Odin Project (https://www.theodinproject.com/courses/ruby-programming/lessons/using-git-in-the-real-world#a-git-workflow-for-open-source-contribution) as well as to this blog post that seems to describe the flow well (https://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/)? 

If you find other resources helpful, please feel free to add those here too!

web_development_101/installations/installation_overview.md
nearmint: Back when I did the set up process, I completely skipped the VM and dual boot descriptions, because I decided to format my hard drive and do a fresh linux install. That's why I'd like to add something there.

Would this be okay? "We'll be using Ubuntu, one of the most popular and user-friendly versions available, or a lighter weight alternative based on Ubuntu, such as Xubuntu." 

I basically would like the user to know that he can feel save to use a distribution based on Ubuntu.

contributing.md
nearmint: Would you like me to delete the summary of the video? I feel like the articles are quite long. The idea was to have a very short summary of the commands themself without a long description, since the person already watched the video once or twice before and just needs to remember the workflow

nearmint: The issue is that this link (https://www.theodinproject.com/courses/ruby-programming/lessons/using-git-in-the-real-world#a-git-workflow-for-open-source-contribution) makes it harder to grasp the concept, because there is at least one step added that is not present in the videos: adding a feature branch.

Let me know how I should proceed. We can discard this change alltogether, or only use the links as you suggested.

nearmint: I have updated contributing.md as per your suggestions :)

leila-alderman: Creating a feature branch to work in is part of the correct Git workflow, so I do think it's worth including that extra step. 

As far as offering a quick reminder, my hope was that the blog post would provide that: 
image

Do you think that's sufficient or would something else be more helpful to you here?

nearmint: Yes, I guess that's totally sufficient! I added the two links. I hope we can get the changes approved. Also merry christmas, leila-alderman!

PR #13813, opened 2019-12-25
New solution to Rock, Paper, Scissors

ridwan102:
I would like to add my solution for Rock, Paper, Scissors. This is the version that is only played in the console.

Thank you,

The Odin Project team.


105ron:
Hi, we don't need a copy of your solution, only a link to the repository where you have pushed your solution to add to the student solution list. If you need help with this, head to discord and ask in the general chat. Thanks.

PR #13882, opened 2019-12-29
Fix minor wording errors

nearmint:
- Error: "Book prototype", has to be: "Book prototype instance" 
- Error: "it's", has to be: "its"


mindovermiles262:
👍 Thanks @nearmint

PR #13885, opened 2019-12-29
Added RPi Ruby installation

seoul2045:
This is the Ruby on Rails instructions for a Raspberry Pi Model 4 Computer Model B  4GB RAM. This installation worked while running the official Raspbian OS which was installed using [NOOBS](www.raspberrypi.org/downloads/noobs/). Some additional text from more experienced devs will help clarify the directions for first-time users. Also, there are additional instructions available to ensure the RVM source script doesn't need to be run for every new terminal. Further, additional instructions are available to install VS Code on the RPi.

### DIFF
--- a/web_development_101/installations/installing_ruby.md
+++ b/web_development_101/installations/installing_ruby.md
@@ -133,6 +133,174 @@ Well done! Pat yourself on the back! The hard part is done, and it's time to mov
 
 
 <details markdown="block">
+<summary class="dropDown-header">Raspbian for Raspberry Pi
+</summary>
+
+### Step 1: Install Updates, Packages and Libraries
+
+Before we can re-install Ruby, we need to install some base packages.
+
+#### Step 1.1: Open the Terminal
+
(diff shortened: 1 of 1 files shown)


web_development_101/installations/installing_ruby.md
```diff
@@ -132,6 +132,174 @@ Well done! Pat yourself on the back! The hard part is done, and it's time to mov
...
+
+#### Step 3.1: Update Gems
+
+Let’s make sure our gem files are up-to-date.
+
+~~~bash
+gem update
+~~~
+
+#### Step 3.2 Install Ruby Patch
```
I3uckwheat: This may be the ticket for making this work.

mindovermiles262:
I don't think we have the capacity to support this at the moment.

I think this would be an excellent blog post and we would happily re-post it (and maybe even make a small "additional link" to it.)

Other mods - what are your thoughts?

I3uckwheat:
Unfortunately, this is not something we're prepared to support here on The Odin Project.

PR #13898, opened 2019-12-30
corrected link for video

run-after:
The link that was originally on the page gave the overview video. The corrected link actually shows how to use the array acting as a queue


zachmmeyer:
👀

PR #13958, opened 2020-01-02
Adds UTF-8 charset meta to the basic HTML skeleton

Aodham:
If there is no charset meta, the web console logs an error. With this fix, you get a cleaner console.


105ron:
Thank you.

PR #13990, opened 2020-01-04
Updated section name from Chrome DevTools Documentation

xOC97:
In Web Development 101 > JavaScript Basics > Developer Tools 2 > Assignment > Elements Panel, we need to go over 2 sections called "CSS Reference" and "DOM - Get Started".

On Chrome DevTools there are no such sections but there are 2 sections called "Edit Styles" and "Edit the DOM". I am not sure if these are the same 2 sections only renamed, or if they are completely different from what the curriculum is currently pointing too.


105ron:
Looks good :100:

PR #14031, opened 2020-01-07
Add link to Rails installation instructions.

wmisback:
The instructions as is relate to the installation of Ruby alone. Those students who do not have Rails installed on their machine may not find the proper instructions to install Rails because of this. The proposed change fixes this problem, giving installation instructions for both Ruby and Rails.


105ron:
Looks good, thanks for the fix. :+1:

PR #14045, opened 2020-01-08
RawNoodles's solution for rock paper scissors

DiegoASandoval:

Thank you,

The Odin Project team.


105ron:
Hi @DiegoASandoval ,
you've deleted the file we use to render this page on the website. In this file you should be adding a line to link to the solution hosted in your Github repository. We don't host your solution within our repository, so the extra files added need to be removed from this PR. Thanks.

PR #14174, opened 2020-01-13
Update lesson_building_blocks.md

orionos-dei:
Replaced a outdated link in Assignment No. 5 with the latest official Ruby documentation which won't 404.


leila-alderman:
@orionos-dei Thanks for taking the time to update this URL! I tweaked it slightly so that it would jump directly to the relevant section.

Thanks for contributing to The Odin Project - we really appreciate it!

PR #14177, opened 2020-01-14
Fixed student's submission that was outside closing tag

martink-rsa:
A student's solution was outside of the `</details>` tag, causing it to appear outside of the expandable list.

Moved the student's solution so it is contained within the `<details> </details>` tags.


105ron:
Thanks for that. Looks like I missed that one :eyes:

PR #14223, opened 2020-01-15
Here's my solution

laz-k:

Thank you,

The Odin Project team.


105ron:
Hi, there is a merge conflict in this PR. Please either resolve it or allow the maintainers to push commits to your fork so we can resolve and merge this for you. Thank you.

laz-k:
I've allowed edits as I failed to resolve this. I would appreciate it if you could tell me what went wrong. I've just followed the video instructions on the contributing page. Thank you.

105ron:
Hi, someone has submitted a PR with a solution to this file after you have made your commit. Git then had two changes to the same line and it cannot merge automatically because it doesn't know which one it should use. In this instance we want to keep both but we need to do it manually. For some reason GitHub isn't letting me resolve it at the moment, it's something I'm looking into so I can fix it for you. In the meantime I'm asking students to resolve it but for those in the early stages of the curriculum it will be difficult.

laz-k:
Hi, thanks for your reply. That makes sense now. Can you let me know how could I have fixed it?
Thanks

105ron:
The is a very button to resolve it towards the bottom of the page or they give you line by line instructions to resolve it via the command line which is pretty good, but the GitHub GUI way is the easiest.

PR #14267, opened 2020-01-18
Update repl.it description

nearmint:
Someone in the discord was unable to do the repl.it exercises, and the solution was to first click "Take and Learn" and then not start the exercises from there, but instead, to go to the student dashboard and select the TOP classrooom from there. I was able to reproduce the issue. 

Therefore, I am suggesting an edit in the curriculum. 

By the way sorry about all the other commits, which I reverted. They were tests for https://github.com/TheOdinProject/theodinproject/pull/1055


105ron:
Thank you.

PR #14361, opened 2020-01-22
Improve LinkedList project documentation

unheavenlycreature:
Two small improvements:

- Some of the requested methods showed method signatures while others did not. I've updated the list of methods to show a signature for all of those that should take arguments.
- "Data" and "value" were being used interchangeably. I changed some of the language so we use "value" consistently.


CouchofTomato:
@unheavenlycreature Thanks. I do think this is clearer.

PR #14405, opened 2020-01-24
Remove HTML DOG add FreeCodeCamp

codyloyd:
... HTMLDOG is outdated.. FCC is not.

### DIFF
--- a/web_development_101/the_front_end/html_css_basics.md
+++ b/web_development_101/the_front_end/html_css_basics.md
@@ -30,11 +30,9 @@ Look through these now and then use them to test yourself after doing the assign
 
 <div class="lesson-content__panel" markdown="1">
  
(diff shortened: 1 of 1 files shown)


bycdiaz:
Looks good to me. 👍

105ron:
🤔

codyloyd:
@105ron have any more detailed thoughts on the matter?

105ron:
Is the thinking emoji not enough? I'll have a look through FCC and compare, then get back you you. I think it's probably a good idea.

codyloyd:
nice @105ron 

the big issue this should solve is that after doing HTMLDog almost everyone we get in the chatrooms lately have been trying to do the google homepage assignment with absolute positioning and floats... they know nothing about flexbox 😬

I3uckwheat:
Looks good to me, though we may want to suggest skipping the project at the end. Doing the google homepage will be redundant at that point. @codyloyd

PR #14534, opened 2020-01-30
Add data model to source explanation

coped:
This proposed addition is for [Ruby on Rails: Active Record Associations](https://www.theodinproject.com/courses/ruby-on-rails/lessons/active-record-associations#source), where the lesson introduces `:source`.

The lesson shows the reader three Active Record models involved in a `has_many :through` association. The associations in these models have various options/specifiers such as `foreign_key:`, `through:`, `source:`, etc.

I feel it's a bit hard to parse for a newcomer, and I thought this concept would be a little easier to understand if the reader was also provided the current database schema, so they could more easily understand which options/specifiers referred to which column/table.

There was also an extraneous colon `:` that I removed from one of the sentences in the lesson.

### DIFF
--- a/rails_programming/advanced_forms_and_activerecord/active_record_associations.md
+++ b/rails_programming/advanced_forms_and_activerecord/active_record_associations.md
@@ -96,31 +96,27 @@ For example, perhaps we change the example above so a Post actually can have mul
   end
 ~~~
 
-And to recap, here is the current state of our database schema (without the timestamp columns):
+And our data model looks like:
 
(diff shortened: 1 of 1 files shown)


coped:
I decided that it would probably be better to use markdown tables to show the current data model instead of shoving the db schema in the reader's face. This should make it much easier to understand what's going on.

It seems this syntax is supported by the markdown converter TOP uses: https://kramdown.gettalong.org/quickref.html#tables

PR #14582, opened 2020-02-02
Added solution to the Google homepage HTML/CSS assignment.

hypreme:

Thank you,

The Odin Project team.


105ron:
This is the wrong file for this solution? This is for a project later in the curriculum, not the Google homepage.

PR #14680, opened 2020-02-09
Add missing list item in the installation plan.

timlockridge:
The missing list item was removed in [this commit](https://github.com/TheOdinProject/curriculum/commit/be2369dd5606c70848bc08e911eff47f2c98299b#diff-50083222e59ba895ea5dc8026879e49a), but the curriculum still includes the rails app (and following paragraph still mentions the app), so I'm assuming this is a typo.

Thank you,

The Odin Project team.

### DIFF
--- a/web_development_101/installations/installation_overview.md
+++ b/web_development_101/installations/installation_overview.md
@@ -15,6 +15,7 @@ In the next few lessons, we will walk through these installation steps together:
 * installing the [operating system](https://en.wikipedia.org/wiki/Operating_system) (OS) of your choice;
 * installing a code editor;
 * creating an SSH key (a personal "password" that will identify you to GitHub, Heroku, and many other sites you'll be using); and
+* creating and deploying a very basic Ruby on Rails application to make sure it all works. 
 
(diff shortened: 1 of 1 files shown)


105ron:
I believe it was done on purpose because there is now an option for front end only, or full stack nodeJS and they won't be deploying a rails app.

timlockridge:
> I believe it was done on purpose because there is now an option for front end only, or full stack nodeJS and they won't be deploying a rails app.

I recently worked through the "Web Development 101" curriculum, which is part of all tracks (even the Full Stack JS). The "Installations" section of the curriculum still has the user install ruby and work through a Rails project, regardless of the track. I think that's what confused me about the typo in the list items here.

If you remove rails from that section, there's still a typo on this page: https://www.theodinproject.com/courses/web-development-101/lessons/installation-overview 

In the first unordered list under "The Installation Plan," the third list item ends with a semicolon & "and" implying that the sentence is missing a missing clause.

I hope that contextualizes my edits here. Sorry if I wasn't clear!

105ron:
It's ok. I am just not sure for @I3uckwheat reasons for removing it - whether it was in error or there was a reason for it. Hopefully he can shed some light on it.

I3uckwheat:
I think it was a mistake, I'm cool with it going back in

105ron:
Then it is back in. Thanks @timlockridge .

PR #14681, opened 2020-02-09
Edits for conciseness and clarity

timlockridge:
I edited a few sentences to simplify language or to clarify points.

### DIFF
--- a/web_development_101/git_basics/introduction_to_git.md
+++ b/web_development_101/git_basics/introduction_to_git.md
@@ -8,13 +8,13 @@ A *save* in a text editor records all of the words in a document as a single fil
 
 However, a *save* in Git records differences in the files and folders AND keeps a **historical record of each save**. This feature is a game changer. As an individual developer, Git enables you to review how your project grows and to easily look at or restore file states from the past. Once connected to a network, Git allows you to push your project to GitHub for sharing and collaborating with other developers.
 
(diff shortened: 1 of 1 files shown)


web_development_101/git_basics/introduction_to_git.md
```diff
@@ -37,12 +37,12 @@ By the end of this lesson, you should be able to do the following:
 </div>
 
 ### Additional Resources
-This section contains helpful links to other content. It isn't required, so consider it supplemental for if you need to dive deeper into something.
```
105ron: This is the standard additional resources wording for every page. I think it should be kept consistent across the site.

web_development_101/git_basics/introduction_to_git.md
```diff
@@ -37,12 +37,12 @@ By the end of this lesson, you should be able to do the following:
...
+This section contains helpful links to other content. Use these resources to dive deeper into Git and Github.
 
 * [Git and Github in plain English](https://blog.red-badger.com/blog/2016/11/29/gitgithub-in-plain-english)
 
 ### Knowledge Check
-This section contains questions for you to check your understanding of this lesson. If you're having trouble answering the questions below on your own, clicking the small arrow to the left of the question will reveal the answers.
+In this section you can check your understanding of the lesson. If you're having trouble answering a question, click the small arrow to the left of it to reveal the answer.
```
105ron: Again, this is a standard phrase used across the site. If we change it here, it needs to be changed on every other file. I think it is fine as it was. The other suggestions are good though. :+1:

web_development_101/git_basics/introduction_to_git.md
timlockridge: That makes sense. I restored the original language.

PR #14683, opened 2020-02-09
Removes extra fluff in lessons

I3uckwheat:
This should help get people to read the actual content and not stray too far off the path, such as installing Elementary or Mint


bycdiaz:
I like it.

PR #14684, opened 2020-02-09
Messy solution update

ARodrigues92:
This was meant to be an update to the rock-paper-scissors solution but i made a mess of it when i tried to pull from upstream. If there's stuff i need to fix, i would appreciate you telling me and possibly give some assistance.


105ron:
I didn't read your comment and just fixed it. There was a merge conflict. You should remove the conflict markers that git inserts to show you the two proposed changes that caused the conflict.

PR #14760, opened 2020-02-14
update student solution for restaurant-page project

mjwills-inf:

Thank you,

The Odin Project team.


mjwills-inf:
Sorry guys i'm not sure what I've done wrong this time around

105ron:
FYI, you have an older copy of the curriculum on your machine. You've added a solution on line 51 then tried to merge with with the master on GitHub. Someone else has added their solution on line 51 also, so git does not know which change it should use for this line. It will place markers in the file where the conflict is so we can fix it in a new merge commit, which I have done for you. It's a bit weird but it makes sense after a while. If you add this URL as `upstream` and pull from it before you add your next solution it will prevent this issue.

PR #14854, opened 2020-02-20
Updating the stack survey to 2019

langarus:
I've seen that there's a newer survey from stackoverflow as well


105ron:
Thanks :+1:

PR #14923, opened 2020-02-24
Google API Client Install Error

Cabonilla:
Modern ruby versions use a signet version not compatible with the Google API. Posted the small fix.

Thank you,

The Odin Project team.


105ron:
Thanks for adding this. Could you please make this all on the one line and let the browser automatically apply the wrap?

PR #15031, opened 2020-03-01
Reference the fact that the forums are down

zachmmeyer:
Please god augh


zachmmeyer:
I am bringing this in because people are all over coming in and asking what's up with the forums without doing research.

PR #15035, opened 2020-03-01
Added webpack-cli to instructions

BunnyTheLifeguard:
`webpack-cli` was missing in the instructions.
It was required when I ran `webpack --watch`.


105ron:
Thanks, I guess it's assumed you would install `webpack-cli` globally but it's more popular to install it for each repo. You can also sometimes use the `npx` command.

PR #15045, opened 2020-03-02
added fix for mongoose connection deprecation warning

highpockets:
Hello,

I faced an issue with the tutorial on this page. There was the deprecation warning: "current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor." It's fixed with the change.

Cheers,

Kris


105ron:
Thank you, :100:.

PR #15049, opened 2020-03-02
Fix Common Problem with Guest Additions Install

zachmmeyer:
Users do not immediately understand to replace $USER with their username.


105ron:
Thanks. I thought we had a standard to include variables with something like `<$USER>` syntax?

PR #15067, opened 2020-03-04
Added solution for project rock paper scissors

RilindaA:

Thank you,

The Odin Project team.


105ron:
HI @RilindaA ,
you've overwritten the project description with your solution. Could you please only add a line with your solution in this file, otherwise fellow students can't see the project outline, only your solution when they view this page. Thank you.

PR #15073, opened 2020-03-04
Modifying VBox Guest Additions to match majority

zachmmeyer:
This should reflect the majority of users installation experience and not one persons I suppose.


zachmmeyer:
I originally made the change to have the users change $USER to their username. I making it so it is not the default installation method now.

PR #15167, opened 2020-03-09
Change link to a complete version of the article

Lenn-e:
The Hackernoon article which was previously linked is missing code snippets. I replaced it with a link to the same article on Medium which doesn't have this issue.


105ron:
Thank you.

PR #15228, opened 2020-03-12
add dzarrr's solution

dzarrr:


dzarrr:
Sorry, my solution is wrong. It didn't generate the new grid in the same total space as before. Will work on it asap!

PR #15273, opened 2020-03-15
Virtualbox Lubuntu Refactor

zachmmeyer:
As shown in discord as was requested, the Virtualbox installation section has been reworked to include screenshots of each step. We have also replaced Xubuntu with Lubuntu as it is more similar to Windows for newcomers to Linux.

For whatever reason, the styling does not seem to take on the images. It works in my github markdown viewer on vscode. It might be something set for this repo.


zachmmeyer:
I think I'm done now.

PR #15546, opened 2020-03-30
Update resources for html/css best-practices

ZacTredger:
The link given for pesticide.io was broken. The site seems to no longer
exist. I linked instead to a replacement Chrome extension by Matthew
Baa, as well as the equivalent in Firefox.


105ron:
Thanks.

PR #15658, opened 2020-04-04
Replaces wait video

bycdiaz:
Previous video was made private. Replacing with a 10 minute compilation of jumping goats. Their energy is A1. Would recommend, 10/10.


105ron:
What about [this](https://www.youtube.com/watch?v=G1IbRujko-A). If you have a slow computer and it takes 10 hours this would be perfect.

PR #15798, opened 2020-04-09
update of links in async lesson following 404 now appearing

mjwills-inf:
links in async lesson for 'you dont know js' (assignment no. 3) now show 404. 

updated with links for 1st edition chapters.

thank you for everything guys.


105ron:
Thanks for the fix :+1:

Issue #13681, opened 2019-12-18
Unclear description @ javascript/introduction/js101-review.md


nearmint:
Hi all, 
the following page has an unclear direction, which should be fixed: 
https://github.com/TheOdinProject/curriculum/blob/master/javascript/introduction/js101-review.md

I'm talking about this sentence: "Running through the first part of either of those exercises is a great idea for a refresher on the syntax." 

It's unclear what the author means by "the first part of either of those exercises". Exercises may refer to the MDN and the Javascript.info link, but what does he mean by "the first part"? 

Looking at the Javascript.info link, it has a plethora of Javascript resources.

GraniteBeltWebAndSoftwareSolutions:
Maybe a better word would be keywords instead of syntax. I am thinking that is what is meant.

PR #15997, opened 2020-04-16
Update js101-review.md

I3uckwheat:
fixes: #13681


codyloyd:
based on our discussion in the chatroom I think the resources we should link here are:
MDN
FCC
learnxiny

javascript.info is nice.. but not well structured... FCC is good for a handholdy refresher

codyloyd:
i dig it

PR #16013, opened 2020-04-16
Place "RIght??" outside of link

laurenskrom:
Looks cleaner IMHO


I3uckwheat:
@laurenskrom Agreed, merged!

PR #16146, opened 2020-04-19
broken link for Assignment "Hello, World"

threeaces:
The original link gives "Internal Server Error"
https://www.learnrubyonline.org/en/Hello%2C_World%21 

New Link:
https://www.learnrubyonline.org/en/Hello_World


105ron:
Thank you.

PR #16267, opened 2020-04-22
"p" instead of "puts" in debugging example

threeaces:
updating to "p" from "puts" on second string example since it's supposed to show the difference between those two


105ron:
Thank you.

PR #16382, opened 2020-04-26
Add additional resource on events

aaron-contreras:
This Wes Bos video is a great explanation on Event capturing, bubbling and propagation. It's pretty short too and adds lots of value to one's understanding of events. I believe could save people hours of debugging. Perhaps the title I added could be better?


bycdiaz:
I think the video is great. Adjusting the link text so folks know what they are clicking on.

Issue #16404, opened 2020-04-26
Markdown for Introduction Header not displaying correctly in Google Clone Project (Web Dev 101)


minimalsm:
https://www.theodinproject.com/courses/web-development-101/lessons/html-css

<img width="847" alt="Screenshot 2020-04-26 at 22 22 13" src="https://user-images.githubusercontent.com/62268199/80320206-f0d99d80-880c-11ea-993e-bbce5c07dd13.png">

boyuan12:
Can I work on this issue?

aaron-contreras:
I'll take it

I3uckwheat:
@aaron-contreras Awesome, i'll assign this to you.

marvingay:
Issue resolved in #16406 

Closing.

PR #16406, opened 2020-04-26
Fix "Introduction" not showing as a heading

aaron-contreras:
#16404 The .md file was saved with "UTF-8 with BOM" encoding which was messing up with the heading. Saved it with "UTF-8" encoding.


minimalsm:
Thanks for explanation. I tried to fix and couldn't work out what the issue was 👍

PR #16511, opened 2020-04-29
Last link from page was dead.

JoBaHP:
just changed that one with similar link from another source.

Thank you,

The Odin Project team.


marvingay:

It helped determine the proper route to take when approaching the linked issue, which in this case was accessing the current resources via different method.

The issue is resolved. Closing.

Issue #16353, opened 2020-04-25
Small redundancy in Web 101 - DOM manuipulation additional resources


ChargrilledChook:
[DOM manipulation additional resources](https://www.theodinproject.com/courses/web-development-101/lessons/dom-manipulation#additional-resources) contains two separate links to the same page, [Plain Javascript.](https://plainjs.com/javascript/) 

It would make sense to consolidate them into a single dot point, unless there's a good reason not to.

![dom manuip](https://user-images.githubusercontent.com/58506115/80268866-49fedf80-86ee-11ea-9464-4c711ef39a86.png)

aaron-contreras:
I'll take it!

devmohy:
I will like to work on this

AJMcDee:
Dayummm you guys are too fast for me!

PR #16587, opened 2020-05-01
Merge resource items into one

aaron-contreras:
#16353 Kept the jQuery info since it might be an eye-catcher to those who already use it.


I3uckwheat:
@aaron-contreras Looks good to me! Thanks!

PR #16609, opened 2020-05-02
Spell check - (iteration, not iteraCtion)?

ryanfearly:
Hello - I am going through Ruby Basics (phenomenal so far!) and noticed something that might be a small typo. The page says iteraCtion, but I think it should be iteration.

But iteraction is a cool-sounding portmanteau


marvingay:
@ryanfearly  Good catch. Thank you so much!

PR #16642, opened 2020-05-03
Change pathfile in event_manager 

bboylen:
event_manager.rb is in the lib directory, while the event_attendees.csv file is in the root directory of event_manager. So the pathway must be changed to be able to read the file.


105ron:
Thanks. :+1:

PR #16690, opened 2020-05-04
Added additional resources to authentication

highpockets:
I found this article to be extremely helpful when trying to wrap my head around passport local strategy and session-based authentication. I had some requests to commit a PR to include it in the curriculum, so here it is. Thought it was best to just create an additional resources section for it. Cheers


marvingay:
Looks great!

I made some minor revisions to the verbiage and grammar used when describing the resource. In the absence of an additional resources section, this couldn't be a more fitting resource to add.

Thank you so much for your contribution!

PR #16710, opened 2020-05-05
pull request

milicen:

Thank you,

The Odin Project team.


marvingay:
@milicen Nice work!

In the future, be cautious of accidentally altering/changing the format of pre-existing solutions in the markdown. We don't want to break things!

milicen:
Thank you!! I'll be more cautious next time

PR #16771, opened 2020-05-06
updating second "submit" section with working link

Saul-Good-Homie:
second "submit" section did not have a hyperlink to the github page and references pair programming (which is good! but not explicitly relevant to this assignment).


marvingay:
Good catch. Verified that the pair programming reference wasn't relevant in surrounding lessons as well.

Thank you so much for your contribution/

Issue #16776, opened 2020-05-06
Code snippets not properly formatted in NodeJS mini message board project


kikupiku:
You can see from the markdown page of the [mini message board](https://www.theodinproject.com/courses/nodejs/lessons/mini-message-board?ref=lnav) lesson that certain things are supposed to be in bash code format (or javascript, etc.) but only the html snippets render properly. Strangely, they do show up correctly in the github preview.

ZachBaird:
@I3uckwheat I can take this.

I3uckwheat:
@ZachBaird Sweet, assigning it to you.

ZachBaird:
@I3uckwheat PR submitted.

kikupiku:
Thanks! The two last snippets still have the word bash that shouldn't be there, though, and the last snippet shows up as one long command instead in two lines.

ZachBaird:
Numbered list was still resetting and `bash` was getting printed out and was misformatted regardless. Thanks @kikupiku for checking this!

New PR submitted @I3uckwheat

PR #16811, opened 2020-05-07
16776 fixing unformatted code in markdown

ZachBaird:
See Issue #16776 

Second pass on markdown formatting. The bash code block was still printing "bash" so I changed it from backticks to tilde's. Numbered list was also still resetting so I indented the code blocks in an attempt to get markdown to understand the code is part of that numbered item.

[Mini Message Board](https://www.theodinproject.com/courses/nodejs/lessons/mini-message-board?ref=lnav)

Is `web: node ./bin/www` supposed to have "web" in it?
Edit: Note, I ask because I'm not familiar with this above command/code.


ZachBaird:
Tested the lessons locally with the change and it did not work. Back to the drawing board!

PR #16846, opened 2020-05-08
Update project_calculator

mohamednasr20:

Thank you,

The Odin Project team.


ryanford-dev:
Did you set up github pages for your solution? Your second link also goes to the repo, and I tried what *would* be the gh pages url and it 404s.

PR #16873, opened 2020-05-08
removing random </div> tag

Saul-Good-Homie:
Hopefully this doesn't break anything haha. There is/was a random '</div>' after the extra credit section so I just deleted it.


I3uckwheat:
That is the closing div for line 38

I3uckwheat:
That is the closing div for line 38

Saul-Good-Homie:
Gotcha, is there a way to make it not visible in the curriculum? 
Screen Shot 2020-05-09 at 10 03 09 AM

PR #16930, opened 2020-05-10
clarify instructions for learn.co tutorial

rlmoser99:
17 days ago I did this lesson & found that there was a "missing" link in the tutorial. I asked about it in Discord & got the link to the repo. It was suggested that I do a PR directly to learn.co to add this link to their repo, so I did. However, it is still sitting there as an open PR. 

After I did submitted the PR to learn.co, I did this tutorial, and I ran into another problem. The directions instruct you to "run `learn`  to see the failing test". Unless, I am missing something, `learn` does nothing (I think it might be a special command that learn.co includes in their set-up material). Instead, I found that I needed to run `rspec spec/pry_debugging_spec.rb` to run the test. 

Please, know that there is a chance that I am missing something from the previous lessons. I was partially through the Ruby material before it was recently updated. So, I only skimmed through the changed material.

### DIFF
--- a/ruby_programming/basic_ruby/debugging.md
+++ b/ruby_programming/basic_ruby/debugging.md
@@ -167,7 +167,7 @@ Obviously, if available, the stack trace is the first place you should look when
 
 1. Go through the Ruby Guides [Ruby Debugging](https://www.rubyguides.com/2015/07/ruby-debugging/) tutorial, which covers everything about debugging in more depth.
 2. Read through the [Exceptions and Stack Traces](https://launchschool.com/books/ruby/read/more_stuff#readingstacktraces) section of Launch School's online book *Introduction to Programming with Ruby*.
-3. Follow along with [this Pry tutorial](https://learn.co/lessons/debugging-with-pry) using [this repo](https://github.com/learn-co-students/debugging-with-pry-v-000) by Learn.co. During Part II, use `rspec spec/pry_debugging_spec.rb` instead of 'learn' to see the failing test.
+3. Follow along with [this Pry tutorial](https://learn.co/lessons/debugging-with-pry) using [this repository](https://github.com/learn-co-students/debugging-with-pry-v-000) by Learn.co. During Part II, use `rspec spec/pry_debugging_spec.rb` instead of `learn` to see the failing test.
 </div>
 
 ### Additional Resources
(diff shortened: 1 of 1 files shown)


marvingay:
@rlmoser99 

Can you attach the appropriate lesson (from TOP) that refers to this resource?

The `learn` command is something native to their setup (based on their site, as you've pointed out). 

If the instructions are from their website, unfortunately, it would be on them to make the appropriate changes. If it seems the project were to no longer be properly maintained, then a resource change would be taken into consideration.

rlmoser99:
The TOP lesson is assignment #3 at https://www.theodinproject.com/courses/ruby-programming/lessons/debugging

> If the instructions are from their website, unfortunately, it would be on them to make the appropriate changes.

I understand your point & I agree, except in this specific situation. The two things that I am suggesting to change, I believe, is outside the scope of learn.co's responsibility. Let me explain further...

As I was digging around on learn.co's website to figure out the correct command, I noticed that their lessons do not have links to their corresponding repo (for example: https://learn.co/lessons/method-scope-lab). Maybe this is because each lesson is the readme.md file for each repo. Whatever the reason, I do not believe that they will change their format, so "outside" students know where their repo is located. 

In a similar line of thinking, their native `learn` command does not make sense to TOP students. It does not make sense to have them change their lesson to accommodate "outside" students.

So, the bottom line is this resource either needs to be removed from TOP curriculum or needs both of these clarifications for TOP students to understand & use. However, I would hate to see if removed because I found it very helpful, once I figured out the "correct" command to run the test.

An alternative could be to just link the repo as the tutorial (instead of the article, which is the readme of the repo). Then either, clarify the non-learn.co command to run the rspec test or just suggest that they do part I. At this point in the current curriculum, I do not think testing for ruby has been covered, but they should have done testing for javascript in web dev 101.

marvingay:
Touching this PR so that it's not left in limbo.

Currently vetting if this issue comes up with a small sample of people so the most appropriate, relevant change, if any can be made.

Self-assigned.

marvingay:
@rlmoser99 Thank you for your contribution!

marvingay:
Verified that there seems to be a **consistent** issue with using the `learn` command within the scope of this tutorial.

As previously stated above, the necessary repository to use with the tutorial is absent.

Merging.

PR #16935, opened 2020-05-10
Removing duplicate paragraph

jordanperkins98:
Just removing a paragraph which is repeated twice.


marvingay:
@jordanperkins98 Thanks for your contribution!

PR #16942, opened 2020-05-10
Update project_linked_lists.md

urbanfog:
Deleting a <div> tag

### DIFF
--- a/ruby_programming/computer_science/project_linked_lists.md
+++ b/ruby_programming/computer_science/project_linked_lists.md
@@ -59,7 +59,6 @@ Build the following methods in your linked list class:
 
 1. `#insert_at(value, index)` that inserts the node with the provided `value` at the given `index`
 2. `#remove_at(index)` that removes the node at the given `index`. (You will need to update the links of your nodes in the list when you remove a node.)
-</div>
 
 ### Student Solutions
 


ZachBaird:
@urbanfog Hey! Could you provide more information on why this change needs to be made? Looking at the [live site](https://www.theodinproject.com/courses/ruby-programming/lessons/linked-lists) I don't see a reason a closing `div` unless I'm missing something.

urbanfog:
Hey Zach, I didn't see a reason for the closing tag either. Looks like it's just hanging there, maybe the site was previously converted from html? Not sure why it's there.

ZachBaird:
@urbanfog I took a look at the original markdown file and it looks like it closes the lesson content:
```md
  You will need two classes:

1. `LinkedList` class, which will represent the full list.
2. `Node` class, containing a `#value` method and a link to the `#next_node`, set both as `nil` by default.

Build the following methods in your linked list class:

1. `#append(value)` adds a new node containing `value` to the end of the list
2. `#prepend(value)` adds a new node containing `value` to the start of the list
3. `#size` returns the total number of nodes in the list
4. `#head` returns the first node in the list
5. `#tail` returns the last node in the list
6. `#at(index)` returns the node at the given `index`
7. `#pop` removes the last element from the list
8. `#contains?(value)` returns true if the passed in value is in the list and otherwise returns false.
9. `#find(value)` returns the index of the node containing value, or nil if not found.
10. `#to_s` represent your LinkedList objects as strings, so you can print them out and preview them in the console.
    The format should be: `( value ) -> ( value ) -> ( value ) -> nil`

### Extra Credit

1. `#insert_at(value, index)` that inserts the node with the provided `value` at the given `index`
2. `#remove_at(index)` that removes the node at the given `index`. (You will need to update the links of your nodes in the list when you remove a node.)
</div>
```

I noticed though that your original code is different from what we have, andI know there was a commit that put this div back in indentation. Were you seeing a trailing div on the page as content awhile back or pulled the repo in awhile?

urbanfog:
I was seeing a trailing div in the content. I just refreshed the page just now and I no longer see it.

PR #16945, opened 2020-05-10
Add an extra resource to 'intro to frameworks'

aaron-contreras:
I've made a PR on this before but ran into trouble with my fork and had to close it.

This adds a really helpful resource I found that clarifies and really nails down on what a framework is, especially on the first few sections of it. I believe it's worth being part of the assignment reads.


marvingay:
I remember this one! The resource supplements the other, nicely. Thank you for your contribution,  @aaron-contreras .

PR #16953, opened 2020-05-11
add solution

phankthusar:
I added a solution, but i also added some equal symbos and som numbers and letters, dont know if its normal or i pressed something i shouldnt.

Thank you,

The Odin Project team.


105ron:
> added some equal symbos and som numbers and letters, dont know if its normal or i pressed something i shouldnt.

It's normal. Git did that because of a merge conflict. You should be able to google, but it is confusing for the first couple of times it occurs.

PR #16995, opened 2020-05-11
adds extra steps to VirtualBox to avoid corruption

zachmmeyer:
This step to eject the VBox_GAs ISO supposedly helps avoid corruption of the VDI. This PR is to help troubleshoot the error users have been receiving. It is also good housekeeping.

Also removes some bad kramdown formatting and removes a redundant instruction.


zachmmeyer:
Users can eject before rebooting, but it would take more steps and possibly a picture. I have not been able to replicate the corruption with a single reboot with the ISO mounted.
If we get more corruptions after a certain amount of time has gone by for new people to follow these instructions, I'll fix it.

The people who have been having the corruption have been going more than a few days at a time, multiple reboots before the VDI corrupts. We have a fix posted in the #virtualbox-help channel on the Discord, but I am hoping my change here fixes it completely.

zachmmeyer:
I think this is good now.

PR #17066, opened 2020-05-13
Library solution

mortvicious:

Thank you,

The Odin Project team.

### DIFF
--- a/javascript/organizing-js/library-project.md
+++ b/javascript/organizing-js/library-project.md
@@ -38,8 +38,9 @@ Send us your solution so we can show others! Submit a link below to this [file](
 
 <details markdown="block">
   <summary> Show Student Solutions </summary>
+   
 - Add your solution below this line!
-- [mortvicious's Soultion](https://mortvicious.github.io/library/) - [View in Browser](https://mortvicious.github.io/library/)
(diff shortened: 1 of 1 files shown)


ZachBaird:
Hello @mortvicious ! I replaced the spacing your pull request was deleting - the space is necessary for the list to render properly in markdown. There was a type as well with "Solution" which I fixed.

Is the repo for your library project [here](https://github.com/mortvicious/library)? Both your links point to your live project, and the first one usually should point to the repo itself (so students can see your code).

mortvicious:
> Hello @mortvicious ! I replaced the spacing your pull request was deleting - the space is necessary for the list to render properly in markdown. There was a type as well with "Solution" which I fixed.
> 
> Is the repo for your library project [here](https://github.com/mortvicious/library)? Both your links point to your live project, and the first one usually should point to the repo itself (so students can see your code).

Okay, got it! 

Yes, it is there. Should I fix it by myself? I saw a few people with this way of links order (and thought it's a bit strange to put two same links with different labels, but I'm not the person who ruin rules) and since that moment started to always order in this way

ZachBaird:
@mortvicious cool, no worries! 

I can just change it here. If you see other students doing that, feel free to let us know :) It's more helpful for folks to see both live product and the code that built it.

PR #17126, opened 2020-05-14
Update library-project.md

stefank-29:

Thank you,

The Odin Project team.


marvingay:
@stefank-29 In the future, be sure that the markdown for your solution is consistent with other **solutions**.

PR #17143, opened 2020-05-15
Change in lesson outcomes

muhymenulhaque:
Since the curriculum has added JS track, we do need to install JS in our backend these days. So the lesson outcome needs to be changed as someone might get a little bit confused too. But its just a minor change.


marvingay:
@muhymenulhaque Thank you for your contribution!

Changed the verbiage to better accommodate the lesson itself. Other than that, good to go!

PR #17172, opened 2020-05-16
Added my solution to rock paper scissors solution 

bellabuchanan:

Thank you,

The Odin Project team.


marvingay:
@bellabuchanan Hello. The View in browser portion of your solution submission was removed because your `game()` function is never called and therefore doesn't work.

If you want to add it, after correcting the code, you may submit a new pull request.

PR #17180, opened 2020-05-16
Capitalize first words after question mark

aaron-contreras:
The first word after a sentence is always capitalized.


marvingay:
@aaron-contreras Thank you, as always.

PR #17182, opened 2020-05-17
typo

DeanWhitten:
At the bottom for additional resources it says "Angular have their own tutorial you can try" grammatically it should be " Angular HAS their own tutorial..."


105ron:
Well spotted. Thanks.

PR #17219, opened 2020-05-18
Update project_html_css.md

alishahwee:

Thank you,

The Odin Project team.


alishahwee:
Thank you so much for catching that. You were absolutely right; it was an auto-formatter. I have reverted the change and added the new change without additional formats. Let me know if there's anything else.

105ron:
All good now. Merged, thank you.

PR #17243, opened 2020-05-18
Nestor-C's solution submission 

nestor-c:

Thank you,

The Odin Project team.


marvingay:
@nestor-c  Thank you for your submission! Be cautious not to include __extra__ files in your solution submission commits. The only file that should be modified and contained in the PR is the curriculum file itself with the appropriate markdown.

Thanks again.

PR #17249, opened 2020-05-18
Sara's solution.

Sarajvega:
Please add my solution!


marvingay:
@Sarajvega  Will do.

Be careful to not include extra spaces in your markdown for links.
[Sara's solution] (https://www.google.com/) -- will not display properly.
[Sara's solution](https://www.google.com/) -- will. (no space between the bracket/parenthesis)

PR #17343, opened 2020-05-21
Add solution file for google homepage (Easy)

dev-opus:

Thank you,

The Odin Project team.


marvingay:
@dev-opus Thanks for your submission.

Due to how large this particular file is __and__ given that solution submissions should only be modifying one or two lines, this pull request will be closed. It's not necessarily practical to determine what changes were actually made in your commit to the solutions. Please re-submit making **only** the necessary change to provide your solution.

PR #17357, opened 2020-05-21
Suggestions for lesson improvement

fortypercenttitanium:
Added in two tips that had me caught up for awhile regarding assets and deploying to GH-Pages.


marvingay:
@fortypercenttitanium Thank you for your contribution.

The first suggestion was removed as it's more practical to allow the developer to determine how they would like to handle these assets.

Revised the verbiage and markdown on the second suggestion.

Thanks again!

fortypercenttitanium:
No problem. The first suggestion was a gotcha moment for me. I didn't
understand why webpack wasn't finding my assets when my live server version
was working fine.

PR #17379, opened 2020-05-21
Improve sentences' composition/aesthetics

loumarven:


marvingay:
@loumarven  Thank you for your contribution.

A few revisions were made (both additional and current), beyond that looks good!

PR #17418, opened 2020-05-22
Update library-project.md

salvillalon45:
Hello I been noticing that a lot of people have been leaving there firebase config public. They are not putting them in a .gitignore. I was wondering if this is okay for people to be doing this.

Thank you,

The Odin Project team.


105ron:
I think it is better practice to keep it in `.gitignore`. There are bots crawling GitHub looking for private keys, it's definitely better not to publish them IMO.

PR #17421, opened 2020-05-23
Update some misspelled words

izaiasneto4:
with in -> within
single page -> single-page

:smile:


marvingay:
@izaiasneto4 Thank you for your contribution!

Made some additional revisions that I caught. Thanks again.

PR #17446, opened 2020-05-23
added Khan Academy video to Additional Resources section of recursion…

Tenacious-Qi:
This Khan Academy video that breaks down the recursive Fibonacci method really helped me understand recursion. I think it would be beneficial to add it to Additional Resources.


marvingay:
@Tenacious-Qi  I quite liked the video.

Thank you for the contribution.

PR #17487, opened 2020-05-25
Fix learning outcome.

aaron-contreras:
The "any?" learning outcome was duplicated. Replaced it with "all?"


105ron:
Nice catch. :+1:

PR #17575, opened 2020-05-27
Update hieu-ng's solution

hieu-ng:

Thank you,

The Odin Project team.


marvingay:
@hieu-ng  The "live links" to both projects are "backwards". Clicking on the calculator live link takes me to a library and vice-versa.

Respond back if this is not an error on your end.

hieu-ng:
@marvingay Thank you very much for pointing that out! It was, indeed, an error on my end when I migrate files from the original repo to the new ones. I have rectified this, and the pages should be as intended.

marvingay:
@hieu-ng Thanks for the quick response! And very nice work!

PR #17581, opened 2020-05-27
miss use in any? method

feseka61:
used "numbers.any? { |number| item > 500 }" instead of "numbers.any? { |number| number > 500 }"
and also the other example has the same mistake


marvingay:
@bycdiaz  Can you take a look at this?

bycdiaz:
@marvingay This looks good to me.

@feseka61 Nice catch! Thank you for correcting this.

PR #17601, opened 2020-05-28
small fix for problem_solving.md

TrungNNg:
In line 74, I believe there is a missing "and" in the sentence ",print each number between one (and) the number the user entered,". It is a really small issue, but it quite confusing the first time I read it.


marvingay:
@TrungNNg  Good catch. Thank you for your contribution.

PR #17604, opened 2020-05-28
adds video from previous lesson as a step to watch

zachmmeyer:
People may not be understanding what a VM is and might be ignoring the last lesson section that has this video.


marvingay:
I personally think the Prereqisites section concerning this needs a minor re-write, wherein this video should be __required__.

I don't think it's practical to include a prereq in the steps to this lesson.

I will follow up on this later.

marvingay:
Addressed here -> #17926

Closing.

PR #17623, opened 2020-05-28
Clarify what "familiarize" means.

btdmaster:
I have found that the word "familiarize" is quite vague in the "Introduction to the Front End" lesson. Personally, I thought that the next lesson would involve a project. Because of this, I made the significant mistake of spending three days looking around MDN as if that was all the guidance I was getting. In retrospect, it made no sense given the next course title 'HTML and CSS Basics', but I really do not want this to happen to someone else.

Feel free to change the wording, but I hope the message is useful.


ZachBaird:
Hey @btdmaster , appreciate the suggestion! I agree familiarize can be vague, but I'm not sure the sentence you added further clarifies the paragraph a whole lot. Almost right before it:

"You will not understand most of what is there just yet, but bookmark the pages for later reference."

Maybe we should find a replacement for familiarize?

btdmaster:
From what I could find, few of its synonyms are any better - best I could find is 'get accustomed to'. Perhaps 'get an overview of what you can do with [HTML, CSS, JS]' might be more specific? I agree that my solution is not very useful, but I do not see an easy way out.

ZachBaird:
I think your second suggestion works :) Would you mind updating the PR?

btdmaster:
I'm not too sure how exactly I can edit this PR as I deleted my local branch. Should I link a new PR?

btdmaster:
I created a new PR, #17674. Sorry for the messy PRs, I need to get used to this.

PR #17639, opened 2020-05-28
Update dead link

Joewebsta:
I've updated a dead link. Thanks for considering my change.

Joe


marvingay:
@Joewebsta  As always, thank you for your contribution.

PR #17640, opened 2020-05-29
small update

killzabit:


marvingay:
@killzabit Good catch. Thank you for your contribution.

PR #17854, opened 2020-06-04
Add my easy version of the HTML/CSS solution.

emilravn:

Thank you,

The Odin Project team.


emilravn:
Curious whether I had to paste in the same format as previous additions. Or wondering if that would be handled automatically?

marvingay:
@emilravn  Yes, the markdown for solution submissions should be consistent with the others.

Thanks for your submission!

PR #17951, opened 2020-06-08
Add my solution to HTML/CSS

DAVIxD99:

Thank you,

The Odin Project team.


marvingay:
@DAVIxD99  This submission isn't consistent with the other solutions. The markdown should look similar to the other solutions in the file.

Make the appropriate changes and we can put your solution on display! Thanks!

PR #17971, opened 2020-06-08
Update fundamentals-2.md

tomas-vaclavek:
I've made a small change in ### Strings section, swapped the order of 1.) and 2.)
It makes a lot more sense to me this way, you build upon the skills you've learnt in the first lesson instead of having to google it all only to find that it's all explained well on the w3s tutorial. IMHO w3s should be studied before attempting the MDN tasks


marvingay:
MDN Documentation clarifies that it's "Test Your Skills" requires information from a subsequent article (and links to it) albeit discretely.

Closing.

(Discussed in our Discord community)

PR #18104, opened 2020-06-13
Add my solution to Etch-a-Sketch Project

pklepa:

Thank you,

The Odin Project team.


ZachBaird:
Hey @pklepa It looks like there's some extra hotness going on in your pull request. If you look under `Files changed`, it should only show one diff on the lesson page you're putting your submission on. Could you try and cut this down so that's the case?

PR #18248, opened 2020-06-17
add my solution to etch a sketch project

david8988:

Thank you,

The Odin Project team.


marvingay:
@david8988 Hello. It appears you have an extra file in your commit that needs to be removed. The merge markers also need to be removed from each of the markdown files before we can post your solution on display!

Once you make those changes we can get you set up!

marvingay:
Closing due to lack of response/change.

PR #18318, opened 2020-06-19
Tweak ES6 import note in testing-practice.md

brackenl:
Currently the first sentence of the 'Special Note on ES6 import statements with Jest' section in testing-practice.md doesn't read very well (it took me a few goes to understand what it was saying - I initially thought there were some words missing). 

I have proposed an amended version which is easier to read.


marvingay:
@brackenl  Seems good to me. Thank you for the contribution!

PR #18334, opened 2020-06-19
Typos and similar in the Angular lesson

scottBowles:
This PR fixes typos and very simple formatting inconsistencies in the Angular lesson.

Nothing bigger than a word or character here and there.


marvingay:
@scottBowles  As always, thank you for the contribution, Scott.

PR #18341, opened 2020-06-20
Update event_manager project.

lopezaxel:
The class Date doesn't have a #hour method, the #hour method it's from the DateTime class https://rubydoc.info/stdlib/date/DateTime#hour-instance_method. And here it says #hour to DateTime too: https://ruby-doc.org/stdlib-2.7.1/libdoc/date/rdoc/index.html.


105ron:
Thank you.

PR #18342, opened 2020-06-20
Added rbenv rehash for users with rbenv installed

bharat-patodi:
Also, as a fail safe, shouldn't users be checking for a correct installation with the 'rails -v' command?

Thank you,

The Odin Project team.

### DIFF
--- a/rails_programming/introduction/installing_rails.md
+++ b/rails_programming/introduction/installing_rails.md
@@ -35,11 +35,13 @@ We've previously installed Ruby, and now it's time to install Ruby on Rails. It'
 gem install rails -v 5.2.3
 ~~~
 
-Although, if you already have rbenv installed on your system, you'll need to run the following command:
+Once the installation finishes, you can check if everything went well by running the following command:
 
 ~~~bash
-rbenv rehash
+rails -v
 ~~~
+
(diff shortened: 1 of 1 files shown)


ZachBaird:
Hey @bharat-patodi , it doesn't look like `rbenv rehash` is necessary anymore. See [here](https://github.com/rbenv/rbenv/issues/967#issuecomment-277507719) for more detail.

That being said, I think having students check their Rails version with `rails -v` would be a great addition! Would you mind updating your PR to adding that rather than the `rbenv hash`?

rails_programming/introduction/installing_rails.md
```diff
@@ -35,6 +35,14 @@ We've previously installed Ruby, and now it's time to install Ruby on Rails. It'
...
 ~~~
 
+Once the installation finishes, you can check if everything went well by running the following command:
+
+~~~bash
+rails -v
+~~~
+
+This should display the version of Rails installed on your system indicating the installation went smoothly.
+
```
bharat-patodi, line 45: Made the changes by removing rbenv rehash and introducing the call to rails -v.

ZachBaird: Good stuff :)

PR #18355, opened 2020-06-20
Add explanation that addEventListener uses callbacks

EmilioAK:
I was super confused when I saw that function(e) just worked somehow, and wondered where e was passed in. I got it explained to me on discord that e was passed in by addEventListener because it was defined to try to pass in arguments to the second input. Just seeing this without explanation made me confused, so I think something along these lines should be added.

### DIFF
--- a/web_development_101/javascript_basics/DOM-manipulation.md
+++ b/web_development_101/javascript_basics/DOM-manipulation.md
@@ -407,7 +407,7 @@ btn.addEventListener('click', function (e) {
   console.log(e);
 });
 ~~~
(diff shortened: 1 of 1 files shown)


ZachBaird:
Hey @EmilioAK, the information you wrote isn't quite right unfortunately. A callback is a **function** that is passed into another function as an **argument**. The `e` parameter is an object & argument being passed into that callback function.

However, this is a great resource on callbacks! If you can, please feel free to edit your stuff so it isn't saying `e` is a callback and we can get this merged into the curriculum :)

ZachBaird:
I added a period for punctuation and just made sure that the note says that `function (e)` is a callback and not `e` itself. Otherwise, this is very useful :)

PR #18391, opened 2020-06-22
Express messageboard project -- correct app.[method]s to router.[method]s

scottBowles:
Per conversation [here](https://discordapp.com/channels/505093832157691914/597796701650026517/702163935385878588). If you're wondering what took so long, I was waiting for my earlier formatting pr to go through first, since they seemed like issues that should be addressed separately. Suffice it to say, I didn't anticipate how drawn out the earlier pr would be.

Repl.it you can play with [here](https://repl.it/@ScottBowles/Express-Mini-Messageboard). Use this rather than the repl.it I linked in the discord conversation, as this has the whole app uploaded.

In the routes/index.js file, it currently says to use app.get() and app.post(). Unless I'm missing something, it should be router.get() and router.post().

When I was going through the project, I had to make this change to make it work. I looked at a number of other student solutions and they all seem to do the same.


ZachBaird:
I went over the instructions and your changes definitely seem in line with what the express-generator builds up. Thanks for shooting up this pull request :) Keep up the good work man!

PR #18463, opened 2020-06-24
added solution

benjamin-gambling:

Thank you,

The Odin Project team.


ZachBaird:
Hey @benjamin-gambling , looks really good! I like the design you put into this :) the only thing I noticed was that your Add Book modal doesn't let you scroll down to the Save Book button. It might be my resolution or something, but it was off-screen for me and I couldn't scroll down to click it. I had to zoom out to see it. 

Aside from that, great work :)

PR #18563, opened 2020-06-26
added information to show that Firefox supports subgrids 

RyanAmeri:
Slight modification to show that Firefox now supports subgrids but other browsers still dont. 
https://caniuse.com/#feat=css-subgrid


marvingay:
@RyanAmeri  Thank you for the contribution, Ryan.

PR #18569, opened 2020-06-26
Added Two videos to the additional Resources

Spayco:
in this PR i added two videos to the additional resources, as suggested by one of the Discord members, 
Dev Ed is a great YouTuber, and his CSS Grid and FlexBox helped me a lot in learning, so i think this should be here


ZachBaird:
I love DevEd! Thanks recommending this :)

PR #18591, opened 2020-06-27
Removed e parameter from the second code block in the Event subsection of DOM manipultaion lesson under ‘attaching listeners to groups of nodes'.

rahimratnani:
Hi there,

I'm using The Odin Project as my primary learning for about 2 months now. I just came around a typo or I think it’s a typo in Web Development 101 JavaScript basics section. In the Event subsection of the DOM Manipulation lesson under ‘attaching listeners to groups of nodes’, there’s a second code block describing how to use forEach method on a nodelist. I think there’s no need to pass ‘e’ as a parameter to the callback function as we’re accessing the button object’s id property, not e’s. I’ve removed e parameter from the code in DOM-manipulation.md file.

Thank you.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,5 @@
-This is a PR template. If you are adding a solution link to the curriculum, leave this as is. If not, delete it and write the message you wish.
+Hi there,
 
-Thank you,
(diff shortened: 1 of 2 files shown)


marvingay:
@rahimratnani  Thanks for the contribution! You're right, the `e` isn't necessary and given that the use of the event object is covered, there's no reason to include it here.

Be careful to only commit the files that require changes. In this case, the Pull Request template was modified and committed, but it was restored.

Thanks again!

rahimratnani:
Well, Thank you so much. This was my first contribution to an open source project.

PR #18598, opened 2020-06-27
"What is pseudocode " Answer isn't found in main section

LrdVnz:
The answer is found in the optional section


marvingay:
@LrdVnz  The answer is in the first video. But the added resource is fine, merging.

PR #18599, opened 2020-06-27
Added Coding Shawn's solution for ruby hangman project

CodingShawn:
Added solution link for ruby hangman and also removed line about adding partner's github handle in project_file_io.md as it was not applicable for that page


CodingShawn:
Also removed line about adding partner's github handle in project_file_io.md as not applicable

Issue #17956, opened 2020-06-08
Binary Search Tree Project Suggestions


andrewjh271:
Lesson: https://www.theodinproject.com/courses/ruby-programming/lessons/data-structures-and-algorithms

I recently finished the Binary Search Tree project in Ruby, and I have a number of suggestions I wanted to put out there, in case they’re helpful.

Step 3. I wonder if it could be more explicit that making a balanced tree with the `build_tree` method is possible because the array it builds from will be sorted. I was confused here for a while, because the curriculum doesn’t mention building balanced trees from sorted arrays. At first I was thinking we were supposed to build a self-balancing tree. I also feel it could be worth clarifying that in the situation where you’re inserting one value at a time, a sorted array would produce the worst case scenario of a linked-list.

Step 4. Regarding the `delete` method: this is the first time in the TOP curriculum I felt like I was being asked to do something non-trivial that wasn’t ever mentioned before. Of course it’s easy enough to google, and lots of students figured it out (I did too), but still — I felt like this wasn’t in the style I’d encountered in the past, and it threw me off. Maybe there could be a link or two to some helpful resources.

Step 10. According to the Ruby Style Guide, an exclamation mark belongs at the end of methods for which a safe version exists:

```
# bad - there is no matching 'safe' method
class Person
  def update!
  end
end

# good
class Person
  def update
  end
end
```
My understanding is that there should not be an exclamation mark in `rebalance`.

Regarding the metaprogramming bonus section, I found these resources very helpful, and not what comes up first in google searches. Clues like using `send`, and that `yield` doesn’t work inside a `define_method` block would make this section a little more friendly. 
http://rubylearning.com/blog/2010/11/23/dont-know-metaprogramming-in-ruby/
http://rubymonk.com/learning/books/2-metaprogramming-ruby/chapters/25-dynamic-methods/lessons/65-send#solution4186
https://stackoverflow.com/questions/2306731/using-yield-inside-define-method-in-ruby

rlmoser99:
Since I am working on the other chore for this lesson, I'll tackle this one too... I have a few questions.

I did not find step 3 confusing, but I can try to clarify it. However, I am confused by @andrewjh271 statement "I also feel it could be worth clarifying that in the situation where you’re inserting one value at a time, a sorted array would produce the worst case scenario of a linked-list."

[...]

PR #18638, opened 2020-06-28
changed definition of step 7 from 'depth' to more accurate 'height'

Tenacious-Qi:
Instructions originally defined depth of a node with the definition for height of a node. I changed it to height. This should prevent some confusion, especially since step 8 (correctly) references _heights_ and not depths of left and right subtree. 

For reference here is related open issue (see comments) : https://github.com/TheOdinProject/curriculum/issues/17956

and this Stack Overflow post: https://stackoverflow.com/questions/2603692/what-is-the-difference-between-tree-depth-and-height#:~:text=The%20depth%20of%20a%20node%20M%20in%20the%20tree%20is,and%20its%20depth%20is%200.


Tenacious-Qi:
also here is the video that clarified it for me: https://www.youtube.com/watch?v=_pnqMz5nrRs

ruby_programming/computer_science/project_data_structures_algorithms.md
```diff
@@ -24,7 +24,7 @@ You'll build a simple binary search tree in this assignment. In this lesson, our
 
   6. Write `#inorder`, `#preorder`, and `#postorder` methods which accept a block. Each method should traverse the tree in their respective depth-first order and yield each node to the provided block. As a bonus, make the method return an array of values if no block is given.
 
-  7. Write a `#depth` method which accepts a node and returns the depth(number of levels) beneath the node.
+  7. Write a `#height` method which accepts a node and returns the height of the node. Height is defined as the number of edges in longest path from a given node to a leaf node.
```
leosoaivan, line 27: Grammar-wise, can we change the second part of the first phrase to `and returns its height`? We're already referring to `node` in the first half, so this reduces redundancy.

Otherwise LGTM. Thanks @Tenacious-Qi

Drew-Goddyn: I personally find this more confusing, I always use the root node as a reference point and almost think the tree spreading downwards (depth). In which case the original definition I think was just missing the word "root): 

```
Write a `#depth` method which accepts a node and returns the depth(number of levels) beneath the root node.
```

Per your linked stacked overflow, depth and height are just the inverse of each other depending on whether you're talking about root node to a leaf or a leaf node to root so neither is more correct. Maybe we should mention both since I would find thinking of height as more confusing than depth?

Drew-Goddyn: Maybe we should have a method for both depth and height since they're just inverses depending on how you view tree.

Tenacious-Qi: @leosoaivan Thanks that is better that way, new commit submitted. I had weird things happen when I tried to pull upstream master, so there is an extra commit in there with no changes.

Tenacious-Qi: @Drew-Goddyn good idea. Should that be a separate PR? I'm not really sure. My main concern was that step 8 mentions depth and then #balanced? in step 9 mentions height. That inconsistency is confusing if you don't yet know depth and height are inverses. But, as you said, having both methods required could help clarify the difference and then students could decide which they wanted to use in their code.

Drew-Goddyn: @Tenacious-Qi sorry for delay. It can be in a followup pr if you prefer since this isn't _technically_ incorrect

Drew-Goddyn:
@Tenacious-Qi can you resolve the merge conflict and I'll approve

Tenacious-Qi:
@Drew-Goddyn I think I did it correctly, I just pulled and then committed a couple more times. Should be good to go I think.

PR #18672, opened 2020-06-29
Change ../event_attendees.csv to event_attendees.csv in one instance …

msespos:
In all other instances "event_attendees.csv" is used, so the first one (which was "../event_attendees.csv") is changed to keep it consistent with the others.


marvingay:
@msespos Thank you for the contribution.

PR #18772, opened 2020-07-02
updated 2019 stackoverflow survey to 2020

jooji-san:


marvingay:
@jooji-san  Thank you for the contribution.

PR #18796, opened 2020-07-03
Join-Startups have moved to underdog.io

langarus:
So the previous website doesn't work, and I checked it out on waybackmachine and it showed that they were moving. It was last available in 2019
https://web.archive.org/web/20181204173328/https://join-startups.com/


marvingay:
@langarus Thank you for the contribution.

PR #18797, opened 2020-07-03
Name updated on CWJobs

langarus:
We should add an update to the name, because from what I saw they had no remote jobs, and all were just for UK


marvingay:
@langarus This addition isn't necessary.

PR #18822, opened 2020-07-03
Add project and fix link

Lucas-Bide:

Thank you,

The Odin Project team.


marvingay:
@Lucas-Bide Hello. The original link isn't broken and goes to the intended location. That change was reverted.

The link to your repository is incorrect in the markdown. If you could change it to the appropriate link, we can then merge your solution, thank you!

PR #18837, opened 2020-07-04
add npm instruction and another way to create project

jooji-san:


105ron:
Thank you. I hate `yarn`.

PR #18883, opened 2020-07-05
Add a tutorial to improving the git experience in terminal (MacOS and Ubuntu ).

vishalagrawal22:
I have added the tutorial to configure the terminal on MacOS and Linux (Ubuntu) for better git experience


marvingay:
I'm of the opinion that this sort of customization should be left up to the learner, on their own accord. This creates something superficial for us to support, while only providing a marginal benefit to the learner (and a potential point of failure in the setup process).

I do think this would make a great blog post or article, though. I'll have other eyes look at this to see what they think.

I3uckwheat:
on top of what @marvingay said, complicating the setup process is a good way to add more typo, broken installs, and frustration to the user.

marvingay:
@vishalagrawal22 Including the comments here and further discussion, this pull request will be closed.

Thank you for putting this forward, and to reiterate, an article or post on customization is useful (so do consider making it public for yourself after revision), it just opens up more potential problems than it helps in the context of our curriculum.

Thanks again.

PR #18903, opened 2020-07-06
Improve code style slightly for JS Problem Solving Introduction

chrisjwelly:
I changed `==` in `i % 5 == 0`to `===` in order to be consistent with the rest of the text. One of the paragraphs after this even mentioned:

> With the condition *`if (i % 3 === 0 && i % 5 === 0)`* coming first, we ensure we check that `i` is divisible by both 3 and 5 before moving on to check if it is divisible by 3 or 5 individually in the `else if` conditions.

While it is not very obvious, I hope that new JavaScript users will not be bothered about the difference with the usage of `==` and `===` this early on


marvingay:
@chrisjwelly  Thank you for the contribution.

PR #18978, opened 2020-07-08
Fix typo

Cyber-Mat:

Thank you,

The Odin Project team.


marvingay:
@Cyber-Mat  In the future, solution submissions should not be combined with other contributions unless they are in the same file.

Cyber-Mat:
Alright, thanks. I'm still trying to understand how it all works

PR #19014, opened 2020-07-09
Clarify order of operations policy in WD101 calculator project and ad…

msespos:
…d a new example website

This is a clarification of the order of operations policy for the WD101 calculator project. The previous version of the instructions linked to a calculator that used order of operations, and did not specify explicitly whether or not order of operations would be required. This version links to a calculator that does not use order of operations, and states that order of operations is not required, giving an example with solution, as well as stating that order of operations can be implemented, optionally.


marvingay:
@msespos Looks good. Bolded a couple of key words; changes will be reflected tomorrow. Thank you for the contribution.

Issue #19019, opened 2020-07-09
Joshua Kemp's Link on Gearing Up doesn't work anymore


zachmmeyer:
https://github.com/TheOdinProject/curriculum/edit/master/web_development_101/the_basics/gearing_up.md
http://joshuakemp.blogspot.co.uk/2013/11/how-blacksmith-learned-to-code-and-9.html

He redirects away from that site. You can still get to it if you stop the page loading.
I am going to bug him on twitter to see if he'll allow access to that blog post or not.

marvingay:
Will allow a week turnover before closing the issue.

marvingay:
@zachmmeyer Did he ever respond to you regarding this issue?

zachmmeyer:
@marvingay no. He's currently MIA from the internet.

PR #19020, opened 2020-07-09
Comment the blacksmith link out while it does not work

zachmmeyer:
#19019


marvingay:
@zachmmeyer Thanks for the contribution, Zach.

PR #19028, opened 2020-07-09
Update responsive_design.md

tomigm:
I find it interesting to add like an extra resource, he proposes writing the phone layout first, and then, desktop.

Greetings!


marvingay:
@tomigm Thank you for the contribution.

PR #19035, opened 2020-07-10
Update project_apis.md

leetie:
Hey I don't know if this is unique to me because I've never used flikr and didn't have an existing account prior to this assignment, but it appears that as of the time of this PR when you make an account your Flikr ID can be seen in the URL for your profile rather than using the idgettr service. I'm guessing this might not be the case with older accounts, so if any of you with older flikr accounts can confirm that the part of the path that is an ID for me might be a username for you, feel free to decline the PR or maybe add a line about using idgettr if you have an older account.


ZachBaird:
Hey @KevinMulhern , I'm not familiar with this project. Can you take a look at this?

leetie:
Hey so it turns out that older profiles do have just a string for an identifier in the url, but someone making a Flickr account for the purposes of this project can see their id on their page without having to use any third party service

marvingay:
@leetie  I tested both sides of it. I revised the instructions to be more formal and easier to follow, with the addition of a format example of the ID.

PR #19111, opened 2020-07-12
Fixes typo

aaron-contreras:
"such as a tests" should be "such as tests"


marvingay:
@aaron-contreras Thank you for the contribution.

PR #19112, opened 2020-07-12
Rewrite text to make it easier to understand

lopezaxel:
I think "Write a program" is easier to understand than "Put together a script".


marvingay:
@lopezaxel Hello. Thanks for the submission. The change itself is redundant, given the usage of "script" is appropriate in this context.

PR #19113, opened 2020-07-12
Update installation_overview.md

dhatGuy:
I have once used CodeSandBox and I feel it should be recommended here too. It's similar to vscode

Thank you,

The Odin Project team.


marvingay:
@dhatGuy  Yeah, we tend to recommend it in our Discord community when it comes to projects that use a frontend framework.

Thank you for the contribution.

PR #19115, opened 2020-07-12
Fixes typo

aaron-contreras:
Throughout the lessons the phrase it's quoted as "red-green-refactor" cycle


105ron:
You are correct. It didn't make sense like it was. Thank you.

PR #19306, opened 2020-07-18
Update factory-functions.md

DoviMaj:
I think reads better without as is

not a grammar expert thou
https://www.reddit.com/r/grammar/comments/2bwvjs/as_much_as_is_possible_or_as_much_as_possible/

what do you think?


marvingay:
It's grammatically correct, as is.

PR #19320, opened 2020-07-19
Update tic-tac-toe-project.md

miknick:

Thank you,

The Odin Project team.


marvingay:
@miknick  Hello. Your submission isn't consistent with the other solution markdown. Your links are present but you need to format them in the same fashion as the solutions immediately below where you pasted your links. Once that's done we can display your solution! Thank you!

PR #19361, opened 2020-07-20
Change line 100 on 'No such file or directory'

kuz3:
Found that user on Discord was getting confused where "No such file or directory" was written twice. I restructured this paragraph in an attempt to make it less confusing.


marvingay:
@kuz3  Thank you for the contribution.

PR #19371, opened 2020-07-21
define REST acronym

ChargrilledChook:
A whole section is dedicated to REST in the Rails Web Fresher module but the term is never directly defined.

Small change to give slightly more context to the acronym / jargon being used.


105ron:
Great idea. Thank you!

PR #19380, opened 2020-07-21
Keeping in-line with ruby convention

aaron-contreras:
Words in Ruby method names can't be separated by a hyphen. Updated it to follow the convention of using an underscore.


bycdiaz:
@aaron-contreras, nice catch. Thanks for updating this.

Issue #18528, opened 2020-06-25
Pseudo code in ruby problem solving lesson is formatted poorly


ChargrilledChook:
The pseudo code examples in this lesson appear as blocks of text on the website and are hard to read eg

![odin-site-pcode1](https://user-images.githubusercontent.com/58506115/85695022-fffc9f00-b71a-11ea-83d0-1524e092a5e5.png)

![odin-site-pcode2](https://user-images.githubusercontent.com/58506115/85695045-0428bc80-b71b-11ea-8f2f-fe971de0590a.png)
***
However the same examples look good on the github page eg

![odin-repo-pcode1](https://user-images.githubusercontent.com/58506115/85695267-3803e200-b71b-11ea-996e-e07177e202cf.png)

![odin-repo-pcode2](https://user-images.githubusercontent.com/58506115/85695278-3afed280-b71b-11ea-9b34-35375b166c55.png)
***
These examples would be much more effective / easy to read if the live site looked like the examples from the repo.

[Live page](https://theodinproject.com/courses/ruby-programming/lessons/problem-solving)

[On Github](https://github.com/TheOdinProject/curriculum/edit/master/ruby_programming/basic_ruby/problem_solving.md)

dm-murphy:
@marvingay I noticed this issue after fixing the backticks for this page. I can change this as well if that's alright. I think separating each pseudo code line onto its own line might be sufficient and more readable. That way it also keeps it distinguished from the full code blocks.

marvingay:
@dm-murphy  Go for it.

dm-murphy:
@marvingay @ChargrilledChook  

I believe I have resolved this issue. The 3 backticks for the pseudo code blocks were replaced with 3 tildes and that separated the lines. I have also submitted the same fix to the JavaScript Problem Solving page along with the rest of this page's editorial fixes so once that PR is merged the pseudo code blocks should have line spacing on both pages. 

It should be noted the text color is now black, not magenta. Don't know if that is an issue for anyone but it appears readable on my end and is a different font than the rest of the body text to distinguish the pseudo code from the paragraphs.

marvingay:
Will follow up on what the page looks like now that it's merged.

marvingay:
@dm-murphy  THe text color is quite trivial. I am in agreement in terms of it not affecting the readability, as it is in a code block regardless.

Thanks again for your contribution.

PR #19417, opened 2020-07-22
Update problem_solving.md

dm-murphy:
Trying to resolve issue #18528 (Pseudo code in ruby problem solving lesson is formatted poorly)

The 2 sections of pseudo code here both render correctly in the Git preview with appropriate line breaks, but on the actual TOP page they do not show any line breaks. I replaced the use of 3 backticks with 3 tildes instead to see if that will fix it. If that doesn't work I could also try using 4 spaces to indent each line or use 2 carriage returns to force the line break.

### DIFF
--- a/ruby_programming/basic_ruby/problem_solving.md
+++ b/ruby_programming/basic_ruby/problem_solving.md
@@ -45,12 +45,12 @@ Pseudo code is writing out the logic for your program in natural language instea
 
 Here's an example of what the pseudo code for a simple program that prints all numbers up to a inputted number might look like:
 
-```
+~~~
 When the user inputs a number
 Initialize a counter variable and set its value to zero
 While counter is smaller than user inputted number
 Print the value of the counter variable
-```
+~~~
 
 This is a very simple program to demonstrate how pseudo code looks, there will be more examples of pseudo code included in the assignments.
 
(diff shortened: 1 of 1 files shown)


ZachBaird:
Hey @dm-murphy , there's a similar issue over on [The Javascript version as well](https://www.theodinproject.com/courses/javascript/lessons/problem-solving-javascript). Do you want to add that to this PR to see if it fixes both?

dm-murphy:
Hey @ZachBaird, thanks for pointing that out. Yes I will go ahead and make the same change on the JS version from 3 backticks to 3 tildes and we'll see how that goes. For the Ruby page I had already fixed a number of typos, grammatical and clarity errors. And @marvingay had me change the output terms for Fizz, Buzz and FizzBuzz from double quotes to backticks. I can go ahead and make all those same changes manually on the JS page, unless there is an easier way to migrate the same fixes from the new Ruby page over that I'm not sure about. I'm still learning the Git process here. Thanks!

ZachBaird:
@dm-murphy I'm afraid it'll have to be manual :\ Thanks for contributing!

dm-murphy:
@ZachBaird No worries and happy to help! I'll go ahead and make all the JS page changes now and then we can see how the new code blocks turn out.

PR #19434, opened 2020-07-23
remove future reference to asset pipeline

ChargrilledChook:
In the rails deployment lesson, the line 'We'll cover the Asset Pipeline in a future lesson...' doesn't make sense, as this lesson immediately follows the asset pipeline lesson.


105ron:
Thank you.

PR #19479, opened 2020-07-24
Delete sentence talking about lessons that don't exist

Adub648:
These lessons now do not exist in the current Web Development 101 curriculum and should therefore be deleted.


marvingay:
@Adub648  Thank you for the contribution.

PR #19558, opened 2020-07-26
Proposal to change the assignment

mattgcc:
Hello,

I watched the both videos. The content is the same, but the video I propose is much shorter (because the information is condensed and presented in a scripted manner) and more helpful. The end of the video contains a practice section (which back then would net you an award - sticker), it isn't relevant but it provides a visual representation of how real git usage would look like. 

Proposed video: https://youtu.be/HkdAHXoRtos
Current video: https://www.youtube.com/watch?v=HVsySz-h9r4

Matt,

### DIFF
--- a/web_development_101/git_basics/git_basics.md
+++ b/web_development_101/git_basics/git_basics.md
@@ -15,7 +15,7 @@ By the end of this lesson, you should be able to do the following:
 
 <div class="lesson-content__panel" markdown="1">
 
-  1. Watch [this video](https://www.youtube.com/watch?v=HVsySz-h9r4) by Corey Schafer for a great overview of some basic Git commands.
+  1. Watch [this video](https://youtu.be/HkdAHXoRtos) by Jeff Delaney for a great overview of some basic Git commands.
 
 </div>
 


marvingay:
@mattgcc A few questions:
1. How can the content be the same when one video is 2.5x the length of the other?
2. What advantages does the proposed video offer over the current video? (for example, the usage of VSCode)?
3. The current video is presented in a similar "code along" format. How is the practice section in the proposed video different?
4. Do the advantages outweigh the cons? (For example, the front half of the proposed video contains a lot of anecdotes about Git, such as different flags, options, and other verbiage/terminology that potentially raise questions by a newer learner that aren't explicitly answered in the video.)

Thank you for the proposal.

I3uckwheat:
It's important to note that there's a reason the current video has over a million views, and the proposed video has roughly 1/10th that. 

I don't think we should replace this video, but adding the suggested video in the "additional resources" section would be a good idea. I would recommend making a PR to add the video you proposed to the "additional resources" section.

web_development_101/git_basics/git_basics.md
```diff
@@ -15,7 +15,7 @@ By the end of this lesson, you should be able to do the following:
 
 <div class="lesson-content__panel" markdown="1">
 
-  1. Watch [this video](https://www.youtube.com/watch?v=HVsySz-h9r4) by Corey Schafer for a great overview of some basic Git commands.
+  1. Watch [this video](https://youtu.be/HkdAHXoRtos) by Jeff Delaney for a great overview of some basic Git commands.
```
I3uckwheat, line 18: This does not need to be replaced, please add it to an "additional resources" section.

marvingay:
This is corrected in #19697

PR #19589, opened 2020-07-27
Just fixed a small typo

Korak-997:
the word too , i found it wrong there . while in that position should be to instead of too.


marvingay:
@Korak-997  Good catch, thank you for the contribution.

Korak-997:
Thanks for you guys , for providing us that super amazing Curriculum :)

PR #19678, opened 2020-07-30
fix dead link

jooji-san:


marvingay:
@jooji-san Thank you for the contribution.

PR #19681, opened 2020-07-30
Changed download path for csv samples

Northrain-cmd:
For the code provided to work the samples have to be in the same directory as the 'event_manager.rb' file. If we save csvs in the root folder, then we have to add ' ../ ' to every code sample which uses paths in the lesson


rlmoser99:
Thank you for taking the time to submit a pull request. However, navigating the file structure is an important concept to understand. In addition, as you progress to using more complicated file structures, you will learn that downloaded files, such as this csv, do not believe in the lib folder. This pull request will be closed for convenience. Thanks again.

PR #19692, opened 2020-07-30
Update broken link for Foundation docs

run-after:
The original link comes up with a 522 error. This is the corrected link


marvingay:
@run-after  The current link redirects to the one you added. Thank you for the contribution.

PR #19713, opened 2020-07-31
Add note regarding Weather API Key activation time

nearmint:
added note that user needs to wait for api key activation


marvingay:
@nearmint  Thank you for the contribution.

PR #19772, opened 2020-08-02
Fix statement about Macs and development

### LINKED ISSUE #19295: OS market share among developers
Even when focusing solely on developers I haven't seen data to support the claim most use macs. Going by the Stackoverflow Developer Survey 45% use windows as their platform, 29% MacOS, and 25% Linux. Objectively that's not a majority, windows has a clear lead with a plurality of 45% and MacOS is nearly tied for second place with Linux.

Outside of the development world MacOS has a single digit marketshare. I think it's objectively inaccurate to say a majority of developers use macs and misleading not to be clear that MacOS globally has a marketshare that's almost negligible.

aaron-contreras:
Fixes #19295

### DIFF
--- a/web_development_101/installations/installation_overview.md
+++ b/web_development_101/installations/installation_overview.md
@@ -33,7 +33,7 @@ Modifying or dual booting a computer to work with the tools you'll need will mak
 
 #### Mac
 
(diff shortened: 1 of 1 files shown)


marvingay:
@aaron-contreras  Thanks!

PR #19846, opened 2020-08-04
Update project_rock_paper_scissors.md

redhendery:

Thank you,

The Odin Project team.


marvingay:
@redhendery Be sure not to include a space between the `()` and `[]`, otherwise your solution will not display properly. Thank you for the submission!

PR #19882, opened 2020-08-06
Update factory-functions.md

Lucas-Bide:
I may be wrong, but since `sayName` is a variable name, I think the enveloping curly brackets are a typo. Maybe there's a concept I'm unaware of that allows that type of variable naming.


marvingay:
@Lucas-Bide It's a destructuring assignment.

PR #19899, opened 2020-08-06
Fixed small error

constantinginga:
Added js file extension to the import statement


marvingay:
@constantinginga This isn’t an error. The file extension is not needed in this case.

PR #19943, opened 2020-08-07
Fix inaccurate installation tip for guest additions

nlakritz:
When installing Guest Additions for the Xubuntu VM, I could not find any message saying to close the terminal window after running `sudo ./VBoxLinuxAdditions.run` as instructed.

Here is a screenshot of my terminal:

guest-additions-install-terminal


marvingay:
@nlakritz Thank you for pointing that out. It's appreciated!

PR #19956, opened 2020-08-08
changed snake-case to snake_case

Musabii:
it says that words are separated by underscore but instead, it uses minus sign. it is definitely trivial but I thought I'd propose to fix it.


marvingay:
@Musabii  Good catch. Thank you for your contribution.

PR #20006, opened 2020-08-10
Fix bulleted list formatting

nlakritz:
It looks like a bulleted list is intended for the list of resources at the bottom of this page, under Additional Resources: https://www.theodinproject.com/courses/web-development-101/lessons/introduction-to-the-front-end

However, the list isn't showing up formatted with bullets (see image below).

image


marvingay:
@nlakritz Thank you for your contribution.

PR #20031, opened 2020-08-10
JavaScript does have real actual classes

Jeff-Mott-OR:
ECMAScript editors have explicitly said JS classes are real. Brian Terlson, for one example, was/is spec editor from ES7-present (you'll find his name in the spec for those editions). He occasionally comments on reddit, including [this comment](https://www.reddit.com/r/javascript/comments/4bmiqj/using_classes_in_javascript_es6_best_practice/d1avktu/). "It is not at all helpful for beginners to constantly be exposed to the notion that JS classes are 'fake'," he says, "They are not."

JS classes actually behave the same as classes in Python, Ruby, or Smalltalk, among others. If you write `instance.foo` in Python, for example, then it searches "instance" for a property named "foo". If it doesn't find it, then it follows a special link to another object and searches that other object for "foo". If it doesn't find it, then it follows that object's special link to another object, and so on until it finds a "foo" property or until the chain of object links ends. Here's a look at [JavaScript and Python classes side-by-side](https://i.imgur.com/p9Kw815.png).

### DIFF
--- a/javascript/organizing-js/classes.md
+++ b/javascript/organizing-js/classes.md
@@ -1,7 +1,7 @@
 ### Introduction
-JavaScript does _not_ have classes in the same sense as other Object Oriented languages like Java or Ruby. ES6, however, _did_ introduce a syntax for object creation that uses the `class` keyword. It is basically a new syntax that does the _exact_ same thing as the object constructors and prototypes we learned about in the constructor lesson.
+ES6 introduced a syntax for object creation that uses the `class` keyword. It is basically a new syntax that does the _exact_ same thing as the object constructors and prototypes we learned about in the constructor lesson.
 
-There is a bit of controversy about using the class syntax, however. Opponents argue that `class` is basically just _syntactic sugar_ over the existing prototype-based constructors and that it's dangerous and/or misleading to obscure what's _really_ going on with these objects. Despite the controversy, classes are beginning to crop up in real code bases that you are almost certainly going to encounter such as frameworks like React.
+There is a bit of controversy about using the class syntax, however. Opponents argue that `class` is basically just _syntactic sugar_ over the existing prototype-based constructors and that it's dangerous and/or misleading to obscure what's _really_ going on with these objects. Proponents argue that `class` is implemented similar to Python or Ruby, where classes are mutable runtime objects and inheritance is delegation, and JavaScript's `class` is just as safe and clear as it is in those other languages. Despite the controversy, classes are beginning to crop up in real code bases that you are almost certainly going to encounter such as frameworks like React.
 
 Since we've already gone fairly in-depth with Constructors, you don't have too much left to learn here beyond the new syntax. If you choose to use classes in your code (that's fine!) you can use them much the same way as object constructors.
 


marvingay:
@Jeff-Mott-OR Hello. Upon conducting quality assurance on the curriculum content in question, it has been determined that the information provided is accurate and up-to-date.

See the following resources for further clarity:
[The MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes)
[JavaScript classes](https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/this%20%26%20object%20prototypes/ch4.md#javascript-classes) in You Don't Know JS by Kyle Simpson.

Jeff-Mott-OR:
I cited and quoted one of the ECMAScript spec editors. Is that considered a lesser resource than Kyle Simpson?

I3uckwheat:
> I cited and quoted one of the ECMAScript spec editors. Is that considered a lesser resource than Kyle Simpson?

In that cited comment, he states that they use prototypes under the hood for their implementation. 

See: [this in classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) at the bottom of the page as well for how not understanding JS's prototype chain can get you into trouble. 

We are not denying the similarities in other languages, however the ambiguity of "similar to Python or Ruby" is hard to separate. As those languages do not have the Prototype chain to worry about.  That is the main issue here.

marvingay:
This [table from the MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Details_of_the_Object_Model#:~:text=JavaScript%20is%20an%20object%2Dbased,of%20properties%20and%20their%20values.) may also provide further clarity on how JavaScript's use of prototypes is different from classes, by comparing Java and JavaScript.

Jeff-Mott-OR:
> In that cited comment, he states that they use prototypes under the hood for their implementation.

While also stating that JS classes are real.

[Here's another quote.](https://www.reddit.com/r/javascript/comments/75j4ju/question_js_class_design/do6rdn8/) And just a friendly reminder that this is not some random blogger. This is an ECMAScript spec editor saying, "FWIW, if class means anything, JS has them."

> As those languages do not have the Prototype chain to worry about.

They don't call it by that name (they call it the "inheritance chain"), but yes they do have that. In both Python and Ruby, inheritance is done by delegation, objects linked to other objects. So if you write `instance.foo` in either Python or Ruby, then at runtime it will search through a chain of objects looking for a property named "foo", just like JavaScript does.

> by comparing Java and JavaScript.

Java doesn't own the concept of a class. If we're going to say JavaScript doesn't have classes because they don't behave like Java's, then that means Python and Ruby also doesn't have classes, because their classes don't behave like Java's either.

Another quote from an ECMAScript spec editor:

> Some people argue that class means some common subset semantics among Java, C#, and C++ (forgetting that those languages did not invent classes). But, that common subset is very small (those languages behave quite differently) and applies to JS classes as well.

I3uckwheat:
I think the real argument here isn't about the "class" concept itself, but the differences between prototypal and classical inheritance. 

https://stackoverflow.com/questions/19633762/classical-inheritance-vs-prototypal-inheritance-in-javascript

Jeff-Mott-OR:
Which specific detail are you looking at? That SO link claims:

> we can't add new methods at run time in classical inheritance

Except... Ruby lets you add new methods at runtime. As does Python. Are we saying those languages don't have classes either?

Also, I'm not downplaying that JS classes are built on prototypes, I'm only removing the line, "JavaScript does _not_ have classes."

"if class means anything, JS has them." -- ECMAScript spec editor

I3uckwheat:
It says they aren't the same implementation as other languages.

Jeff-Mott-OR:
Which "other" languages? Because every language implements classes and inheritance slightly differently than every other. JavaScript's implementation is strikingly similar to Python and Ruby's implementation. JavaScript's implementation is similar even to Smalltalk!

Here's **another** ECMAScript spec editor, Allen Wirfs-Brock, giving a [video talk](https://www.youtube.com/watch?v=EPHmm8JiGbg&t=37m28s) comparing JavaScript classes to Smalltalk classes. “The punchline,” he says in the talk, “is they actually aren’t as different as you might think.”

I3uckwheat:
I've closed this for now, @Jeff-Mott-OR If you'd like to see this change, please make a new PR and we'll re-review it. However, it's important for people to know that JS classes, do in fact, work differently under the hood.

PR #20036, opened 2020-08-10
JavaScript does have real actual classes v2

Jeff-Mott-OR:
This is a continuation of #20031. I was explicitly told, "If you'd like to see this change, please make a new PR and we'll re-review it."

This PR removes, "JavaScript does not have classes."

The reasoning is:

- ECMAScript spec editors have explicitly said JavaScript *does* have classes.
- Under-the-hood implementation details matter, but implementing a feature differently than Java does not make it fake. If it did, then JavaScript's objects, arrays, functions, and much more would all also be fake. Those are all implemented differently than in Java too.
- JavaScript is not the only language to implement classes and inheritance differently than in Java. Python, Ruby, and even Smalltalk classes are implemented differently than in Java. In fact, classes in JavaScript, Python, and Ruby are strikingly similar. They are all runtime mutable objects that inherit at runtime via a delegation chain (aka objects linked to other objects).
- Repeated for emphasis: - ECMAScript spec editors have explicitly said JavaScript *does* have classes.


I3uckwheat:
https://github.com/getify/You-Dont-Know-JS/issues/836

After doing research, I cannot in good faith disagree with YDKJS on the opinion on classes. After talking to the team, we have decided to leave this section as-is

PR #20064, opened 2020-08-11
Change location of end quotation mark

kuosandys:
I believe the quotation marks are meant to go around the words "transpiling code". I hope this helps clarify the meaning of the sentence!


marvingay:
@kuosandys Good catch!

PR #20086, opened 2020-08-12
Fix form example typos

lopezaxel:
Fix indentation, change lost 'f' to 'form' at line 173 and because in line 173 there is a form.submit then in the example should be a input submit. Also change 'Javascript' to 'JavaScript'.

### DIFF
--- a/rails_programming/forms_and_authentication/form_basics.md
+++ b/rails_programming/forms_and_authentication/form_basics.md
@@ -167,18 +167,19 @@ From the Rails Guide:
 ~~~
 
 ~~~erb
-  #app/views/articles/new.html.erb
+  # app/views/articles/new.html.erb
   <%= form_with model: @article do |form| %>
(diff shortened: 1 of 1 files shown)


marvingay:
@Drew-Goddyn  Specifically, with the `f.submit` on line 173. `f.submit` seems valid to me, is there a difference here where it needs to be changed?

Good to merge, if so.

lopezaxel:
In line 171 is used the parameter |form| and in line 172 is used form.text_field, so in line 173 there must be used the |form|  parameter instead of the |f| parameter.

```
  171  <%= form_with model: @article do |form| %>
  172    <%= form.text_field :title %>
  173    <%= f.submit "Create" %>
  174  <% end %>
```

leosoaivan:
LGTM

marvingay:
@lopezaxel  Thank you for your contribution, as always!

PR #20087, opened 2020-08-12
Fix minor text error in DOM-manipulation.md

karen-sarkisyan:
Fixing the "on_event_" property to "onclick" property.

I'm a beginner, but as far as I know, there's no such property as "on_event_" for a DOM object, so I believe it should be "onclick", which I fixed.

That's my first ever PR, so please don't be too harsh if I'm wrong. Any feedback is highly appreciated.

Karen.


marvingay:
@karen-sarkisyan event is italicized to indicate that the event would go in that spot. “onclick” is not the only event available. event is a placeholder in that context.

PR #20286, opened 2020-08-18
change wording

meiyaismywaifu:
current: "What should be a class? Instance variable? Method?"
change: "What should be a class? Instance? Variable? Method?"
Seems to be original intent. Parallels four items, rather than a set with a strange middle.


105ron:
I believe the original wording was correct. Instance variable makes sense to me... whereas instance does not. Have another read of the material and see if you still think it should be changed?

meiyaismywaifu:
It makes sense, yes. It just flows better to me this way. If I were to extend it, it would be something like, 'What should be a class? A method? A variable? Should the variable be a class variable, an instance variable, or a local one?'. Does this illustrate the principle I'm following to you?

marvingay:
"Instance Variable" is the original intent and correct term for the context.

Closing.

PR #20305, opened 2020-08-18
Add my solution to the Testing Practice Project

pklepa:
Hi! This is an addition of a solution to a project but contains multiple commits because i had made a mistake earlier (my Linter eddited the whole document). This meant i had to rollback the changes to correctly match the Upstream before inserting my solution back again.

Thanks and sorry for the trouble!


pklepa:
Also, I had some difficulty learning how to revert changes (whether to use `git reset`, `git reverse` or `git checkout`. In the end, `git checkout` was the solution i used.

marvingay:
@pklepa  Thanks! You don't have to worry about multiple commits, they get "squashed" into one commit.

You could reset, checkout a particular branch and push it to a fork, etc. in this instance there's no right way, thanks again!

PR #20317, opened 2020-08-18
Update project_data_structures_algorithms.md

Jacobus-afk:
I found that the pretty print function's tree 'branches'  didn't nicely line up, until adding a few spaces.. has anyone else experienced the same?


marvingay:
@rlmoser99 Feedback requested.

PR #20488, opened 2020-08-24
Fix small typos on project_forms_advanced.md

lopezaxel:
Delete extra commas, add final dots and add uppercase letters.


ZachBaird:
Thanks for the grammar fixes, keep up the great work!

As an aside, the Oxford comma is grammatically correct but optional :)

PR #20639, opened 2020-08-28
Practicing git basics: Change "git add ." to "git add -A"

bryansoftdev:
Step 6 in the Add Another File section says that the full stop in "git add ." means to add all unstaged files, but the full stop actually means to add all unstaged files from only the current directory (and its subdirectories).

The correct command to add all unstaged files regardless of the current directory would be "git add -A". 

I propose we either change step 6 to use the "git add -A" command instead as seen in this PR, or clarify that the full stop only adds files from the current directory and its subdirectories.


marvingay:
@bryansoftdev  Why not both? "git add ." is probably used more often, however, both have different use cases and are worth mentioning here.

PR #20653, opened 2020-08-29
Remove explicit reference to 'render()' fn

marvingay:
This lesson has faced a high volume of questions concerning the `render()` function named within the instructions. The name often misleads students in numerous ways and as a result has been generalized.


bycdiaz:
I'm just dropping in to say this is a fine PR.

PR #20673, opened 2020-08-29
add missing brackets on two solutions

aderose:
Added two missing brackets for two solutions.

Added my own solution.


105ron:
Thanks for that :+1:

PR #20675, opened 2020-08-29
Change '<' to '<=' in the for-loops.

deniselupe:
Hello,

It seems that the conditional operator in the last three for-loops need to be changed to '<=' so that the for-loop output includes the number the user enters in the FizzBuzz prompt.

Sincerely, 
Denise Rodriguez


ZachBaird:
I think you're correct, good catch! @deniselupe

PR #20723, opened 2020-08-31
Change the conclusion to fit the JS track

sher-s7:
The conclusion assumes that the user knows about Rails back-ends and states that the next course is the Getting Hired section, which is true in the Ruby/Rails track, but not for the JS track


marvingay:
@sher-s7 Thank you for your contribution.

PR #20859, opened 2020-09-04
Link updated

tomigm:
Link 2) Was throwing server error, updated it with its 2017 copy from Archive.org

Thank you,

The Odin Project team.


marvingay:
Appears they've sorted out their issues.

Closing.

PR #21036, opened 2020-09-10
Add pry-byebug to ruby debugging

berubenic:
Added additional resource on how to install pry-byebug. I think it might be worth adding to the lesson.


rlmoser99 (changes requested):
Thanks for taking the time to submit this PR. Instead of adding it as an additional resource to this lesson, update the instructions to install `pry-byebug` globally instead of just `pry`. Include the link in the instructions, so students can read the documentation. 

If you would like add more to this lesson, feel free to add brief instructions on how to use some basic commands, like `next` and `step`.

berubenic:
Alright. Wanted to make sure it was worth adding to the lesson before working on it. I will work on this tomorrow.

berubenic:
#21079 Changes made are here.

PR #21050, opened 2020-09-11
Correct request type support context

kneepdee:
Previously, 'Form Basics' lesson stated that browsers don't support other
HTTP methods apart from GET and POST. That is not true [1] and probably 'HTML'
was meant in place of 'browsers'. The avoid confusing the reader, this
change corrects the mistake.

[1] https://stackoverflow.com/a/16812862


marvingay (approved):

marvingay:
@kneepdee Thank you for your contribution!

PR #21093, opened 2020-09-13
Update objects-constructors.md

lghiggino:
Hello.

Modified an example for additional clarity, and easier understanding of a new concept. 

As someone seeing "new" for the first time it got me confused for a while, maybe it will help others


marvingay:
The example that follows has the appropriate naming conventions for multiple players. The change made isn't related to the `new` keyword itself. 

Passing on this particular contribution because there isn't a difference in this context. Thank you for putting this out there!

PR #21152, opened 2020-09-14
Add disclaimer to hashes additional content

anth-dev:
The linked content suggests setting a key's value to nil in a hash to delete the key. Currently it just changes the value and doesn't delete the key. Added a disclaimer to use the #delete method instead.


rlmoser99 (approved):
Thank you for taking the time to submit this pull request! I did update your explanation, because I wanted to specify that the example was 'incorrect' as opposed to 'no longer working'. Thank you for bringing this to our attention.

PR #21254, opened 2020-09-18
Adds  juanpa89's Solution

juanpa89:
Very nice and Challenging Project, thx for make us do it.
The hardest was to make the calculator to do the same operation over and over again just pressing the equal sign.
(should be extra credits).

thx again

TGIF


105ron:
It is a challenging project, well done on completing it. :+1:

PR #21498, opened 2020-09-25
Update conditional_logic.md

bronsonavila:
Although the code at issue is clearly just an example demonstrating a `case` statement, I do not think we should permit it to contain code that is syntactically incorrect. Here, the example shows that an integer is allowed to contain commas. This is not true, and would result in the following error if attempted:

```ruby
> foo = 0
> foo += 50,000

# SyntaxError (syntax error, unexpected ',', expecting end-of-input)
# foo += 50,000
#          ^
```

### DIFF
--- a/ruby_programming/basic_ruby/conditional_logic.md
+++ b/ruby_programming/basic_ruby/conditional_logic.md
@@ -241,7 +241,7 @@ grade = 'F'
 case grade
 when 'A'
   puts "You're a genius"
-  future_bank_account_balance += 5000000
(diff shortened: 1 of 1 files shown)


rlmoser99 (approved):
Thank you for taking the time to submit this pull request. In looking closely at this fictitious example, I agree that we want to provide valid code. Therefore, I also edited the assignment operator, which was producing an undefined method error. I also decided to use an underscore to visually parse 5 million.

PR #21540, opened 2020-09-27
Changed this.countUp to () => this.countUp() in state lesson

paolotiu17:
I was having errors regarding the said function so I found out that it had to be in a arrow function and also the parenthesis. I don't know if it's a problem only for me though.

### DIFF
--- a/javascript/react-js/state-and-props.md
+++ b/javascript/react-js/state-and-props.md
@@ -172,7 +172,7 @@ class App extends Component {
   render() {
     return (
       <div>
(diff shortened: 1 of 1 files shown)


marvingay:
@paolotiu17 This is a good catch, you either need to place the method in the callback OR bind the method to the appropriate context

ex:
In the constructor ->
`this.countUp = this.countUp.bind(this) `

which will resolve the code block working.

I'll make the appropriate adjustment to the example.

marvingay:
Fixed in #21548

Issue #21568, opened 2020-09-28
Exhaustive dependency check fails on useEffect() example


clormor:
The [useEffect code sample](https://github.com/TheOdinProject/curriculum/blob/master/javascript/react-js/hooks.md#useeffect) throws an error in React (tested using 16.13.1) because of the exhaustive dependency check:

~~~
React Hook useEffect has a missing dependency: 'changeColorOnClick'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
~~~

The `changeColorOnClick` should either be memoized or placed in the `useEffect` hook. Since it's not needed elsewhere, and the use of `useCallback` could add confusion to this example, I would suggest moving it directly into the hook.

It may also be worth adding a note that explains why this is necessary - I am still learning React so I am not the best person to suggest what this new wording might be.

PR #21569, opened 2020-09-28
Fix exhaustive dependency error for useEffect()

clormor:
### Currently

See https://github.com/TheOdinProject/curriculum/issues/21568.
When you copy/paste the [useEffect code sample](https://github.com/TheOdinProject/curriculum/blob/master/javascript/react-js/hooks.md#useeffect) into your app, you get the following error:

~~~
React Hook useEffect has a missing dependency: 'changeColorOnClick'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
~~~

### This PR

This PR fixes the code issue identified in https://github.com/TheOdinProject/curriculum/issues/21568
but does not add any additional notes / warnings explaining why this step is necessary.

Avoids this potentially confusing error message by moving the helper function inside
useEffect(). An alternative would be to memoize it using a callback, but since this is
an educational example that might risk adding to any potential confusion.

Thank you,

The Odin Project team.

### DIFF
--- a/javascript/react-js/hooks.md
+++ b/javascript/react-js/hooks.md
@@ -52,20 +52,20 @@ const App = () => {
   const [color, setColor] = useState("black");
 
   useEffect(() => {
+    const changeColorOnClick = () => {
+      if (color === "black") {
+        setColor("red");
+      } else {
+        setColor("black");
+      }
+    };
+    
     document.addEventListener("click", changeColorOnClick);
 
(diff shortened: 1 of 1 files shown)


marvingay:
Thank you for the submission.

The dependency warning could definitely lead to confusion. Will follow up later with the course of action (which may involve me adding more to this PR).

PR #21608, opened 2020-09-29
Update DevTools2 Formatting

zjakwani:
The way the Assignment section was worded was very confusing, I thought the "elements you are already familiar with:" was referring to the bulleted subsections, since it had a colon directly to the list. Moving this sentence after the list would make it more clear in my opinion.


marvingay:
@zjakwani  The current formatting seems to be the most practical.

PR #21656, opened 2020-10-01
Update project_etch_a_sketch.md

GraceSchwar:
Adding a suggested limit for user input since multiple people have had issues with typoing/entering very large numbers and the computer making thousands of divs/square and taking a long time to make them all or locking up

Thank you,

The Odin Project team.

### DIFF
--- a/javascript/organizing-js/factory-functions.md
+++ b/javascript/organizing-js/factory-functions.md
@@ -160,7 +160,7 @@ counter(); // 2
 counter(); // 3
 ~~~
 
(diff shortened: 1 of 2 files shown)


GraceSchwar:
Added an explanation of what `counter()` does. Without the explanation that counter with () on the end was calling the return value of the factory function was very confusing to me, but once I figured that out it made a lot more sense. I have explained this to a few people already in the help channel. SO I think it may be a common misunderstanding

PR #21668, opened 2020-10-01
Remove line indication 'most' developers use Unix

ryanford-dev:
Let's avoid baiting the Unix vs Windows discussions.


marvingay:
@ryanford-dev Agreed.

PR #21682, opened 2020-10-01
Update state-and-props.md

jaredaritter:
Updated Line 61 for improved clarity about the example below it. Changed 'this works with functions' to 'props work with functions'. On my initial read I thought it was referring to function components vs class components. It wasn't until I looked at the example that I realized the sentence was referring to passing functions as props.


marvingay:
@jaredaritter Context is everything when it comes to `this`... *ba dum tsss*

Good call. Thank you for your contribution.

PR #21689, opened 2020-10-01
Etch a sketch input limit 

GraceSchwar:
Adding a suggested limit for user input since multiple people have had issues with typoing/entering very large numbers and the computer making thousands of divs/square and taking a long time to make them all or locking up

### DIFF
--- a/web_development_101/javascript_basics/project_etch_a_sketch.md
+++ b/web_development_101/javascript_basics/project_etch_a_sketch.md
@@ -29,7 +29,7 @@ If you get totally stuck drop by the chatroom.. someone will be there to point y
    2. There are multiple ways to change the color of the divs, including:
       1. adding a new class to the div
       2. changing the div's background color using JavaScript.
-4. Add a button to the top of the screen which will clear the current grid and send the user a popup asking for how many squares per side to make the new grid.  Once entered the new grid should be generated _in the same total space as before_ \(e.g. 960px wide\) and now you've got a new sketch pad. Tip: Set the limit for the user input to a maximum of 100.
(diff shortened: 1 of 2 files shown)


web_development_101/javascript_basics/project_etch_a_sketch.md
```diff
@@ -29,7 +29,7 @@ If you get totally stuck drop by the chatroom.. someone will be there to point y
    2. There are multiple ways to change the color of the divs, including:
       1. adding a new class to the div
       2. changing the div's background color using JavaScript.
-4. Add a button to the top of the screen which will clear the current grid and send the user a popup asking for how many squares per side to make the new grid.  Once entered the new grid should be generated _in the same total space as before_ \(e.g. 960px wide\) and now you've got a new sketch pad.
+4. Add a button to the top of the screen which will clear the current grid and send the user a popup asking for how many squares per side to make the new grid.  Once entered the new grid should be generated _in the same total space as before_ \(e.g. 960px wide\) and now you've got a new sketch pad. Tip: Set the limit for the user input to a maximum of 100. If a very large number is entered it will take a long time for the computer to make that many cells, this will prevent any delays.
```
marvingay, line 32: The markdown for Tip should either be bold or italicized for emphasis.

Also, I believe this would be better phrased:
"Set the limit for the user input to a maximum of 100. A larger number of squares results in more computer resources being used, resulting in possible delays, freezing, or crashing that we want to prevent."

Something like that, the main takeaway is that we should be explicit in saying that it consumes more resources!

marvingay (changes requested):

GraceSchwar:
Changes made. Used your  phrasing. Thank you

marvingay (approved):

PR #21719, opened 2020-10-02
Update state-and-props.md

jaredaritter:
Line 187. Changed 'binded' to 'bound' for correct past tense and added the bind method  at the end of the sentence in parenthesis for clarity.

As an aside, please let me know if these submissions are too pedantic. I know this is a new section and I'm glad to help out, but I don't want to create unnecessary work for the maintainers if this is too trivial. Thanks.


marvingay:
Looks good to me.

Such is the process when new content is rolled out. User feedback is invaluable and things get caught that weren't initially. Thanks as always.

PR #21759, opened 2020-10-07
Fix typo in filename

daxas-boop:
Changed "chart" to "cart"


marvingay:
@daxas-boop This change would have some unintended side effects on its own. Thank you for pointing that out. We will make a note of this and address it appropriately. For now, the site, the content, and everything else associated works as is. The Lesson is also appropriately named on the view.

Closing.

PR #21762, opened 2020-10-07
Delete duplicate content

pwiggers:
There are two sections containing the same information about Opening files in VSCode from the Command Line. Remove one such section

This is a PR template. If you are adding a solution link to the curriculum, leave this as is. If not, delete it and write the message you wish.

Thank you,

The Odin Project team.

PR #21763, opened 2020-10-07
Remove redundant copy

marcbrandon:
This commit removes copy that was repeated in the Command Line Basics lesson which describes how to open VS Code from the command line.

When going through the lesson on Command Line Basics, I discovered a block of text that was repeated verbatim. The only distinction between the removed copy and the proceeding block was that one header mentioned opening files with VS Code. Everything else was duplicated which was confusing. This will be the first PR I've ever made, so hopefully I've done it correctly.


marvingay:
@marcbrandon  You did it correctly. This was corrected in #21762. Thank you for your time.

It will be reflected on the site by tomorrow.

PR #21764, opened 2020-10-07
Fix subject-verb agreement

marcbrandon:
Use the correct conjugation of the verb 'has' to agree with the subject 'document'. Also correct the end of the sentence to match the initial subject of 'document'. The original sentence seems to have accidentally morphed the subject from 'document' to 'text editors' resulting in the error.


marvingay (approved):
Good catch! Thank you for your contribution.

PR #21778, opened 2020-10-09
Update project_practicing_git_basics.md

nearmint:
A user wasn't sure how to find the files he created in this lesson. I've added some directions.


I3uckwheat:
I'm not sure we want to be this specific in these directions. If somebody makes their folder in a different place, this wouldn't work for them. Not closing yet, waiting for more thoughts.

marvingay:
In step 5, under the "Create the Repository" section, we explicitly state:

 **"In the command line on your local machine, navigate to where you want to store this project ..."** 

Having a narrow, specific, or opinionated piece of instruction, in this case, contradicts are previous instructions. I am of the opinion that we shouldn't try to enforce where people store their files and that having this instruction would be misleading if they had already chosen to store them elsewhere.

I3uckwheat:
I think @marvingay's points are enough to close this. @nearmint Please comment if you think otherwise.

PR #21779, opened 2020-10-09
Change push to GitHub command

pwiggers:
Change "git push origin main" to "git push origin master"

The original command fails since there is no main branch on the local machine. The command should be "git push origin master" to match the default local branch.


marvingay:
@pwiggers  Moving forward, we have chosen to modify our instructions for installing Git such that the default local branch will be `main`, this will complement the change that GitHub has made in this regard.

We will provide support to those who reach out via our Discord community, GitHub, or the forum for the people that are affected by this nuance  before the changes were made within the past week.

PR #21773, opened 2020-10-08
Update Git Installation to set default branch

marvingay:
GitHub has recently changed the default branch on GitHub-initialized repositories from `master` to `main`, however `git` initialized repositories locally commit to the master branch by default. Students have faced an unintended error due to this discrepancy when following later lessons.

This commit:
- Redefines the WSL/Linux installation instructions to use an up-to-date repository for Git.
- Includes setting the default Git branch to `main` as a part of the installation process.

PR #21782, opened 2020-10-09
Fix error in "Push Your Work to GitHub".1

milofultz:
There was a discrepancy between the description and the picture that follows it:

1. Type git push origin main.

I believe it should read:

1. Type git push origin master.

This change rectifies that and makes them reflect each other.


marvingay:
We're working on updating the screenshots. The instructions, however, are intentional.

See #21779 and #21773 for why.

PR #21786, opened 2020-10-09
Fix branch name typo

juanromo1:


xandora:
No, this is correct. Main is the new default branch name.

https://github.com/github/renaming

PR #21787, opened 2020-10-10
Update project_practicing_git_basics.md

calebmirwin:
"git push origin main" doesn't work and isn't the command shown in the picture. Changed to "git push origin master"

Thank you,

The Odin Project team.


marvingay:
The instructions are intentional. We will be adding a blurb to clarify a possible discrepancy with this lesson that will be reflected tomorrow.

See #21779 and #21773

xandora:
main *is* the default branch for all new Github intialised repositories since they made the change detailed here: https://github.com/github/renaming

If you `git init` and then push to an empty Github repository then you get a `master` branch instead of `main`.

calebmirwin:
Gotcha.
I wonder why it didn't work for me then. XD
Thanks!

PR #21789, opened 2020-10-11
Add a relevant link in the lesson

The-sad-philosopher:
As I was going through the installation lesson, setting up _SSH keys_ was mentioned in brief. Naturally curious, I had to look it up because I didn't have much idea on _SSH_. 
It would be convenient to have a link to an explanation on _SSH keys_, and how to set it up thus improving the overall reading experience for students or rather mention that it would be discussed later in the future lessons if that was _the_ case.

### DIFF
--- a/web_development_101/installations/installation_overview.md
+++ b/web_development_101/installations/installation_overview.md
@@ -14,7 +14,7 @@ In the next few lessons, we will walk through these installation steps together:
 
 * installing the [operating system](https://en.wikipedia.org/wiki/Operating_system) (OS) of your choice;
 * installing a code editor;
-* creating an SSH key (a personal "password" that will identify you to GitHub, Heroku, and many other sites you'll be using);
+* creating an [SSH key](https://www.digitalocean.com/community/tutorials/how-to-set-up-ssh-keys-2) (a personal "password" that will identify you to GitHub, Heroku, and many other sites you'll be using);
 
 At the end of this section, you'll be up and running with many of the tools you need to write and run code! It may seem like a lot of steps, but we'll get through it as painlessly as possible together! If anything goes wrong, remember to use these steps:
 


web_development_101/installations/installation_overview.md
```diff
@@ -14,7 +14,7 @@ In the next few lessons, we will walk through these installation steps together:
 
 * installing the [operating system](https://en.wikipedia.org/wiki/Operating_system) (OS) of your choice;
 * installing a code editor;
-* creating an SSH key (a personal "password" that will identify you to GitHub, Heroku, and many other sites you'll be using);
+* creating an [SSH key](https://www.digitalocean.com/community/tutorials/how-to-set-up-ssh-keys-2) (a personal "password" that will identify you to GitHub, Heroku, and many other sites you'll be using);
```
The-sad-philosopher, line 17: Added a link to Digital Ocean's site on the explanation of SSH protocol and how to set up the keys.

bycdiaz:
@The-sad-philosopher We have a future lesson that walks learners through the process of setting up their SSH key. The link you shared has a similar walk through.

It would be too early for that walk through since learners who need to install Linux would not yet have an environment to set up an SSH key on. I could see some value in adding a link that explains what SSH keys are. But that link you shared would not be a good spot for that. Can you dig up a resource that talks about SSH keys generally?

I3uckwheat (changes requested):
Agreed with @bycdiaz, I also think it's pretty clear that this will be explained later. we do not want people trying to install an SSH key at this point.

The-sad-philosopher:
It makes sense now, that you've mentioned that it is explained in the future lessons. In that case, we could switch the link to a wikipedia page explaining what SSH protocol is, and such. Or maybe we could also add a phrase like '_More on that later_' or something like that to let the readers know that it will be discussed later.

The-sad-philosopher:
I was going through many good articles that explain what SSH keys are, but I find that most of them also mention how to set them up.

I3uckwheat:
@The-sad-philosopher 

If this single line needs a "more on that later" doesn't every bullet point need that too?

If the lesson isn't clear that this is an overview of what is to come, we should probably change the language of the lesson to describe that instead of worrying about this single line.


So it comes down to this point:

* Is the lesson unclear this is an overview?

If the answer is "yes" we shouldn't be focusing on ssh specifically.

The-sad-philosopher:
Well, the other points that are mentioned in the bullet points (operating system and code editor) are briefly mentioned in the same lesson _except_ for SSH keys. So, students could be left wondering when it will be discussed again. Or maybe its just me. But then again, Installation overview is just chapter 1 of web development series and the next mention of SSH keys happen in chapter 5 "Setting up Git". So to answer your question, yes I do think that this lesson leaves students hanging and a bit confused.

I3uckwheat:
@The-sad-philosopher Would you be interested in fixing that? It would be great :)

The-sad-philosopher:
Yes sure I'd be interested in that; but I had a doubt just now. I was wondering why Online editors was mentioned in this lesson. Is it related to the installation overview?

The-sad-philosopher:
I feel like I should have rather opened an issue instead of sending this PR, now that I look back.

PR #21790, opened 2020-10-11
Added clan name for codwars

GraceSchwar:
title ^


bycdiaz:
Could we toss in info on how to join the clan? Looks like they'd just need to go to their account settings on the codewars site and add the clan name to the `Clan (Company, School or Organization)` field.

PR #21800, opened 2020-10-15
Replace 'master' with 'main' in git push command

Peuchen:
Update the wording since GitHub now initializes new repositories with a 'main' branch instead of a 'master branch': https://github.com/github/renaming


marvingay (approved):
Thanks for the contribution!

PR #21808, opened 2020-10-19
Update contributing.md for doc improvements 

devenpklnddfsdf:

Thank you,

The Odin Project team.


contributing.md
```diff
@@ -1,5 +1,5 @@
 
-## How to add and make changes to the curriculum
+## How to add and make changes to the curriculum ??
```
I3uckwheat, line 2: I do not think we need a question mark here. Especially not two. This is a title of a section, not a question to be asked

PR #21814, opened 2020-10-20
Update inputs-and-lists.md

DoviMaj:
added a live demo of the working app with bootstrap from codesandbox.io


I3uckwheat:
Thank you for the PR, but I do not believe this fits here.

PR #21832, opened 2020-10-23
note not to include $ when following test key instructions in Git Setup

GraceSchwar:
I see lots of people in discord having issues because they include the `$` (the test key article shows the dollar sign). so they type it in and then it doesnt work. (section 2.5 article)

### DIFF
--- a/web_development_101/installations/setting_up_git.md
+++ b/web_development_101/installations/setting_up_git.md
@@ -147,6 +147,7 @@ Now, go back to GitHub in your browser window and paste the key you copied into 
 #### Step 2.5 Testing your key
 
 Follow the directions in [this article from GitHub](https://help.github.com/en/articles/testing-your-ssh-connection) to verify your SSH connection. If the output doesn't correctly match up, then try going through these steps again or come to [the Discord chat](https://discord.gg/hvqVr6d) to ask for help. 
+Note: Do not inlcude the `$` when typing in the commands given in the article. 
 
 ### Step 3: Let us know how it went!
 


web_development_101/installations/setting_up_git.md
```diff
@@ -147,6 +147,7 @@ Now, go back to GitHub in your browser window and paste the key you copied into
 #### Step 2.5 Testing your key
 
 Follow the directions in [this article from GitHub](https://help.github.com/en/articles/testing-your-ssh-connection) to verify your SSH connection. If the output doesn't correctly match up, then try going through these steps again or come to [the Discord chat](https://discord.gg/hvqVr6d) to ask for help. 
+Note: Do not inlcude the `$` when typing in the commands given in the article.
```
marvingay, line 150: Is this the first lesson where commands are presented with a `$`?

If so, it would be good to point out what the `$` is or the purpose it serves in conjunction with that it should not be included in the commands.

GraceSchwar:
Will get to this tonight or tomorrow!

GraceSchwar:
@marvingay  Looked back through the curriculum, the section right before this one "Command Line Basics" talks about what `$` is and explains it. I guess people just forget/dont understand it doesn't need to be re-typed when they're doing the Git Installation section right after. I don't think its necessary to repeat what the `$` is again there, the Note not to include it should be sufficient I think. 

From Command Line Basics. 

"Before we do anything, take a look at the following text: $ whoami This is a terminal command because it begins with a $. The $ is saying “Hey! Enter what follows in your terminal.” This means that we must exclude the $ when entering any command. In the example above, we would only enter whoami in our terminal. This is a common indicator so make sure that you aren’t entering $ before a command. Now that you are aware of what $ does, take your terminal for a test run! Make sure your terminal is open, type the command mentioned above, and press enter on your keyboard"

marvingay (approved):

PR #21865, opened 2020-10-30
Update project_event_manager.md

annafirtree:
- [X ] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).

(Yes, technically, I read it, but I'm frankly not sure if I'm doing this right. Was it important to clone and set the upstream result instead of making the changes directly in the forked repository on GitHub, or is that only for bigger issues? Was it important to create a new branch, and if so, does "improving the flow of information" count as a bug fix?)

#### 1.Describe the the changes made:

* Added sentences here and there to make context clearer, to improve the flow of information
* Added commas (after introductory phrases) & other minor grammar changes for slightly improved readability (or, in the case of the "5"s, just for consistency)
* Updated File.exists? to File.exist? because when I tried it, rubocop says the former is deprecated in favor of the latter
* Changed "Iteration" to "Assignment" for the final steps to make it clear to fellow students that these are challenges to try, not the next demonstrated steps

#XXXXX

N/A. I couldn't find an existing issue related to this page, but the page itself did specifically ask for improvements.

### DIFF
--- a/ruby_programming/files_and_serialization/project_event_manager.md
+++ b/ruby_programming/files_and_serialization/project_event_manager.md
@@ -23,15 +23,15 @@ After completing this tutorial, you will be able to:
 
 Imagine that a friend of yours runs a non-profit org around political activism.
 A number of people have registered for an upcoming event. She has asked for your help in
-engaging these future attendees. She wants to help attendees contact their governmental 
-representatives by finding those representatives based on the attendees' zip codes.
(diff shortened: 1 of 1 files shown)


rlmoser99 (commented):
Thank you for taking the time to clarify this lesson. I edited several of your changes, but want to review this pull request again tomorrow before I approve it.

PR #21879, opened 2020-11-07
Fix contents: Differentiate between components and modules in React Introduction

themetar:

Currently, https://www.theodinproject.com/courses/javascript/lessons/react-introduction#components states: "In React, each component is an ES6 module."

This is misleading and it will confuse people. (Just like it made me do a double take.) While there is a convention to place components each in own module, *the React functionality is independent of the module mechanic.* "IS A" misrepresents both components and modules.

### DIFF
--- a/javascript/react-js/react-introduction.md
+++ b/javascript/react-js/react-introduction.md
@@ -33,8 +33,8 @@ For example, this simple website could be divided into the following sections (c
 - `MainArticle`, which will be the component that renders your content.
 - `NewsletterForm`, which is a simple form that let's a user input his email to receive the weekly newsletter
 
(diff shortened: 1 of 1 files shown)


marvingay:
@themetar  It's a peculiar circumstance where the convention in this case is perhaps the only practice should be followed. Having multiple components in a single file is largely an "anti-pattern" for lack of a better word..

Though, this could be explained better, I do agree.

marvingay (approved):
LGTM, thanks again!

themetar:
Thanks, @marvingay 

I was more worried about the module side of things. If someone asks what are JS modules, would you say "React components are modules!" as a way to explain them? No you wouldn't.

Anyway, thanks again.

PR #21898, opened 2020-11-12
Removed a reference to front end course

ChargrilledChook:
Per title, there was a reference to the now defunct front end course in this [lesson](https://www.theodinproject.com/courses/javascript/lessons/using-ruby-on-rails-or-baas-for-your-back-end).


GraceSchwar (changes requested):
I’d suggest changing it to say “or on an alternate path” instead of specifically the full stack js course. That way it would still apply to those doing the front end only path (it still exists even though it is not advertised).

I3uckwheat:
@CatQueenCodes Nice name change!

I think I agree with that. 
> “or on an alternate path"

ChargrilledChook:
Updated per feedback - "If you skipped Ruby and/or Ruby on Rails or are on a different path, you're not quite ready to build an entire web app from scratch. The good news: you don't have to! We will provide you with resources that guide you through setting up your backend on [Firebase](https://firebase.google.com)."

PR #21922, opened 2020-11-18
Delete "Alternate Paths to a Web Dev Job" lesson

marvingay:
Because:
The lesson does not align with the fundamental goal of The Odin Project, as by the point a learner has reached the "Getting Hired" course, they ought to already be employable. The user has also been exposed to how to contribute back to The Odin Project.

This commit:
- Deletes the lesson file.


I3uckwheat (approved):

I3uckwheat:
Tyvm

PR #21935, opened 2020-11-22
Fix front end section link

CatQueenCodes:

It just linked to the whole foundations course
the link was: /courses/foundations#section-the-front-end 
but needed to be /courses/foundations#the-front-end to go directly to that section.


I3uckwheat (changes requested):
image

There's additional text here that shouldn't be. See the red goofy circle.

CatQueenCodes:
Oops! Fixed it!

PR #21950, opened 2020-11-28
Exclude RaspberryPi as option for environment

zachmmeyer:
Explicitly mention that only a laptop, desktop or supported chromebook is currently supported by the curriculum, and anything else (including raspberry pi) is not.

### DIFF
--- a/foundations/installations/prerequisites.md
+++ b/foundations/installations/prerequisites.md
@@ -5,7 +5,7 @@ If you are already using **MacOS** or **Ubuntu**, you can skip this section. Oth
 
 **IMPORTANT**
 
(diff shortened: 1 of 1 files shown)


CatQueenCodes (changes requested):
“This curriculum only supports using a Laptop, Desktop or supported Chromebook. We cannot help you set up a developer environment on a RaspberryPi or any other device” 

I think it could be shortened to this, and laptop/desktop should be lowercase.

zachmmeyer (commented):
Sounds good to me.

CatQueenCodes:
Thanks! Hopefully this deters people asking questions about it

PR #21954, opened 2020-11-29
Fixed inconsistency in Ruby lesson on loops

nanyaDev:

In the Ruby course's lesson on loops, there are 6 code examples of various loops. 5 of them use a _do/end_ block while the for loop example does not. I did some testing and it seems like the _do_ keyword is optional in all loops. Some tutorial websites use it while others do not. I have made the for loop example use the _do_ keyword for consistency, since I was often confused about when a _do/end_ block is required as opposed to when only the _end_ keyword suffices (eg. in if/else/end). The original inconsistency in the article made me think for loops were some kind of exception, which it seems they are not.

#XXXXX


I3uckwheat:
But you learned something here that you may not have otherwise. It's likely this was intentional to reward the curious and observant. I appreciate the PR, but because the syntax is accurate and correct. I think we should leave it as-is to give others a chance to learn a bit extra. TOP is sprinkled with a lot of that kind of thing on purpose.

PR #21964, opened 2020-12-02
Removing duplicate example of #split method.

philipdonaldhitchcockstilwell:
"hello".split("") is referenced first in it's own section about splits on line 215, and then again in the "other awesome ways to use string methods" section on line 233. Not necessarily hurting anything, but seems redundant to have it twice when all the other methods in the "other awesome ways you can use strings" section are new.


rlmoser99 (approved):
@philipdonaldhitchcockstilwell Thanks for taking the time to submit this PR. Instead of deleting this duplicate method, I replaced it with the delete method.

Issue #21958, opened 2020-11-30
Low quality resource


nanyaDev:
In the [Files and Serialization Lesson](https://www.theodinproject.com/courses/ruby-programming/lessons/files-and-serialization-ruby-programming), Assignment 3 links to an [article](https://skorks.com/2010/04/serializing-and-deserializing-objects-with-ruby/) that is very poorly formatted. The bulk of the article's lesson is contained in a 200+ line block of code that does not use word wrap. Since the author uses long comments to explain each part of the code, you have to constantly horizontal scroll to read the lesson, and it's a hassle to scroll back to reference the code. 

This article is over a decade old, and I'm sure a better article is out there, but since I don't currently have the expertise to differentiate between good and bad articles, I've opened up this issue.

I3uckwheat:
This article is still good and relevant. Serialization is one of those super fundamental things.

rlmoser99:
I agree with @I3uckwheat that the article is well-written (besides the code snippets). The best way to read through these code snippets is to copy and paste it into your code editor and use word wrap. 

However, I think it is worth seeing if someone can find a more updated resource before we decide.

PR #21966, opened 2020-12-02
Switch assignment order and add clarification

rlmoser99:

Assignment: https://www.theodinproject.com/courses/ruby-programming/lessons/files-and-serialization-ruby-programming#assignment

After reviewing these assignments closely, the order of these two assignments should be switched because the video in the second assignment ties into the "Right Serialization Format" link really well.

In addition, I down-graded Alan Skorkin’s post to 'skim', instead of 'read' because I think the important information is easily found by skimming. Honestly because of the poor code formatting, I wonder if most people had been skimming it anyway.

#21958


CouchofTomato (approved):
Looks good to me Mosby

Merge whenever you're ready @rlmoser99

PR #21976, opened 2020-12-04
Update fundamentals-2.md

NerdsAreIn:
Proposed punctuation change to Conditionals bullet point 5.

I punctuated the fifth bullet point under Conditionals for grammar and readability. It previously consisted of two sentences without a full stop between them.

#XXXXX

### DIFF
--- a/foundations/javascript_basics/fundamentals-2.md
+++ b/foundations/javascript_basics/fundamentals-2.md
@@ -39,7 +39,7 @@ Now it's time for the fun stuff...  So far we haven't done much with our program
 2. [This tutorial](https://www.w3schools.com/js/js_if_else.asp) is a great first glance at conditionals in JavaScript.
 3. [This tutorial](http://javascript.info/logical-operators) will teach you about logical operators.
 4. [This article](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/conditionals) reinforces the concept and provides several interesting examples of how you could use it building websites.
-5. [This article](http://javascript.info/ifelse) covers the same basic concept \(read through it as a review!\), and, more importantly, offers the usual 'tasks' at the bottom of the page!
+5. [This article](http://javascript.info/ifelse) covers the same basic concept \(read through it as a review!\) and - more importantly - offers the usual 'tasks' at the bottom of the page!
 6. [This tutorial](https://www.digitalocean.com/community/tutorials/how-to-use-the-switch-statement-in-javascript) teaches you about the `switch` statement, which comes in handy when you have multiple conditions.
 
 ### Practice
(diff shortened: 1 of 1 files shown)


NerdsAreIn:
I'm new, so I didn't realise that "#" had a reserved use when I used it in my commit message. Sorry! Finding this a bit confusing...

CatQueenCodes:
The original version seems a bit better. Thank you!

NerdsAreIn:
But surely there should at least be a full stop/period before the "And"? That's the start of a new sentence.

CatQueenCodes:
I would say the only change that needs to be made is making the "And" lower case. I don't think the extra commas are necessary. The ! inside the () is not ending the sentence

ZachBaird:
@NerdsAreIn 
Sometimes a sentence sort of "works out" that way in spoken English, but in written English that wouldn't be correct.

With written English, a sentence doesn't start with an 'and' because a conjunction is actually meant to "splice" two _idea fragments_ together. The combined fragments then represent one cohesive idea which is what the sentence presents. 

Conjunctions aren't the only things used to combine sentence fragments either. Sometimes folks combine a conjunction with a comma (,) or use a semi-colon(;). When a semi-colon is used, it should **not** be followed with a conjunction. 

Either way, I think the original definitely needs revisited but that you had added a little too much punctuation around `and`. I think something like the following would be best since the 'more importantly' is a sort of conversational interjection in the text:

5. [This article](http://javascript.info/ifelse) covers the same basic concept \(read through it as a review!\) and - more importantly - offers the usual 'tasks' at the bottom of the page!

I3uckwheat:
Lets go with what @ZachBaird said.

CatQueenCodes:
@NerdsAreIn Please let us know if you plan to make this change!

NerdsAreIn:
Hi. Sorry! I have been sick! 

I do find GitHub and its lingo rather confusing. I'll admit that I don't know how to actually make a change. Do I need "write access"? Can someone help me?

Concerning the grammar, I think @ZachBaird's suggestion of using dashes instead of commas to enclose "more importantly" is fine; this just creates a stronger break in the flow of the sentence. The point is that "more importantly" is a _supplementary_ adverbial, an adjunct, which could be removed from the sentence without affecting its meaning at all. The punctuation both reflects the natural pauses in speech that occur on either side of "more importantly" when reading the sentence aloud and its secondary grammatical role.

marvingay:
@NerdsAreIn 
re: "Do I need write access?" 
No., it shouldn't be necessary to make any further changes.

Someone will get back to you regarding the change itself.

NerdsAreIn:
Hi. I can see that no one has changed this yet and there is still a capital initial letter in the middle of what is currently a single sentence. Could somebody please change this, at least? Thanks in advance. Major grammar errors stress me... Haha.

CatQueenCodes:
@NerdsAreIn you can update this pull request with the changes!

CatQueenCodes:
Pushed it! Thanks for the change.

PR #21985, opened 2020-12-08
Add section on API key best practices

jessicaawhy:

I added a section introducing API Key best practices to the API section of the Javascript course.

n/a

### DIFF
--- a/javascript/async-apis/APIs.md
+++ b/javascript/async-apis/APIs.md
@@ -208,8 +208,9 @@ If you've gotten lost along the way, check out [this jsbin project](http://jsbin
 For now, we will not be using any APIs that are vulnerable to attacks like key theft. If you are using APIs from sources outside of our recommended curriculum, you should be aware of the potential security risks of publishing your API keys on github.
 
 1. Learn about why you should not expose your API keys [here](https://www.programmableweb.com/news/why-exposed-api-keys-and-sensitive-data-are-growing-cause-concern/analysis/2015/01/05). The article references [My $2375 Amazon EC2 Mistake](https://web.archive.org/web/20150102022540/http://www.devfactor.net/2014/12/30/2375-amazon-mistake/) written by Andrew Hoffman after he had accidentally pushed code to GitHub that included his AWS API keys.
-2. Read about why and how to secure your API keys [here](https://nordicapis.com/hiding-and-securing-your-api-keys/). This is just one method that details how to secure your key with `dotenv`.
-3. If you accidentally publish an API key, the first thing you should do is to immediately invalidate the key. Read [here](https://jordan-wright.com/blog/2014/12/30/why-deleting-sensitive-information-from-github-doesnt-save-you/) about why just deleting the information from github does not work.
+2. If you accidentally publish an API key, the first thing you should do is to immediately invalidate the key. Read [here](https://jordan-wright.com/blog/2014/12/30/why-deleting-sensitive-information-from-github-doesnt-save-you/) about why just deleting the information from github does not work.
+
+This does not currently apply to the projects we are working on and there will be a section later in the course that goes over security. It is good to understand that there are risks to publicly storing API keys (and other sensitive data) with your code.
 
 ### Assignment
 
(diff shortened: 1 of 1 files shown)


javascript/async-apis/APIs.md
```diff
@@ -203,6 +203,14 @@ If all goes well, you should see a new image on the page every time you refresh!
 
 If you've gotten lost along the way, check out [this jsbin project](http://jsbin.com/canofar/edit?html,output). Besides the glorious styling, this is what your version should look like.
 
+### APIs: Best Practices
+
+For now, we will not be using any APIs that are vulnerable to attacks like key theft. If you are using APIs from sources outside of our recommended curriculum, you should be aware of the potential security risks of publishing your API keys on github.
+
+1. Learn about why you should not expose your API keys [here](https://www.programmableweb.com/news/why-exposed-api-keys-and-sensitive-data-are-growing-cause-concern/analysis/2015/01/05). The article references [My $2375 Amazon EC2 Mistake](https://web.archive.org/web/20150102022540/http://www.devfactor.net/2014/12/30/2375-amazon-mistake/) written by Andrew Hoffman after he had accidentally pushed code to GitHub that included his AWS API keys.
+2. Read about why and how to secure your API keys [here](https://nordicapis.com/hiding-and-securing-your-api-keys/). This is just one method that details how to secure your key with `dotenv`.
```
I3uckwheat, line 211: This doesn't apply to frontend code. If you hit the API directly from the frontend, the key has to be there, dotenv or not. 

This could be misleading to students. 

We should talk about the *only* way to secure a key is to keep it on the backend in a file that is not committed to git. Which we will get to later in the course.

I3uckwheat (changes requested):

javascript/async-apis/APIs.md
jessicaawhy: That makes sense. I think we can remove #2. Let me know if you have any suggestions for the rest of this. Thanks!

marvingay:
I'm not entirely sold on the necessity of including this portion here.

While it is imperative to not push your API keys to GitHub, it isn't possible to hide the API Key on the frontend if that is where the call is made. To truly obfuscate that information, the requests involving secure/sensitive information would need to happen on the frontend.

As the curriculum is currently laid out, people coming from the Ruby on Rails path will have already been exposed on how to handle the security for this, if they choose to incorporate a small backend API for any of the JS projects. With the Node.js path, they are exposed to learning about API Security in the Node.js course. Given that you can't actually secure a frontend only application in the desired manner and given that these projects are initially intended to apply the material learned (where revisiting projects to make them as good as you'd like is highly encouraged), I don't see the actual benefit of including this here.

While not including keys on GitHub is a great first step, they will still be exposed to the client which may raise questions about the integrity of the information at this point if someone notices that.

I'm opposed to the inclusion of this at this time, but am open to more discussion from others.

jessicaawhy:
I took a look at the Security Configuration / API Security section section yesterday and it looks solid. I think the general thing I'm trying to convey is that security is one of the core concepts for if we want to work on the web and I think an introduction to it is necessary earlier than in the node course section. Even though we can't secure things until we learn the backend, I think it's important to be aware of risks earlier. 

I'm currently just learning and really glad I had someone warn me about using rate limited APIs in one of my projects. I thought a mention of it would be useful as I would not have known that otherwise going through the course.

jessicaawhy:
It's not about front end security specifically but rather the exposure of keys in general. Students are being implicitly shown here that it's okay to just push keys to GitHub where they can be crawled and exploited. Sure these apis are free but you're expecting students to assume the risks for something they aren't aware of. IMO it's a disservice to not make a mention of the potential risks involved with having an api key or any private data publicly available until a new developer gets to node and the backend. Security is a fundamental concept and we should be aware of the risks however minimal they are at this stage. That way we can make the conscious decision to push keys even though it is not a best practice.

I3uckwheat:
@jessicaawhy This isn't a major problem we've seen. I really don't think we need to make a big deal about it for the reasons @marvingay pointed out. While security is important, it's really hard for people at this stage to truly understand the ins and outs of it. 

I think a simple note at the bottom saying:

> While we are pushing this api key to the frontend, this isn't something you should do with any key that is not free, anything on the client is public knowledge. Handling keys that are unsafe to push to the frontend will be taught in later sections if you haven't been exposed in the Ruby course.

would be enough.

I3uckwheat (changes requested):
Lighten up to the discussed snippet.

I3uckwheat (approved):

Issue #21945, opened 2020-11-25
Remove Bootstrap mentions in the React portion of Javascript


I3uckwheat:
See: #21944 for more info on *why* we want to make this change.

Completion criteria:

- [x] Remove Bootstrap references from [this](https://github.com/TheOdinProject/curriculum/blob/3ab3f314d3d13b818db43fa4ed09d8b7042fbc17/javascript/react-js/inputs-and-lists.md) page.
- [x] re-work the content to make sense given the new context without bootstrap
- [x] identify other places in the JS section that reference bootstrap and remove them.

vinicius77:
May I help with this issue? I could also make a brief introduction about styling React components using CSS in order to keep the styling context valid even without Bootstrap.

marvingay:
@vinicus77 You're welcome to start working on this if you'd like, I'll go ahead and assign it to you.
If you have any questions, concerns, or comments do not hesitate to ask here.

vinicius77:
Thanks for assigning me @marvingay . I am already working on it. :)

twalton83:
The first bullet has had changes merged. Thank you @vinicius77!

TaylorJonesTRT:
I have found another mention of Bootstrap in the React course on this [page](https://github.com/TheOdinProject/curriculum/blob/master/javascript/react-js/project-cv.md) that I'll take care of removing!

twalton83:
Thank you @TaylorJonesTRT!

twalton83:
The last PR fulfilled the criteria of completion for removing all instances of Bootstrap from the Javascript course, so this may now be closed.

PR #22007, opened 2020-12-17
21945

vinicius77:

- Removed Bootstrap references from this page,
- Re-worked the content to make sense given the new context without Bootstrap,
- Identified other places in the JavaScript section that referenced Bootstrap and removed them.

Remove Bootstrap mentions in the React portion of Javascript #21945

### DIFF
--- a/javascript/react-js/inputs-and-lists.md
+++ b/javascript/react-js/inputs-and-lists.md
@@ -84,14 +84,14 @@ class App extends Component {
     const { task, tasks } = this.state;
 
     return (
-      <div className="col-6 mx-auto mt-5">
+      <div>
         <form>
-          <div className="form-group">
+          <div>
             <label htmlFor="taskInput">Enter task</label>
-            <input type="text" id="taskInput" className="form-control" />
+            <input type="text" id="taskInput"/>
           </div>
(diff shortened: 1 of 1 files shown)


twalton83:
@vinicius77 Do you mind updating the title of the PR to reflect the changes? Thank you!

javascript/react-js/inputs-and-lists.md
```diff
@@ -84,14 +86,14 @@ class App extends Component {
     const { task, tasks } = this.state;
 
     return (
-      <div className="col-6 mx-auto mt-5">
+      <div>
```
twalton83, line 89: Bootstrap's framework relies on a lot of divs to wrap elements. Removing the Bootstrap classes from these divs is a great start, but it may be a poor practice to keep all of these unnecessary divs in the HTML. I think we should remove these divs as well, since, for example, we really don't need a div wrapping a single button. The outermost div can stay for React's single root element needs.

vinicius77: Hi, @twalton83 

Good catch! I am gonna make all the chances you have suggested. Thanks for the advice.

PR #22012, opened 2020-12-18
js / react / inputs-and-lists.md ... misc typos and minor edits

qooqu:

js / react / inputs-and-lists.md ... misc typos and minor edits


javascript/react-js/inputs-and-lists.md
```diff
@@ -165,6 +168,7 @@ In your `App.js` component in your render method, add an onChange handler to you
 ~~~
 
 Notice that we also have to specify the `value` attribute for React input elements. In this case we want the value of the input field to be what we saved in our `task` state.
+
```
twalton83, line 171: I think there should just be a comma after "state", followed by the "and also add (...)" so that it flows, rather than starting a sentence with "and".

twalton83 (commented):
Thank you for catching these!

javascript/react-js/inputs-and-lists.md
qooqu: The "And" is not a continuation of "Notice that ...". It's a continuation of "add an onChange handler ..." from above the code block.

It's not the most elegant language, but I think it gets the job done as is.

twalton83: Agree, I see it.

PR #22015, opened 2020-12-19
Update the Rails Basics project to use the Rails Guides getting started project

CouchofTomato:
Because the existing blogger 2 project uses Rails 4 which is now
outdated

This commit changes the project to use the Rails getting started blog
project

### DIFF
--- a/rails_programming/databases_and_activerecord/active_record_basics.md
+++ b/rails_programming/databases_and_activerecord/active_record_basics.md
@@ -137,8 +137,7 @@ That was really just a teaser about what Active Record can do. In the reading be
 
 #### Basic Active Record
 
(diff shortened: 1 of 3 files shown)


rlmoser99 (commented):
Do you think we should indicate that they can skip the installation information? Another option would be to say that they can start at section 3.2 or to just link them to that location. 

Also, do you think we should remind them to set it up on github, since the tutorial does not?

Otherwise, it looks good to me.

rlmoser99 (commented):
Also, this Getting Started tutorial is currently an additional resource on this project/lesson, so we should probably delete it from that section.

CouchofTomato:
@rlmoser99 

Thanks Mosby. Changes made. I think they'd benefit from at least reviewing the install section so I just made it clear they shouldn't need to install Rails.

rlmoser99 (approved):
Looks great!! ⭐

PR #22023, opened 2020-12-23
Trying to fix formatting issues

CatQueenCodes:
On the webiste #3 is showing as #1 and the info on linking github was not in line. I tried to set it so all the bullets for 2. are indented, and then indented the github info at the bottom. This *seems* to make it render correctly in the https://www.theodinproject.com/lessons/preview. Previously the GitHub render looked correct, but wasn't once actually on the site

Screen Shot 2020-12-22 at 9 39 19 PM

I *think* this should fix it. 

...your text here

#XXXXX


rlmoser99 (approved):
LGTM

PR #22050, opened 2020-12-30
Fix spelling error and switch order of two steps

msespos:

a) Fixed one spelling error
b) Switched order of two steps in form_for section of project

#XXXXX


CatQueenCodes:
Can always count on you to catch spelling and grammar errors 😸 Thanks!

msespos:
I switched steps 4 and 5 because I think the user needs the info that was in step 5 before they can start playing around with the suggestions that were in step 4.

msespos:
Apologies for adding the second commit to this one; that was an error on my part - I meant them to be separate PRs.

CouchofTomato:
Hey @msespos 

Thanks for the PR. None of the options in step 4 require a working create method. However I can see your point about having to change the create method to accept the params that will come from the form. The only thing I don't like is moving "test it out". That feels like it should be the last step to test out the form and make sure everything is working. How would you feel about leaving step 4 where it is currently, and instead just moving the line "You'll need to switch your controller's `#create` method again to accept the nested `:user` hash from `params`." to the end of point 3? Then we can just leave point 5 as "test it out".

I'm open to other suggestions if you think something else looks better.

msespos:
@CouchofTomato So I think you are absolutely right and I was actually not sure about moving "test it out." So I think your suggested change sounds like the best one. Do I make this change and resubmit, or is it better if you make it at this point? Just let me know. Thanks!

CouchofTomato:
@msespos 

All yours mate.

CouchofTomato:
Thanks @msespos

CouchofTomato (approved):

Issue #22051, opened 2020-12-31
Add Nested Arrays Section in Basic Ruby Lesson


rlmoser99:
Lesson: https://www.theodinproject.com/courses/ruby-programming/lessons/arrays

Background: 

Students are often confused when they update one value in a 2D array and multiple values change. This happens when they create a create a 2D array similar to an example in our lesson, `Array.new(3, Array.new) #=> [[], [], []]`.

Currently, the lesson does not mention that the second optional argument populates the array with references to the same object. 

> Therefore, it is only recommended in cases when you need to instantiate arrays with natively immutable objects such as Symbols, numbers, true or false.
https://ruby-doc.org/core-2.7.0/Array.html

Acceptance Criteria:

In the "Creating Arrays" section:
- Remove this example: `Array.new(3, Array.new) #=> [[], [], []]`
- In your own words (using the documentation as a guide), explain why the second optional argument should only be used for natively immutable objects. 

Create a new section "Nested Arrays" at the end of the lesson
- In your own words (using the documentation as a guide), explain how to create a nested array that will not mutate.
- Show an example of creating a nested array without using a block, such as `Array.new(3, Array.new(3))`. Then update one value, for example `array[0][0] = 1` and reveal the updated value of the original array. 
- Show another example of creating a nested array using a block, such as `Array.new(3) {Array.new(3)}`. Then update one value, for example `array[0][0] = 1` and reveal the updated value of the original array.

wise-king-sullyman:
I would love to work on this one.

rlmoser99:
@wise-king-sullyman Thank you so much! Feel free to ping me if you have any questions.

PR #22055, opened 2021-01-02
Lesson/ruby nested array section

wise-king-sullyman:

Added "Nested Arrays" section to ruby "Arrays" lesson. 
Added a note to explain why not to use mutable objects in the second argument of Array.new(). 
Revised Array.new() examples used.

#22051

### DIFF
--- a/ruby_programming/basic_ruby/arrays.md
+++ b/ruby_programming/basic_ruby/arrays.md
@@ -137,6 +137,37 @@ Here is a brief look at some other common array methods you might run into:
 [1, 2, 3].join("-")     #=> "1-2-3"
 ~~~
 
+### Nested Arrays
+As mentioned in the note at the bottom of the "Creating Arrays" section, nested (or multi-dimensional) arrays of mutable objects (such as other arrays, hashes, or strings) created with the `Array.new(size, object)` syntax may behave unexpectedly. Examples:
+
+~~~ruby
+array_of_arrays = Array.new(3, Array.new(3))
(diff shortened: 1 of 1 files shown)


rlmoser99 (changes requested):
As discussed with you on Discord, the majority of these changes will make more sense to the students after they have been introduced to mutability and passing a block to a method (in the Basic Enumerable Methods lesson). 

In talking with other maintainers, it seems best to create a new lesson focusing on nested arrays and hashes. This material focuses more on making an array of mutable objects, then nested arrays like the original issue stated. 

Let's keep the changes to the code examples that you made in the "Creating Arrays" section. If you would like, you could provide another example using a symbol or boolean. 

The rest of the changes should be removed. I know that you really wanted to include a note without mentioning any of the new concepts because mutable strings tripped you up. However, it would be confusing to the student before these concepts are fully explained. Students need to know and understand the reasons behind our suggestions, therefore it would be best to clarify the use of this second argument in the future nested arrays & hashes lesson. 

I really appreciate the time that you have put into this new section. I wish I had the foresight to realize that this information needed to be a new lesson, instead of a new section in this lesson. As we both know, this information will be very beneficial for students to understand before they do the Ruby projects. Writing up the outline for the new lesson is a bit more time-consuming than I had expected. As soon as I finish it, I will touch base with you to see if you are interested in contributing.

wise-king-sullyman:
10 4, expect a new commit with just the adjusted code examples shortly.

rlmoser99 (approved):

Issue #22040, opened 2020-12-29
Clarify prerequisites note about supporting other operating systems


rlmoser99:
Lesson:
https://www.theodinproject.com/courses/foundations/lessons/prerequisites

Acceptance Criteria:
- Clarify the last line "If you feel you can contribute and support Windows at The Odin Project, please create a PR with Windows installation directions, and fixes for wherever the Windows commands might differ from Linux."
- Optional: Add that our Discord community will not support anything outside of these prerequisites.

zachmmeyer:
I think with the PR closed, that means this issue can be closed as well. Do you agree @rlmoser99 ?

rlmoser99:
Yes, totally. I just assumed it closed it automatically, but I should have checked.

PR #22059, opened 2021-01-03
Clarify prereq note about supporting other OSs

zachmmeyer:
In response to #22040.

Added a few extra things regarding WSL and posting in the Discord.


marvingay:
I am of the opinion that this section be nixed altogether. Let's get more feedback on here.

rlmoser99:
My biggest concern is that it sounds like we are one PR away from supporting WSL, which I think is misleading. I also think the top portion could be summarized into 2-3 sentences & even placed in the WSL section. 

I do not think this section should be nixed, because I think we need to properly educate our students on why we do not support WSL and that they should not expect to receive alternative OS support from our Discord server.

twalton83:
I agree with @rlmoser99 in that it seems like we are framing it like we "might" support WSL. I also agree that I don't think we need to nix the section, because people ask "why" Linux/Mac frequently enough. I don't think we need to necessarily emphasize WSL this much, though. I'm of the belief that bringing attention to things you don't approve of only spurs more interest in that thing.

marvingay:
#21992 Only semi-relevant, I recently nixed a section from the prereqs in the aforementioned PR.

It's somewhat of a slippery slope, because by that same logic, we would then need to justify somewhere why we don't teach Angular or Vue (anymore) in favor of React, as an example. 

When the question arises it can be handled on an as-needed basis, in my opinion.

zachmmeyer:
I think the idea is that with this problem specifically (WSL), it should not have to be brought up anymore and explained every single time, as it comes up a lot.

lucitemple:
My comments here are in regard to this section, not just the edit here in the  PR. 

The wording of this section might need to consider that, as it is at the beginning of the course, and there are many complete newbies, jargon (e.g "PR") and the complexities of different environments can be highly confusing. 

I understand there may have been questions about WSL and environments on Discord, so it may need to be addressed somewhere, however this content (both the existing and the edit) somewhat puts the cart before the horse - presenting higher level arguments from a defensive position before a newbie necessarily even knows what an "environment" or WSL is.

I recommend _first_ ensuring students understand the necessary concepts being spoken of (teaching). This should be the priority.

What would be helpful would be to have a heading "SetUp for Mac / Linux users" which basically says how this is optimal and needs no special set up because... [reasons].

Another heading "SetUp for Windows Users" which explains why things will be harder for them, what the work-around options are, and makes clear that TOC is developed to support students with a particular set-up. Provide guidance to set up the preferred environment.

While TOP can't provide curriculum to support every possible environmental configuration, it's also important to be aware that in the world at large more people use PCs and Windows, many newbies entering this space are not going to have perfect gear, or be confident making radical changes to their existing system, so it would be best to  provide information to help them navigate this space to A) get themselves set up as best they can with what they've got and B) know what is ideal, so when they are able to upgrade (and/or their tech skills and confidence improve) they can set themselves up better. 

I must admit that when I first joined Odin, I was using a Windows PC, and trying to use Virtual Box was a pretty horrible entry experience. It came after first having done LaunchSchool's prep course - where preferred environment for windows users was free level AWS linux environment - which was a much easier entry point imho. Now I have a mac, and the experience is night and day easier compare to trying to hack Windows. But I wasn't in a position to start with a Mac... so anything we can do to help people get on the path to learn coding, despite limitations, is a good start.

marvingay:
@lucitemple Having the terminology spelled out is agreeable, especially at this point.

Providing "workaround" options is the same as supporting it, which isn't on the table at this time.

I3uckwheat:
@lucitemple 

We cannot spend the resources to find "workarounds", we have decided to not support windows because of the numerous problems it causes people. While it may be better than it used to be, we're tired of explaining it in the chat ad nauseam. Wearing out maintainers is a good way to kill an open source project. 

We've tried Windows before, and it's proven to be an issue, adding workarounds is going to just add to the load of the already stretched thin dev team here. It works fine for some people, but for many it does not.

Long story short, we will not support windows. We have explained the justifications over and over and over. It's time to put it to bed so we can actually help people learn to code.

I3uckwheat (approved):

PR #22063, opened 2021-01-04
Update grid_layout.md

DoviMaj:
- [x] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).

#### 1.Describe the changes made:

Added a cheatsheet reference that is visual, easy to use and based upon the W3C CSS Grid Specification,

#### 2. If this PR is related to an open issue please reference it with the `#` operator and the issue number below:

#XXXXX

PR #22064, opened 2021-01-04
Update flexbox_layout.md

DoviMaj:

Added a cheatsheet reference that is visual, easy to use and based upon the W3C CSS Flexbox Specification.

#XXXXX

### DIFF
--- a/html_css/flexbox_layout.md
+++ b/html_css/flexbox_layout.md
@@ -31,3 +31,4 @@ This section contains helpful links to other content. It isn't required, so cons
 * Try the interactive [Flexbox froggy](http://flexboxfroggy.com/) game.
 * This [MDN Flex article](https://developer.mozilla.org/en-US/docs/Web/CSS/flex), which shows examples of what you can achieve with Flexbox.
 * [Dive Into Flexbox](https://bocoup.com/blog/dive-into-flexbox) is an in depth article with examples.
+- A [cheatsheet](https://yoksel.github.io/flex-cheatsheet/) for quick reference.


leosoaivan:
I think a cheatsheet would be a great addition, but as far as cheatsheets go, I think there could be better options.

Could we find/include something that is condensed, so that it is more visual than text-based, and easily viewable without scrolling much?

DoviMaj:
the only other one that does a real good job visualy that im aware of is this one https://flexbox.malven.co/, but imo its dosent have good readability.. the content is too small and the focus there is more on design then on UX since thats what they're selling.

Other reasons i prefer personally this one: https://yoksel.github.io/flex-cheatsheet/ are.
1) u have easy and quick access to what ur looking for through the side bar nav.
2) U can play with the diferent flex options and see the actual diference of chosing one over the other .

But im open to opinions and sugestions of course.

marvingay:
Similar to #22063, this resource is not a 'go' from me due to multiple themes failing contrast/accessibility guidelines.

Leaving open, in the event an alternative is found as discussion is underway.

DoviMaj:
I don't really get it @marvingay, if it's a good resource but some themes are not optimized.. i actually never had tried the other themes until u said they're bad. since the original one works well.

marvingay:
@DoviMaj I'm not sure which is the "default" theme, however, the one I receive on several devices is out of compliance as are most other themes. It's not about being "optimized", many who are vision impaired or colorblind won't be able to discern what's in the resource.

I'm firm in saying I tested the contrast on multiple themes and it fails, which is unacceptable.

DoviMaj:
ok. what do u think of this one https://flexbox.malven.co/ @marvingay?

 off topic: im curious how do u evaluate the contrast/accessibility, by eye or do u use a tool for that?

I3uckwheat:
I really do think that css tricks is enough for this lesson, overloading people is a good way to slow them down.

PR #22083, opened 2021-01-09
Update project_forms.md

berubenic:

Lesson currently indicates that when creating a form with only HTML, when validations fail, previously entered information will be available when rendering `:new`.

This is not the case, this functionality is only available when using `form_for` or `form_with`.

Removed mention in pure HTML and moved it in the `form_for` section.

Would it be worth adding a `form_with` section?


CouchofTomato:
Thanks @berubenic 

Do you think it might be confusing to still reference that the redirect to the new user form will show a blank form?

 "If you successfully save the user, you should redirect back to the New User form (which will be blank) and if you don't, it should render the `:new` form again"

With regards to adding a form_with section I'm unsure if we should keep a form_for as gems like devise still render that or just change all the references from form_for to form_with as that is what users should use going forwards and we don't want to confuse them.

berubenic:
@CouchofTomato
 Yes, it could cause confusion. It could only mention in the form_for section that it will keep previously entered information.
 
 Maybe adding it as an exercise to change the form that is made using form_with, referring to the docs on the correct syntax.
 Explaining that form_for is progressively getting replaced by form_with but it is still common to see.

PR #22084, opened 2021-01-09
Removes mention of online editors.

bycdiaz:
This section feels out of place since this page is dedication to the tools a learner will need to install.

In addition, we make learners aware of repl.it in the Join The Odin Community lesson.

I'm planning on expanding this page with more thoughts on Windows so we'd benefit from trimming the fat, as they say on the streets.


marvingay (approved):
Looks good to me. Having this here does seem somewhat contradictory for what we'd like, long-term.

I3uckwheat:
I do think we need to mention repl.it somewhere though, we're going to wind up having a ton of questions asked with pasted code.

bycdiaz:
@I3uckwheat We do mention it here: https://www.theodinproject.com/courses/foundations/lessons/join-the-odin-community

PR #22087, opened 2021-01-10
Add how to discover responsive width breakpoints

PaperOrb:
Added links showing how to use browser dev tools to show width breakpoints where responsive elements change.

### DIFF
--- a/html_css/project_responsive.md
+++ b/html_css/project_responsive.md
@@ -11,7 +11,7 @@ As usual, don't worry about the small stuff like sharing buttons or comments or 
 2. Follow the instructions atop the [Google Homepage project](/courses/foundations/lessons/html-css) to set up a GitHub repository for this project (of course you'll need to change the title).
 3. Create a new HTML document.
 4. Think about all the elements on the page and how they are grouped together.
-5. Try to discover which browser widths are "breakpoints" where element properties change (e.g. the sidebar disappearing).  Make a plan for which elements need to be modified at which widths.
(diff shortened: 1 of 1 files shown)


I3uckwheat:
@PaperOrb 

I think we should leave this up to the learner to research and discover. Reading documentation is a VERY important skill that most courses/tutorials do not cover. Leaving this as-is helps students by giving them a chance to discover something themselves. I won't close this however, somebody else can chime in.

PaperOrb:
> @PaperOrb
> 
> I think we should leave this up to the learner to research and discover. Reading documentation is a VERY important skill that most courses/tutorials do not cover. Leaving this as-is helps students by giving them a chance to discover something themselves. I won't close this however, somebody else can chime in.

It seems more likely people would just resize the browser window, guestimate the breakpoints, and be none the wiser to the existence of these tools. If leaving students to discover tools and then read documentation is the goal though, I don't think it's been applied evenly throughout TOP--take the instructions on how to initialize a Git repo for instance.

I3uckwheat:
@PaperOrb There are selective things that are left up to the user on purpose. Some things are out of reach to discover without being too deep in a sea of information. I think to selectively chose to let someone discover this themselves is important here. 

I do think encouraging them to research is a good move here, we could instead opt to give them some keywords, and word it as such, "Research responsive mode in your browser and identify the breakpoints" or something like that.

marvingay:
The "Developer Tools" lesson directly precedes this particular project. The learner will have been exposed to the documentation and have some familiarity, likely fresh in their mind due to recency. For them to be "none the wiser", it would mean they skipped material.

I do think it is a tad ambiguous so a rephrase or update could prove useful. But due to the above points, I don't think curbing the problem solving aspect here is beneficial to the learner, as we intend them to apply and research the concepts they are introduced to.

PaperOrb:
@I3uckwheat Changed the commit so it says something very similar to your suggestion.

@marvingay I must have skimmed over it. Was part 2 of step 2 where it was touched on? https://www.theodinproject.com/courses/html-and-css/lessons/responsive-design#assignment

PR #22089, opened 2021-01-10
Add disclaimer to onextrapixel article

dwiehoff:
HTML5 wasn't a thing back then. Today, we would be using semantic elements. Beginners may get the wrong idea when looking at the code that's part of the article?

Added a disclaimer behind the article for beginners not to misunderstand.

#XXXXX


I3uckwheat:
@dwiehoff Instead of a disclaimer, we should find an alternative resource all-together.

PR #22100, opened 2021-01-15
Include ssh success prompt

zachmmeyer:
Comes up often enough that people do not understand this is the success response.


twalton83 (approved):
Thank you for this!

PR #22106, opened 2021-01-19
Move disclaimer about omitting $ from GH page

zachmmeyer:


CatQueenCodes:
Good placement! Thanks :)

PR #22107, opened 2021-01-19
Update project_event_manager.md

OxyF1sh:

I believe there is a bug in the final code block of project_event_manager.md (line 1483), the function parameter for the legislators_by_zipcode function is zip but civic_info.representative_info_by_address sets the google API's adress to zipcode (which is undefined) instead of zip. This leads to the rescue block being executed every time. I suggest changing address: zipcode to address: zip.

#XXXXX


rlmoser99 (commented):
Great catch! It has been a while since I did this project, so I just ran through it again to confirm your findings. Thank you for taking the time to submit this PR!

Issue #21971, opened 2020-12-04
Add zsh related instructions for NVM/node installation


twalton83:
- Currently [this lesson](https://github.com/TheOdinProject/curriculum/blob/master/foundations/installations/installing_node.md) references bash related files such as ~/.bashrc

- For those who are on Catalina and above, the default terminal is now zsh and the correlated file is ~/.zshrc

Lesson should be updated to include instructions on amending a `~/.zprofile or `~/.zshrc` file instead.

SidWorks:
I am working on it!

twalton83:
Thank you.

twalton83:
Closed with #22187

PR #22115, opened 2021-01-21
added nodejs installation instruction for those using zsh

SidWorks:

Added the instruction for users of mac to install node using nvm in zsh.

#21971

@twalton83  Added zsh related instructions for NVM/node installation. Merge the PR.

### DIFF
--- a/foundations/installations/installing_node.md
+++ b/foundations/installations/installing_node.md
@@ -1,12 +1,17 @@
 ### Introduction
 
-Node is a JavaScript interpreter that doesn't run in the browser, there are many tools that require this to be installed. Testing frameworks and bundlers are two of many.
+Node.js is a JavaScript runtime environment, that means that now you can run JavaScript code on the server side also!
+Node uses JavaScript as its scripting language, and achieves high throughput via non-blocking I/O and a single-threaded event loop.
+
+There are many tools that require this to be installed. Testing frameworks and bundlers are two of many.
 
 We're going to install it using `nvm` (Node Version Manager), because it makes it easy to change node versions and upgrade node.  There is another tool called `npm` (Node Package Manager) that you will use later to install the various libraries and tools used in javascript environment.  It can be easy to confuse these two so read carefully!
 
 Node is also very easy to install using nvm, so this should go quickly :)
 
-### Step 0: Prerequisites For Linux
+# Installation in Linux
+
+### Step 0: Prerequisites 
 To install NVM properly, you'll need `curl`. Simply run the command below:
 
 ~~~bash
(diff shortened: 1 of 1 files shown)


foundations/installations/installing_node.md
```diff
@@ -1,12 +1,17 @@
 ### Introduction
 
-Node is a JavaScript interpreter that doesn't run in the browser, there are many tools that require this to be installed. Testing frameworks and bundlers are two of many.
+Node.js is a JavaScript runtime environment, that means that now you can run JavaScript code on the server side also!
```
twalton83, line 3: I’m not sure if this early on we need to explain this to learners- it’s way over a beginner’s head re: “high throughput”

foundations/installations/installing_node.md
```diff
@@ -82,3 +87,112 @@ nvm use node
 Now when you run `node -v` you should see `v14.10.0` or something similar.
 
 If you see that, you have successfully installed node!
+
+
+# Installation on MacOS (zsh)
```
twalton83, line 92: This doesn’t follow my directions- we want to *include* zsh not replace the bash instructions.

Where did you get these instructions from? We really only needed to reference the two zsh files. We didn’t want to completely overhaul our directions.

foundations/installations/installing_node.md
```diff
@@ -82,3 +87,112 @@ nvm use node
...
+
+~~~zsh
+
+source ~/.antigen/antigen.zsh
+# This loads the oh-my-zsh's library.
+antigen use oh-my-zsh
+# Bundles from the default repo (robbyrussell's oh-my-zsh).
+antigen bundle lukechilds/zsh-nvm
+antigen bundle git
+antigen bundle heroku
```
twalton83, line 174: I never asked for the instructions to include these tools.

twalton83 (changes requested):
Requested a few changes. This re-write of the instructions were not in the bounds of what the issue entailed and tacks on a bit too much re: antigen. We just needed a reference to those zsh files and anything else specific to the zsh installation, and we don’t want to send them to the node LTS directly. We want them to use nvm.

foundations/installations/installing_node.md
```diff
@@ -82,3 +87,112 @@ nvm use node
 Now when you run `node -v` you should see `v14.10.0` or something similar.
 
 If you see that, you have successfully installed node!
+
+
+# Installation on MacOS (zsh)
+
+**You can also install Node Using the macOS installer available from the Node.js [website](https://nodejs.org/en/download/)**
```
twalton83, line 94: We want them to use nvm, I believe this only gives them the LTS.

foundations/installations/installing_node.md
```diff
@@ -1,12 +1,17 @@
...
+Node uses JavaScript as its scripting language, and achieves high throughput via non-blocking I/O and a single-threaded event loop.
+
+There are many tools that require this to be installed. Testing frameworks and bundlers are two of many.
 
 We're going to install it using `nvm` (Node Version Manager), because it makes it easy to change node versions and upgrade node.  There is another tool called `npm` (Node Package Manager) that you will use later to install the various libraries and tools used in javascript environment.  It can be easy to confuse these two so read carefully!
 
 Node is also very easy to install using nvm, so this should go quickly :)
 
-### Step 0: Prerequisites For Linux
+# Installation in Linux
```
twalton83, line 12: The Linux instructions should also be followed by OS X users (we mention OS X in this section, and we assume that all students are on one of the two), why don’t we change this header to say Linux / OS X, and only mention the zsh_profile and zshrc on line 51 instead of breaking Mac OS X out into a separate section?

foundations/installations/installing_node.md
SidWorks: The previous definition was wrong! Node is not is runtime environment not an interpreter.
If one has done the foundations and JavaScript part of the project then those words won't be much of a problem.

twalton83: The problem is, these instructions come prior to the Javascript course for most folks. I don’t disagree with updating it to refer to it as a runtime though.

foundations/installations/installing_node.md
SidWorks: antigen is a package manager of zsh, doing things this way will be much easier for Mac users with zsh shell

twalton83: It’s out of scope, here.

foundations/installations/installing_node.md
SidWorks: replacing .bashrc and .bash_profile with the relevant zsh equivalents one may run into permissions issues, doing it this way one won’t have permissions issues when using NPM packages.
Thing listed will be required in later.

foundations/installations/installing_node.md
SidWorks: I guess ,it is is more cleaner this way!
using bash equivalent to zsh_profile and zshrc, one will get into permission issues .

foundations/installations/installing_node.md
SidWorks: users of mac catalina or above will get into many permission issues later.
Many have complained about it, 
Using a package manager is a robust way accomplish the task.

marvingay:
@SidWorks Without touching the same conversation in several places in the changes:

It's more sensible to stick to the [nvm documentation](https://github.com/nvm-sh/nvm). Adding a package manager to the installation is both unnecessary and creates a potential point of failure for new learners.

In regard to adding the appropriate installation instruction, really what's missing is:

`Since macOS 10.15, the default shell is zsh and nvm will look for .zshrc to update, none is installed by default. Create one with touch ~/.zshrc and run the install script again.`

That, in conjunction with another reminder to restart the terminal ought to be sufficient and can be tested before merging.

foundations/installations/installing_node.md
marvingay: I agree with @twalton83 in regards to this particular change.

"Node is a JavaScript runtime environment that allows you to run JavaScript outside of your web browser, which will prove necessary throughout your web development journey" or something similar.

I don't think it's worth introducing anything in this particular spot or at the very least it needs to be framed in a way that doesn't raise 3 more questions.

PR #22124, opened 2021-01-27
clarify purpose of SOLID article

twalton83:
- [X] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).
 - After noticing multiple students having trouble due to the use of Jquery in the Aspiring Craftsman article, I added a note explaining that we care for the concepts here, and not what the code itself does. 

#XXXXX


CatQueenCodes:
Good call!

PR #22138, opened 2021-02-06
Models inherit from ApplicationRecord

jdsutherland:
Rails 5 made the change that models inherit from `ApplicationRecord`. For reference: https://bigbinary.com/blog/application-record-in-rails-5.


rlmoser99 (approved):
Another great catch! Thank you for taking the time to make a PR.

PR #22150, opened 2021-02-13
Fix form_with call with url

jdsutherland:
Calling `form_with` with a url requires a `url` keyword arg.


rlmoser99 (approved):
Thanks again for updating this lesson. As you can tell there is quite a ripple effect with updating this lesson to `form_with`.

jdsutherland:
No problem. It turns out that making commits via browser isn't the most thorough way to do things.

PR #22151, opened 2021-02-14
Clarify form helper differences

jdsutherland:
Previously, I noticed some issues related to `form_with` changes and made several PRs to address them. Unfortunately, I made a bit of a mess of things in making a hasty smattering of PRs – these changes attempt to clear up some of the confusion I encountered when reading Form Basics for the first time.

These changes make the differences between the form helpers more explicit. While these differences aren't complicated, it previously required combing through an older version of Rails Guides to get into view.

Also of note: I made a [PR](https://github.com/TheOdinProject/curriculum/pull/22146) changing the Rails Guides URL because I thought it was mistakenly referencing an older version, but I've since realized that the latest version Rails Guides links to v5.2 because it is the last version to cover `form_for` and `form_tag`, so I've also fixed that. Hopefully this will prevent future confusion in explicitly mentioning that the link is to an older version of the guide.

### DIFF
--- a/rails_programming/forms_and_authentication/form_basics.md
+++ b/rails_programming/forms_and_authentication/form_basics.md
@@ -71,7 +71,7 @@ So, if you want to create your own form that gets handled by Rails, you need to 
   <input type="hidden" name="authenticity_token" value="<%= form_authenticity_token %>">
 ~~~
 
-### Making Forms into `params`
+### Making Forms into params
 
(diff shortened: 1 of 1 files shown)


rlmoser99 (approved):
Thank you for taking the time to write up this summary on the deprecated form helpers. I felt that it was a little long-winded, so I modified it a little bit to keep the focus on form_with.

PR #22159, opened 2021-02-20
Update project_linked_lists.md

BepisGithub:

I changed the wording of the #insert_at method that was under extra credit. It is now "#insert_at(value, index) that inserts the node with the provided value at the given index (it should not replace the node that currently occupies the given index)" I added the part in brackets.

#XXXXX


rlmoser99:
Thank you for taking the time to submit a PR. However, after reviewing your proposed changes, I think the additional wording is unnecessary. It would be more clear to just say "a new node" instead of "the node".

Therefore I can not accept this PR and will be closing it. I will be making a new PR to make the above change.

PR #22165, opened 2021-02-24
Update installation_overview.md

aenyeart:
- [+] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).

Corrects word choice error, from "understated" to "overstated"  (in second sentence @ line 43).
Reasoning: If something cannot be understated, that means there's virtually nothing to be said about it. It's clear from the context that the opposite meaning is intended, i.e. "there are many, many good things we could say about this".

#XXXXX


leila-alderman:
Nice catch!

Issue #22176, opened 2021-02-28
Add clarification to step 11 for installing and enabling guest additions


FriskySaga:
In regards to step 11 from here: https://theodinproject.com/lessons/prerequisites#step-3-install-and-enable-guest-additions

Here are the surrounding steps for context.

> 9. Click **Devices** on the VM toolbar -> **Insert Guest additions CD image** in the menu bar.
> 10. Wait for the CD image to mount, it will show the CD on the desktop as solid, not transparent, and a window will show on the top right of the VM screen saying it was successfully mounted.
> 11. ***Double-click on the CD icon on the VM desktop.***
> 12. In the new window that opens, right click on the white-space or any file/folder, and click **Open Terminal Here**.
> In the newly opened terminal window, paste `sudo ./VBoxLinuxAdditions.run` and hit enter.

For me, after completing steps 9 and 10, a File Manager window pops up on my screen. Instead of minimizing my open windows to navigate to the desktop, I can skip to step 12 and open a terminal from that File Manager window.

I suggest rewording step 11 to indicate that if the File Manager appears, then the reader can proceed to step 12 and open a Terminal window.

FriskySaga:
[I will address this issue on my branch.](https://github.com/TheOdinProject/curriculum/compare/master...FriskySaga:fix/install-guest-additions-step-11)

PR #22177, opened 2021-02-28
Add clarifications to step 11 for installing guest additions

FriskySaga:
- [X] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).

I updated step 11 for installing and enabling guest additions to be more clear. When I was following the guide, I initially found myself confused because I failed to recognize that to find the CD icon on the VM desktop, that meant I would need to literally navigate to the desktop folder.

#22176

### DIFF
--- a/foundations/installations/prerequisites.md
+++ b/foundations/installations/prerequisites.md
@@ -116,7 +116,7 @@ While your VM is running, do the following steps:
   8. Type `Y` when it asks you to and let it finish installing. Close the terminal when it is finished.
   9. Click **Devices** on the VM toolbar -> **Insert Guest additions CD image** in the menu bar.
   10. Wait for the CD image to mount, it will show the CD on the desktop as solid, not transparent, and a window will show on the top right of the VM screen saying it was successfully mounted.
-  11. Double-click on the CD icon on the VM desktop.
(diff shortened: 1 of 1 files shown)


xandora (approved):
Looks good to me, I think we can accept this with a minor change:

Most of the time the File Explorer window should pop up on top, perhaps just reword this slightly to say "in case the CD folder *doesn't* open, [go looking for it this way]".

xandora (commented):
whoops, please ignore.

FriskySaga:
> Looks good to me, I think we can accept this with a minor change:
>
> Most of the time the File Explorer window should pop up on top, perhaps just reword this slightly to say "in case the CD folder _doesn't_ open, [go looking for it this way]".

I took your suggestion to reword step 11.

I interpreted the suggestion to mean to write the "common case" first, meaning: "Hey user, if you see a File Manager window, then you're good."

Then I wrote the "plan B" case last, which is: "But hey, if you don't see a File Manager window, then go to the desktop."

Please review my updated code changes to see if I correctly understood the intent of your suggestion.

xandora (approved):
You nailed it. Looks good, approved.

PR #22194, opened 2021-03-14
Change passportJS URL in Authentication section

marboleda:
The two links in the Authentication section of the "NodeJS - Authentication Basics" lesson go to the same page.

The way the article is worded, it seems that the first link should be going to the home page of the passportJS website, not the specific documentation for LocalStrategy.

image


CatQueenCodes:

PR #22198, opened 2021-03-23
Add screen reader recomendation

EmilioAK:

One thing that really made me able to commit to starting this project was downloading a screen reader. I was not very used to reading articles like this before that, and found the fact that most resources are articles to be very discouraging. I think a screen reader recommendation could help lots of people who are in the same boat

#XXXXX


bycdiaz:
@EmilioAK I do agree that many new learners find the amount of reading challenging. I'm wondering if learners would be better served by a resource that can equip them to take on the amount of reading vs a resource that continues to put off reading. What do you think?

EmilioAK:
@bycdiaz If such a resource exists then sure, that would be great as an alternative! I don't think I agree that this counts as putting off reading though, since a screen reader will always be available whenever there is text for you to read. Personally I think putting in both would be a good idea, but if I had to chose then personally a screen reader has helped me more than any tips on improving reading skills I've encountered so far

PR #22200, opened 2021-03-23
Removes live server language

bycdiaz:
Removes live server language. This isn't super useful to a learner when they have no files to work with. It'll be more timely when we mention it again, at the Google Homepage project.


CatQueenCodes:
Moar greeeeen

PR #22201, opened 2021-03-24
Adds context

bycdiaz:
Gives a bit of context for the purpose of the live server vs code extension.


CatQueenCodes:
Get those green squares!

PR #22203, opened 2021-03-25
Add note that devtools resources are for chrome

CatQueenCodes:
Students seem to be trying to follow chrome tutorials in FireFox devtools. Added a note saying resources are to be followed in Chrome DevTools.

### DIFF
--- a/foundations/javascript_basics/developer_tools_2.md
+++ b/foundations/javascript_basics/developer_tools_2.md
@@ -15,6 +15,8 @@ There are three ways to open the Developer Tools menu:
 
 <div class="lesson-content__panel" markdown="1">
 
+**The following resources are meant to be done in Chrome browser unless stated otherwise.**
+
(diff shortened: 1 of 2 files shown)


rlmoser99 (changes requested):
I don't love adding important/clarifying notes like this, instead I like to see where that information would be the most useful and place it there. 

For the first lesson, the warning is before 2 videos & then looking at documentation. So I wonder if adding a note to the first video would be sufficient. Something like "If you want to follow along, be sure to use the same browser as indicated in the video."

For the second lesson, I would look at clarifying the assignment (Open your Chrome browser and head to the Chrome Dev Tool) and remove the "Opening Dev Tools" mention of Firefox. I think any mention of Firefox should come after the assignment, because I see this is the root cause of the confusion.

Both of my ideas are open to discussion since I may not have the complete picture.

CatQueenCodes:
@rlmoser99 I like your idea! I think we actually discussed in discord adding chrome to the installations and try and get everyone on chrome.

CatQueenCodes:
*I will get to this* I have to make a plan to change it in multiple places and add install chrome directions in the installations sections

marvingay:
Are we still continuing forward this change?

CatQueenCodes:
> Are we still continuing forward this change?

I do still need to do something with this. I will close it out as I think it needs a bit more thought

PR #22204, opened 2021-03-25
Clarify "front-end" is GUI

CatQueenCodes:
Students keep asking about why some RPS projects have a UI and if they should be doing it. Rephrased wording to make it more clear.


rlmoser99 (approved):
LGTM! Great idea to clarify this as it is such a common question.

PR #22213, opened 2021-03-28
Moved `sudo ./VBoxLinuxAdditions.run` on a new line.

ParthParolekar:
Moved `sudo ./VBoxLinuxAdditions.run` on a new line as it is written on two different lines. It is hard to recognize the space between "sudo" and "./VBoxLinuxAdditions.run" thus typing a wrong command in the terminal.

Moved `sudo ./VBoxLinuxAdditions.run` (step number 13 in the "Install and Enable Guest Additions" step) on a new line as it is written on two different lines. It is hard to recognize the space between "sudo" and "./VBoxLinuxAdditions.run" thus typing a wrong command in the terminal.

#XXXXX


xandora:
Thank you for your submission!

This issue seems to be specific to mobile users, so a change like this probably needs to be worked around in a different capacity than just adding a new line.

Mobile View:
image


Desktop View:
image


A better way might be to have the renderer force inline code to appear on a new line *if* it would wrap off the edge of a line.

Ultimately, I don't think this fix is going to improve the content sufficiently without making it look odd for the majority of our users.

PR #22215, opened 2021-03-29
Update project_rock_paper_scissors.md

fortypercenttitanium:

### Minor wording correction

Changes the sentence `Make your function case insensitive [...]` to `Make your function's playerSelection parameter case-insensitive [...]`

It might seem like a minor detail, but it is important to use the exact correct terminology in our lessons if we are going to expect our students to understand and use the correct terminology. One student in discord was confused by this and began googling "Javascript make functions case insensitive" - which, as anyone could guess, led down a misguided and unfruitful rabbit-hole.

n/a


CatQueenCodes:
Thanks for updating this!

PR #22221, opened 2021-03-31
Updated .dig example with the correct output

olliedavis:

I have amended the output to display NoMethodError rather than Nil. I believe this was the intended output here unless you were intending to append .dig.

#XXXXX

### DIFF
--- a/ruby_programming/basic_ruby/nested_collections.md
+++ b/ruby_programming/basic_ruby/nested_collections.md
@@ -241,7 +241,7 @@ vehicles[:zoe][:year]
 vehicles.dig(:zoe, :year)
 #=> nil
 vehicles[:alice][:color]
-#=> nil
+#=> NoMethodError
 vehicles.dig(:alice, :color)
 #=> nil
 ~~~


rlmoser99:
Thanks for taking the time to submit a PR. However, the nil value is what is returned for this example. Here is the output that I get on Ruby 2.7.2:

```
irb(main):001:1* vehicles = {
irb(main):002:1*   alice: {year: 2019, make: "Toyota", model: "Corolla"},
irb(main):003:1*   blake: {year: 2020, make: "Volkswagen", model: "Beetle"},
irb(main):004:1*   caleb: {year: 2020, make: "Honda", model: "Accord"}
irb(main):005:0> }
=> {:alice=>{:year=>2019, :make=>"Toyota", :model=>"Corolla"}, :blake=>{:year=>2020...
irb(main):006:0> vehicles[:alice][:color]
=> nil
irb(main):007:0> 
```

olliedavis:
Apologies, you are absolutely right. I guess my misunderstanding came from that I thought this would return NoMethodError if the nested hash' key didn't exist but re-reading it I can see that if the nested hash doesn't exist then it will raise a NoMethodError but if the nested hash exists and the specified nested hash key doesn't, it will return nil. Thanks for taking your time to clarify!

PR #22230, opened 2021-04-06
Tightens wording for those coming from Foundations

bycdiaz:
Just tightens up the language so that learners coming from Foundations know to skip the lesson.


CatQueenCodes:
This is amazing, thanks!

PR #22236, opened 2021-04-11
Added additional resources column

arbob:
I felt the .sort and .reduce methods were unclear in Wes Bos' video. So, I found MDN docs super helpful regarding that. I think this might help students in this course.

I felt the .sort and .reduce methods were unclear in Wes Bos' video. So, I found MDN docs super helpful regarding that. I think this might help students in this course.

#XXXXX

### DIFF
--- a/foundations/javascript_basics/fundamentals-5.md
+++ b/foundations/javascript_basics/fundamentals-5.md
@@ -28,3 +28,7 @@ At this point you just need a little more practice!  Go back to the [JavaScript 
 - findTheOldest
 - (Solutions for these exercises can be found in the 'solutions' branch of that repo)
 </div>
+
+### Additional Resources
+1. A primer on how [bubble sort](https://www.geeksforgeeks.org/bubble-sort/) algorithm works?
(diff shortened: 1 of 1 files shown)


rlmoser99:
I have not been helping people in Foundations for a while, but I do not think students need to understand the bubble sort algorithm in this lesson and I do not think a link to documentation should be an additional resource. 

I think it would be better to add a sentence in the "Array Magic" section of this lesson to remind student to look at documentation.

I'll ask others to review this PR to see if they have any other insight before deciding to recommend changes or even to close this PR.

twalton83:
I do not agree with adding a bubble sort resource here- you don't need to understand a bubble sort to understand how to use the `sort()` method, and I would instead opt to send students to the documentation.

I3uckwheat:
Agreed, to understand `.sort` all you need to know is how it compares values in the callback. Bubble sort is not something that a student needs to concern themselves with. This would fit better in a CS section, which we're working on, and covers this already.

PR #22238, opened 2021-04-12
corrected section about returning from procs

bschnitz:
I read the contribution guide. This contribution is not related to an existing issue.

Changes:
The explanation how return works inside a proc definition was not completely correct and could be extended a bit.
- corrected statement about return from top-level-context
- added information (and example) about which method it is, proc is returning from
- explained correctly when LocalJumpError occeurs when using procs
- emphasized difference of explicit/implicit return in proc context

### DIFF
--- a/ruby_programming/advanced_ruby/blocks.md
+++ b/ruby_programming/advanced_ruby/blocks.md
@@ -236,6 +236,14 @@ a_proc = Proc.new { puts "this is a proc" }
 a_proc.call
 ~~~
 
+Instead of using `Proc.new` you can also use the Kernel method `proc`.
+
+~~~ruby
+a_proc = proc { puts "this is a proc" }
+
+a_proc.call
+~~~
+
 Arguments are declared inside `||`
 
 ~~~ruby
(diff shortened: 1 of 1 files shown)


ruby_programming/advanced_ruby/blocks.md
```diff
@@ -236,6 +236,14 @@ a_proc = Proc.new { puts "this is a proc" }
 a_proc.call
 ~~~
 
+Instead of using `Proc.new` you can also use the Kernel method `proc`.
```
CouchofTomato, line 239: I like this addition.

ruby_programming/advanced_ruby/blocks.md
```diff
@@ -287,30 +295,40 @@ a_lambda.call
 # => 1
 ~~~
 
-A Proc object returns from the context in which it is called. If you are in the top level context (outside of a class and method) then you'll get an error because you can't return out of the very top level context, as there is no caller to return to.
+In non-lambda procs, `return` means exit from embracing method in which it was *defined* and will throw `LocalJumpError` if invoked outside the method. Check the [documentation](https://ruby-doc.org/core-2.7.0/Proc.html#class-Proc-label-Lambda+and+non-lambda+semantics) about it. In the top level context (not inside any method) a `return` will exit the program.
```
CouchofTomato, line 298: What do you mean in the top-level context a return will exit the program?

ruby_programming/advanced_ruby/blocks.md
```diff
@@ -287,30 +295,40 @@ a_lambda.call
...
 ~~~ruby
-a_proc = Proc.new { return }
+def execute_proc(proc_inside_method)
+  puts proc_inside_method.call # will produce no output!
+end
 
-a_proc.call
+def define_proc
+  proc_inside_method = proc { return 'explicit return' }
+  puts 'Calling the proc inside "execute_proc" will return from "define_proc".'
```
CouchofTomato, line 309: I get the point you are trying to make, but this isn't explained well enough for someone new to blocks.

ruby_programming/advanced_ruby/blocks.md
```diff
@@ -287,30 +295,40 @@ a_lambda.call
...
+  execute_proc(proc_inside_method)
+  puts 'This line is never reached.'
+end
 
-# => localJumpError (unexpected return)
+define_proc
```
CouchofTomato, line 314: Following the code from this method is confusing and a bit cluttered

ruby_programming/advanced_ruby/blocks.md
```diff
@@ -287,30 +295,40 @@ a_lambda.call
...
-def my_method
-  a_proc = Proc.new { return }
-  a_proc.call
-  puts "this line is never reached"
+def define_proc
+  proc { return 'explicit return' }
 end
 
-my_method
+define_proc.call # calling the proc, after define_proc exited
```
CouchofTomato, line 326: We would need to make clear here that the proc is implicitly returned as the last evaluated code in the method. It might seem obvious but when learning a new complex topic the fundamentals are often quickly forgotten. Just a comment saying the proc is implicitly returned would make it clearer.

CouchofTomato (changes requested):
Thanks @bschnitz

I appreciate the attempt to clear things up but I feel the code examples don't do a good job of getting the point across. Also the RUby Docs aren't very beginner friendly so terminology like 'embracing method' unless explained isn't a good addition.

With a bit of clean up we could consider merging this.

ruby_programming/advanced_ruby/blocks.md
bschnitz: In top-level-context a return will have the same effect as 'exit 0'.

Is that better, or is there some other thing unclear about it?

bschnitz:
> Thanks @bschnitz
> 
> I appreciate the attempt to clear things up but I feel the code examples don't do a good job of getting the point across. Also the Ruby Docs aren't very beginner friendly so terminology like 'embracing method' unless explained isn't a good addition.
> 
> With a bit of clean up we could consider merging this.

Thank You for reviewing!

ruby_programming/advanced_ruby/blocks.md
CouchofTomato: That's not something students will be very clear about

marvingay:
Closing due to the pull request becoming stale.

@bschnitz  If you wish to pursue these changes still, do not hesitate to make them and comment here or create a new pull request and referencing this one. Thank you for your time!

PR #22239, opened 2021-04-12
Changes suggestion of downloading 20.10 to 20.04.

bycdiaz:
There was a link pointing to Xubuntu 20.10. This will point to 20.04.


CatQueenCodes:
Wow thank you so much for doing this! You’re the best!

PR #22241, opened 2021-04-13
Update inputs-and-lists.md

DoviMaj:

Updated lesson to use the correct way of assigning keys to list items. Because I've seen a lot of people replicating the way the lesson taught and having bugs.

Docs: Keys should be stable, predictable, and unique. Unstable keys (like those produced by Math.random()) will cause many component instances and DOM nodes to be unnecessarily recreated, which can cause performance degradation and lost state in child components.

#XXXXX


marvingay (commented):
Good catch. Was there a reason we're removing semi-colons in various places but keeping them elsewhere?

marvingay (approved):

PR #22257, opened 2021-04-21
Remove reference to Javascript course

ChargrilledChook:
This conclusion is written as if Ruby is still the only track, and the Javascript module is next, but doesn't make sense for people on the Javascript track. I've removed the offending lines in this PR.


CatQueenCodes:
Good catch! Thank you for the edit!

Issue #22259, opened 2021-04-23
Update Foundation Lessons with Knowledge Checks


rlmoser99:
Background: Only some of the Foundation Lessons that have Learning Outcomes also have a section at the end of the lesson called Knowledge Checks. These knowledge checks can be very beneficial to students, therefore let's make all of the lessons that have Learning Outcomes also have Knowledge Checks.

Suggestions:
Ten of the Foundation lessons need to have a Knowledge Check section added. For reference, here is an example of a Foundation lesson with Learning Outcomes & Knowledge Check section:
https://www.theodinproject.com/paths/foundations/courses/foundations/lessons/problem-solving

There is a **possible exception** to the above pattern for two of the lessons. They have a Learning Outcomes section that is a combination of an older style of Knowledge Check. These two lessons can be updated to follow the above pattern or can follow the older style in the following lesson:
https://www.theodinproject.com/paths/foundations/courses/foundations/lessons/command-line-basics-web-development-101

Acceptance Criteria:
- [x] How Does the Web Work? (**possible exception**) -> _Assigned to @dylan-workhub_
- [x] Practicing Git Basics -> _Assigned to @dylan-workhub_
- [x] Introduction to the Front End -> _Assigned to @jaredmez_
- [x] HTML and CSS Basics -> _Assigned to @aaron-kidwell_
- [x] Developer Tools -> _Assigned to @dylan-workhub_
- [x] Fundamentals Part 1 -> _Assigned to @MC-Rawlings_
- [x] Fundamentals Part 2 -> _Assigned to @thatblindgeye_
- [x] DOM Manipulation (**possible exception**)  -> _Assigned to @dylan-workhub_
- [x] Introduction to Back End -> _Assigned to @aayushmau5_ 
- [x] Introduction to Frameworks -> _Assigned to @aayushmau5_ 

Lesson Assignments:
Before working on a lesson, please comment on this issue indicating what lesson(s) you would like to work on and review our contributing guide. 
https://github.com/TheOdinProject/curriculum/wiki/Contributing-Guide

aayushmau5:
I'm interested to work on this.

dylan-workhub:
I can take a crack at `Practicing Git Basics`.

Henry-Cantrell:
I'd be happy to take on `HTML and CSS basics`.

dylan-workhub:
If you'd like, I can also do `Developer Tools` tomorrow.

jaredmez:
[...]

PR #22268, opened 2021-04-27
added knowledge-check section to lesson

MC-Rawlings:
Added Knowledge-Check section to Fundamentals Part 1.

#22259


rlmoser99 (changes requested):
Thank you for taking the time to write this new section. This lesson is full of new information, so I think it would be good to include a few additional points that could get lost in this lesson. Feel free to adjust the wording & also put them in an order that makes sense to you.

Name the three ways to declare a variable?
What rules should you follow when naming variables?
Explain the difference between `==` and `===`.
Explain the difference between prefixing and post-fixing increment/decrement operators.

MC-Rawlings:
Thank you for the feedback :)
Will have these changes made within the next 24h

rlmoser99 (approved):
The order looks great! Thank you so much for taking the time to update this lesson with a knowledge check!

PR #22270, opened 2021-04-27
Added Knowledge Check to Introduction to the Front End in foundations

jaredmez:

Added Knowledge Check section under "Introduction to the Front End". 

#22259

### DIFF
--- a/foundations/the_front_end/introduction_to_the_front_end.md
+++ b/foundations/the_front_end/introduction_to_the_front_end.md
@@ -38,8 +38,13 @@ This section contains helpful links to other content. It isn't required, so cons
 ### Knowledge Check
 This section contains questions for you to check your understanding of this lesson. If you're having trouble answering the questions below on your own, review the material above to find the answer.
 
-* What does HTML stand for? 
-* What does CSS stand for? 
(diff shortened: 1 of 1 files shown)


rlmoser99 (changes requested):
Thank you for taking the time to work on this lesson's knowledge check! Your last point has a typo for the word 'responsible'.

In addition, the links in this lesson cover a lot of material and makes me wonder if we should include some additional points. I remember having to reference the "anatomy of a CSS ruleset" several times. Also, I know people have a tough time figuring out how to reference files & I think this important information gets lost in this lesson. What do you think about adding any of the following:

* What is the purpose of an opening tag and closing tag in an HTML element?
* What is the selector in a CSS ruleset?
* What is the property and property value in a CSS ruleset?
* What is the declaration in a CSS ruleset?
* How do you reference a file in the same directory as your HTML file?
* How do you reference a file in a directory below your HTML file?
* How do you reference a file in the directory above your HTML file?

jaredmez:
I completely agree, I wasn't sure how much depth to provide initially. I'll go ahead and add these.

rlmoser99 (approved):

PR #22272, opened 2021-04-28
Change learning outcomes, add knowledge checks

dylan-workhub:

Changed learning outcomes to match new style and added knowledge checks.

#22259

### DIFF
--- a/foundations/javascript_basics/DOM-manipulation.md
+++ b/foundations/javascript_basics/DOM-manipulation.md
@@ -3,115 +3,14 @@
 One of the most unique and useful abilities of JavaScript is its ability to manipulate the DOM. But what _is_ the DOM, and how do we go about changing it? Let's jump right in...
 
 ### Learning Outcomes
-<details>
-  <summary>What is DOM in relation to a webpage?</summary>
-  <ul>
-    <ul>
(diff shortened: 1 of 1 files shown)


rlmoser99 (changes requested):
Thank you for taking the time to update this lesson. I remember having to reference this lesson multiple times as I worked through the remaining Foundation lessons. I just re-read through the lesson and found another change that needs to be made, but in sticking with this original issue, I think we should add a few more points in the Knowledge Check.

The following points can easily get lost in the vast amount of information in this lesson and would be good to reiterate them in the Knowledge Check. Feel free to re-word them and put them in the order that makes the most sense to you.

* What is the difference between the return values of `querySelector` and `querySelectorAll`?
* What does a "nodelist" contain?
* Why are event listeners the preferred way to handle events?

dylan-workhub:
Changes have been made as requested. I couldn't think of a better way to word them than how you had put it! 

Thanks so much for this opportunity!

rlmoser99 (approved):
Thank you for updating this PR. After looking at the lesson again, I changed the order of the points, because an event listener is an answer to the 3 ways question. I really appreciate you taking the time to update this lesson!

PR #22273, opened 2021-04-29
Change wording of class_variables

cody-malcolm:
Per discussion in #ruby-help, April 28th, 2021

 - [ x ] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).
 - Sorry I didn't do the branch part, I just modified the markdown on Github (hopefully that is okay for a minor change like this). I _did_ check in a local editor that I didn't mess anything up with the markdown, though.

Just changed the wording of line 111 to fix an inaccuracy and draw a clearer distinction between global variables (which should never be used) and class variables (which have legitimate uses, even if such cases are rare and advanced). As mentioned above, the relevant discussion is in the Discord channel #ruby-help.

It is not.

### DIFF
--- a/ruby_programming/basic_ruby/variables.md
+++ b/ruby_programming/basic_ruby/variables.md
@@ -108,7 +108,7 @@ This example may be hard to completely understand at this point in the lesson. T
 ### Assignment
 <div class="lesson-content__panel" markdown="1">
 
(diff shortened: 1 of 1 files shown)


rlmoser99 (approved):
Thank you for taking the time to submit this change we discussed!

PR #22290, opened 2021-05-06
Clarify JWT wording for Node course

pintagon:

- Tone down advantages of JWT compared to sessions
- Clarify that JWTs are passed between backend/frontend not database/frontend

This section could be reworked - the comparisons here between JWTs and sessions are misleading and is highly contextual when to use either. It'd be beyond the scope of TOP to discuss those design choices and I believe the summary here can do more harm than good. For now, I've reworded a few phrases to make the section more accurate but my overall proposal would be to remove this page. Apologies if this is the wrong place to discuss, I can make a separate GH issue if preferred.

Additionally, none of the references linked discuss when or why to use either - a TOP student would be left with the impression that you use JWTs for APIs and sessions for MVC apps. Even among experienced devs it'd be highly contentions which exact situations have clear advantages for one or the other.

#XXXXX


nodeJS/APIs/API-Security.md
```diff
@@ -10,9 +10,9 @@ By the end of this lesson, you should be able to do or answer the following:
...
 
 ### Overview
-Securing your API is an important step. When we were using Express to serve view templates we used PassportJS along with a username and password to authenticate users, but that is not the only way to secure an Express app, and in the context of an API it often makes sense to use a different strategy.  The username and password session pattern that we learned previously will still work of course, though it is made a little more complicated by the fact that we've separated our front-end code from the back-end, so we have to specifically send our username/password info back and forth to check for authentication when accessing one of our endpoints.
+Securing your API is an important step. When we were using Express to serve view templates we used PassportJS along with a username and password to authenticate users, but that is not the only way to secure an Express app, and in the context of an API it often makes sense to use a different strategy.  The username and password session pattern that we learned previously will still work of course, though it is made a little more complicated by the fact that we've separated our front-end code from the back-end.
```
pintagon, line 13: >so we have to specifically send our username/password info back and forth to check for authentication when accessing one of our endpoints.

Not how sessions work - removed the phrase.

nodeJS/APIs/API-Security.md
```diff
@@ -10,9 +10,9 @@ By the end of this lesson, you should be able to do or answer the following:
...
 
-A better strategy is to generate and pass a secure **token** between our database and front-end code.  Doing so will make sure that our user's username and password are not compromised and will also give us the ability to expire our user's session for added security. The basic idea is that when a user signs in to our app, a secure token is created, and then for all subsequent requests that token is passed in the header of our request object. In the end, the process is pretty simple since you should already be pretty comfortable with using passport to authenticate users.
+Another strategy is to generate and pass a secure **token** between our back-end and front-end code.  Doing so will make sure that our user's username and password are not compromised and will also give us the ability to expire our user's session for added security. The basic idea is that when a user signs in to our app, a secure token is created, and then for all subsequent requests that token is passed in the header of our request object. In the end, the process is pretty simple since you should already be pretty comfortable with using passport to authenticate users.
```
pintagon, line 15: - Sessions / JWTs aren't strictly better or worse than each other. They're useful for different things. 
- Tokens are passed between backend/frontend. How JWTs get verified can use a db but that's not required.

>will also give us the ability to expire our user's session for added security

This is not unique to JWTs. Sessions within cookies accomplish the same thing.

>The basic idea is that when a user signs in to our app, a secure token is created, and then for all subsequent requests that token is passed in the header of our request object. 

This is a common way to pass JWTs but not the only way. You can also store and pass JWTs with a cookie.

marvingay (approved):
Agree with many of the sentiments commented on here. Open to more discussion in discord, since I believe you're present there as well.

Apologies for the long turnaround on this PR.

PR #22293, opened 2021-05-06
pluralize 'shipping_address' within nested form example

Tenacious-Qi:

The nested form example of `forms_advanced.md` contains `@user.shipping_address.build`, which results in a NoMethodError. I changed `@user.shipping_address.build` to the proper plural form `@user.shipping_addresses.build`. Because the association is `has_many :shipping_addresses`, when building the associated object, it must also be in plural form. 

#XXXXX


rlmoser99 (approved):
Good catch! Thank you for taking the time to submit a PR!

PR #22296, opened 2021-05-09
Add dropdown menus for easier navigation

kemonprogrammer:
- [x] I have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/master/CONTRIBUTING.md).

#### Describe the changes made:
- Added dropdown menus for Linux and macOS installation

#### Why the changes were made:
- The user has to choose between the 2 options, but will always install node
- If the user follows the Linux installation, he/she doesn't stop after the Linux part, because the user already saw the last heading


twalton83 (approved):
Thanks for doing this!

PR #22311, opened 2021-05-17
Update project_blog_app.md

bryanluu:

Added note to do everything in the project except for section 8.3. Section 9 seems to be do-able. It makes the articles destroy-able if they have comments...without section 9's code, the blog crashes when deleting articles with comments in them.

#XXXXX


rlmoser99 (changes requested):
Let's keep the wording to go through section 8.2, but rephrase the last sentence to say that section 9 and 10 are optional if you want to explore deleting or basic authentication.

bryanluu:
That makes sense. I can adjust it.

bryanluu:
Done! Check it out and let me know if anything else can be clarified. :)

rlmoser99 (changes requested):
Thanks for making this change so quickly. However, there are a few things that I am unsure about. The sentence that begins "Section 8.3 is causing..." is a long run on sentence. And some of the points are redundant 1) saying to go through 8.2 & skip 8.3 for now and 2)you can do 9 & 10 & saying that they are optional. 

I think we lose student's attention when we are too long winded, so let's shorten up this addition. I'm thinking something along these lines get the main points across:

... Section 8.3 currently has a few problems. However, if you want to explore additional functionality of deleting or basic authentication, you can do Section 9 and 10.

The more I think about this, I do not think we need to mention the "bug" because it is addressed in the remainder of the tutorial. Yes, it would be nice if it was addressed when comments were created, but I don't think it requires a special note.

bryanluu:
How about the current version? Made it shorter like you suggested.

rlmoser99 (approved):
Looks great! I really appreciate you taking the time to make these changes. I think this communicates the most important information & conveys that the tutorial is not complete at 8.2.

PR #22315, opened 2021-05-19
Lesson: Adds Ruby Project for Custom Enumerables

rlmoser99:
Because:
* This project was previously removed while block lesson was updated.

This commit:
* Updates the wording of the archived lesson.

The archived version of this project for reference:
https://github.com/TheOdinProject/curriculum/blob/main/archive/old_lessons/ruby/basic_ruby/project_advanced_building_blocks.md

I am thinking that it makes the most sense to add it at the end of the Advanced Ruby lessons.

#XXXXX


CouchofTomato (approved):
Yeah this seems fine. If we need to clarify anything it will be easier once we get feedback when it's live. This was a good challenge when we had it earlier 👍

Issue #22322, opened 2021-05-20
Remove Drop Down functionality from Knowledge Checks


rlmoser99:
**Before working on a lesson, please comment on this issue indicating what lesson that you would like to work on and wait to be assigned. In addition, please review our contributing guide:**
https://github.com/TheOdinProject/curriculum/wiki/Contributing-Guide

Background: 
There are 3 lessons in the beginning of Foundations that have Knowledge Checks with drop down functionality. We recently added 10 more knowledge checks to lessons in Foundations, using a new format that does not use drop downs. Because of these two differing formats, students occasionally report that these new knowledge checks do not work correctly. 

The drop downs made it too easy for students to just use the drop down and not think through the question, so we have implemented a new format that instructs them to review the lesson if they can not answer a question. 

For each of the following lessons:
* Update the wording to "This section contains questions for you to check your understanding of this lesson. If you’re having trouble answering the questions below on your own, review the material above to find the answer."
* Remove the drop down functionality from the Knowledge Checks, but keep the current questions.

Use the following lesson as a pattern to follow:
https://www.theodinproject.com/paths/foundations/courses/foundations/lessons/problem-solving

Acceptance Criteria:
- [x] Command Line Basics -> Assigned to @dylan-workhub 
- [x] Introduction to Git -> Assigned to @mdlugosz1
- [x] Git Basics -> Assigned to @7u1ian

7u1ian:
I'd like to tackle one of these please! By the way, is the acceptance criteria a list of which lessons need updating?

rlmoser99:
@7u1ian Yes, the lessons are listed in the acceptance criteria. Which one do you want to do?

7u1ian:
> @7u1ian Yes, the lessons are listed in the acceptance criteria. Which one do you want to do?

Git Basics please. Thank you!

dylan-workhub:
I can take command line basics.

mdlugosz1:
I can work on introduction to git.

7u1ian:
That was my first contribution. Let me know if everything went smoothly! By the way, is my name supposed to be assigned next to the item under Acceptance Criteria?

rlmoser99:
> That was my first contribution. Let me know if everything went smoothly! By the way, is my name supposed to be assigned next to the item under Acceptance Criteria?

Yes, it was. Sorry about that. I've fixed it. I'm not sure why/how that happened.

PR #22323, opened 2021-05-20
Remove drop down functionality from knowledge checks, reword the begi…

dylan-workhub:
…nning

Remove drop down functionality from knowledge checks, reword the beginning paragraph to fit new style.

#22322

### DIFF
--- a/foundations/installations/command_line_basics.md
+++ b/foundations/installations/command_line_basics.md
@@ -90,82 +90,16 @@ This section contains helpful links to other content. It isn't required, so cons
 * [Video Series from LearnLinuxTv](https://www.youtube.com/playlist?list=PLT98CRl2KxKHaKA9-4_I38sLzK134p4GJ) contains 24 videos explaining the basics of the command line. Videos are brief enough for beginners but, at the same time, are detailed enough to get started and light your inner curiosity.
 
 ### Knowledge Check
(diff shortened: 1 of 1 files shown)


foundations/installations/command_line_basics.md
```diff
@@ -90,82 +90,16 @@ This section contains helpful links to other content. It isn't required, so cons
...
-  <li>On Linux and Mac, use the <code>mv</code> command, e.g., <code>mv folder/old-file.txt folder/new-file.txt</code>.</li>
-</ul></ul>
-</details>
+This section contains questions for you to check your understanding of this lesson. If you’re having trouble answering the questions below on your own, review the material above to find the answer.
+
+* What is the command line?
+* How do you open the command line on your computer?
+* How can you navigate to a particular directory?
+* Where will <code>cd</code> on its own navigate you to?
+* Where will <code>cd ..</code> navigate you to?
```
rlmoser99, line 99: Instead of using `<code> and </code>` here, we can use backticks. We had to previously use `<code> and </code>` because it was inside li tags, but when used inside markdown, the backticks work perfectly!

rlmoser99 (changes requested):
Thanks for taking the time to work on this so quickly. If you could make the following change, that would be very helpful.

foundations/installations/command_line_basics.md
dylan-workhub: Just did this! Let me know if there's anything else needed.

rlmoser99 (approved):
Looks great! Thanks for making that change so quickly!

PR #22324, opened 2021-05-21
Remove drop down functionality

7u1ian:
Deleted answers but kept the questions under knowledge checks

In the Git Basics lesson of Foundations, removed drop down functionality in Knowledge Check section by deleting answers but keeping questions.

#22322

### DIFF
--- a/foundations/git_basics/git_basics.md
+++ b/foundations/git_basics/git_basics.md
@@ -72,5 +72,5 @@ This section contains questions for you to check your understanding of this less
 * What is the Git command used to view your commit history?
 * What is the Git command used to upload projects onto GitHub?
 * Explain the two-stage system that Git uses to save files.
-* Explain what <code>origin</code> is in <code>git push origin main</code>.
-* Explain what <code>main</code> is in <code>git push origin main</code>.
+* Explain what `origin` is in `git push origin main`.
+* Explain what `main` is in `git push origin main`.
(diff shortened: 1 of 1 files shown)


foundations/git_basics/git_basics.md
```diff
@@ -62,76 +62,15 @@ This section contains helpful links to other content. It isn't required, so cons
...
+* What is the Git command used to get a full copy of an existing Git repository from GitHub?
+* What is the Git command used to check the status of your files?
+* What is the Git command used to track files with Git?
+* What is the Git command used to remove tracked files with Git?
+* What is the Git command used to commit files?
+* What is the Git command used to view your commit history?
+* What is the Git command used to upload projects onto GitHub?
+* Explain the two-stage system that Git uses to save files.
+* Explain what <code>origin</code> is in <code>git push origin main</code>.
+* Explain what <code>main</code> is in <code>git push origin main</code>.
```
rlmoser99, line 76: Instead of using `<code> and </code>` here, we can use backticks. We had to previously use `<code> and </code>` because it was inside li tags, but when used inside markdown, the backticks work perfectly!

rlmoser99 (changes requested):
Thanks for taking the time to work on this so quickly. If you could make the following change, that would be very helpful.

foundations/git_basics/git_basics.md
7u1ian: Fixed it. Thanks for the review

rlmoser99 (approved):
Awesome! Thanks for making these changes so quickly!

PR #22327, opened 2021-05-21
Add new video explanation to assignments

Eduardo06sp:

I added a new video that I believe would be extremely helpful for students to better understand recursive methods, especially prior to tackling the Code Quiz on Recursion (in lesson: Recursive Methods). I also had to change the order of the assignments.

The author dives deeper into a couple of simple examples, breaking the code (JavaScript) down line-by-line, then explains a benefit of using recursion towards the end.

**EDIT:** I added another part of a video to drive the point home. I believe that this is a decent balance between being introduced to new concepts versus figuring things out yourself by diving straight in.


rlmoser99 (changes requested):
I do not feel like the video in the 4th bullet point is near as valuable as the one in the 3rd bullet point & do not want to add it to the lesson. 

In addition, I think the video in the 3rd bullet point, should come before the video that was already in the lesson (2nd bullet point). I think this additional video will help things click a little more during the CS50 video. Plus, I think it does a better job visualizing the factorial problem.

Eduardo06sp:
Thank you for the feedback. 

The reason I wanted to add that second link (by DevSage) was because it shows the "bubbling up" effect of recursion very clearly by showing how the return values change (versus how they stayed the same in Web Dev Simplified's explanation). 

However, using Web Dev's video as a precursor to the CS50 video makes a lot of sense. Once again, thank you.

rlmoser99 (approved):
Thanks for taking the time to make these changes so quickly! Looks great!

PR #22328, opened 2021-05-21
Clearly state expectations for Recursion Code Quiz

Eduardo06sp:

I added a little bit of clarification for the Recursion Code Quiz (in lesson: Recursive Methods). I mistakenly believed we had to read through the answers, as opposed to figure out the solution on our own.
That was mainly due to the last paragraph in the introduction:

> In this brief lesson, you’ll get a chance to learn more about when and how to use recursion and then in the next project you will get the chance to apply some of that (since it probably won’t really stick until you’ve had a chance to try it).

So I made the expectations a bit more clear.

### DIFF
--- a/ruby_programming/computer_science/lesson_recursion.md
+++ b/ruby_programming/computer_science/lesson_recursion.md
@@ -31,7 +31,7 @@ Look through these now and then use them to test yourself after doing the assign
 ### Test Yourself
 
 <div class="lesson-content__panel" markdown="1">
-  1. Complete the [Code Quiz](http://www.codequizzes.com/computer-science/beginner/recursion) on Recursion.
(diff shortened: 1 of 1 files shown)


rlmoser99 (changes requested):
If possible, I prefer to not add special notes at the end. Therefore, I would prefer this to re-worded to communicate that it will take time to complete. Maybe something like this: 

Solve each of the questions from this Code Quiz on Recursion and check your answers. You are not expected to answer them from memory.

I do not think we need to mention that this is a difficult concept to grasp because that could be said for almost every new concept.

Eduardo06sp:
I made the warning a bit more concise and removed the difficulty part.

rlmoser99 (approved):
Thanks (again) for making this change so quickly. This flows a lot better & makes the intent much clearer. Thank you for bringing this up in Discord and taking the time to make this change!

PR #22329, opened 2021-05-24
Clarify expected return value of first assignment (in Recursion Project lesson)

Eduardo06sp:

I reworded the first assignment a bit from:
> Write a method `#fibs` which takes a number and returns that many members of the fibonacci sequence. Use iteration for this solution.

to 

> Using iteration, write a method `#fibs` which takes a number and returns that many members of the fibonacci sequence. It should return an array.

This is to make it clear that it is not the exact same solution as the ones provided in the Code Quiz on Recursion nor the Khan Academy fibonacci walk-through under Additional Resources (in the previous lesson).

### DIFF
--- a/ruby_programming/computer_science/project_recursion.md
+++ b/ruby_programming/computer_science/project_recursion.md
@@ -5,7 +5,7 @@ The [Fibonacci Sequence](http://en.wikipedia.org/wiki/Fibonacci_number), which s
 ### Assignment 1
 
 <div class="lesson-content__panel" markdown="1">
-  1. Write a method `#fibs` which takes a number and returns that many members of the fibonacci sequence.  Use iteration for this solution.
+  1. Using iteration, write a method `#fibs` which takes a number and returns that many members of the fibonacci sequence. It should return an array.
   2. Now write another method `#fibs_rec` which solves the same problem recursively.  This can be done in just 3 lines (or 1 if you're crazy, but don't consider either of these lengths a requirement... just get it done).
 </div>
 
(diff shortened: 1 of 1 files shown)


ruby_programming/computer_science/project_recursion.md
```diff
@@ -5,7 +5,7 @@ The [Fibonacci Sequence](http://en.wikipedia.org/wiki/Fibonacci_number), which s
 ### Assignment 1
 
 <div class="lesson-content__panel" markdown="1">
-  1. Write a method `#fibs` which takes a number and returns that many members of the fibonacci sequence.  Use iteration for this solution.
+  1. Using iteration, write a method `#fibs` which takes a number and returns that many members of the fibonacci sequence. It should return an array.
```
wise-king-sullyman, line 8:
```suggestion
  1. Using iteration, write a method `#fibs` which takes a number and returns an array containing that many numbers from the fibonacci sequence.
```

wise-king-sullyman (changes requested):
Thank you for taking the time to do this PR!

You've definitely moved this in a better direction, but I'm requesting a small change in ordering/wording used.

The change I've suggested gets the input and output type closer together, and moves the key `from the fibonacci sequence` part to the end where it should be less likely that people read part of the instruction and jump to **doing** with the incorrect assumption that they know what the rest of the sentence will say. It also changes the `members of` phrasing to something that feels clearer to me.

ruby_programming/computer_science/project_recursion.md
Eduardo06sp: That makes perfect sense, thank you. I made the change.

wise-king-sullyman (approved):
Awesome! I think these changes will clear up confusion for a lot of people! :)

PR #22340, opened 2021-05-27
Chore: Add Registration Instructions to Project

rlmoser99:
Because:
* Our detailed instructions does not address this step.

This Commit:
* Adds instructions that allow a user to register for an event.
* Adds two optional extra credit steps.


CouchofTomato (approved):
Thanks Mosby

PR #22341, opened 2021-05-28
Thinking in React.

arbob:
Thought process before building any react application. This article helped me a lot in thinking about the react way of doing things. Hope it helps other students too.

...your text here

#XXXXX


CatQueenCodes:
I like this section of the React docs as well and found it helpful. I do think it would be beneficial to have it more towards the beginning of the react section before the first react project. Possibly in the additional resources of this section (right before CV project) https://www.theodinproject.com/paths/full-stack-javascript/courses/javascript/lessons/handle-inputs-and-render-lists

Anyone else have any opinions on that?

twalton83:
I agree with that sentiment.

CatQueenCodes (changes requested):
I like this section of the React docs as well and found it helpful. I do think it would be beneficial to have it more towards the beginning of the react section before the first react project. Possibly in the additional resources of this section (right before CV project) https://www.theodinproject.com/paths/full-stack-javascript/courses/javascript/lessons/handle-inputs-and-render-lists

arbob:
Agreed. I believe it'd be better to put it in the State and Props section of TOP.

PR #22345, opened 2021-05-30
feature: added stack overflow additional link

nluciano233:
Added a link to a stack overflow question that is useful to understand where to declare methods (in the prototype or in the constructor).


CatQueenCodes:

PR #22352, opened 2021-06-04
Update setting_up_git.md (MacOS Install Update)

VasantKesana:
Added command (Line 78-81)
brew upgrade git 
in case git version is less that 2.28. (MacOS)

Added command (Line 78-81)
brew upgrade git 
in case git version is less that 2.28. (MacOS)

#XXXXX

### DIFF
--- a/foundations/installations/setting_up_git.md
+++ b/foundations/installations/setting_up_git.md
@@ -73,7 +73,15 @@ This will install the latest version of Git. Easy, right?
 git --version
 ~~~
 
-If the version number is less than 2.28, follow the instructions again. If you are encountering a `no formulae found in taps` error:
+If the version number is less than 2.28, 
+
+~~~bash
+brew upgrade git
+~~~
+
+If this does not work, follow instructions again.
+
+If you are encountering a `no formulae found in taps` error:
 1. Run `brew doctor`
 2. You will see output similar to the below. NOTE: The actual output of `brew doctor` may vary based on the version of MacOS you're running, and any other issues you may have with your own installation. Ultimately, you must run each command line snippet that Homebrew provides after running `brew doctor` in order to repair your installation of Homebrew, including `brew cleanup` at the end.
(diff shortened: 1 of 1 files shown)


foundations/installations/setting_up_git.md
```diff
@@ -73,7 +73,15 @@ This will install the latest version of Git. Easy, right?
...
-If the version number is less than 2.28, follow the instructions again. If you are encountering a `no formulae found in taps` error:
+If the version number is less than 2.28, 
+
+~~~bash
+brew upgrade git
+~~~
+
+If this does not work, follow instructions again.
+
+If you are encountering a `no formulae found in taps` error:
```
twalton83, line 84: Thanks for contributing! I spoke with the team after getting some clarification about the installation process, and we're of the opinion that if Git didn't update with `brew install git`, because by default it will attempt to install the latest version, we are assuming `brew update git` is not going to do anything and that we should probably want to go to `brew doctor`.  

With that in mind, we can probably update this to instead say "If you are encountering a  `no formulae found in taps` error, or your version is still less than 2.28:" so we can catch those cases.

twalton83 (changes requested):

foundations/installations/setting_up_git.md
twalton83: Are you still interested in making this update?

VasantKesana: I am still interested in making this update. 
I had this particular issue when I was unable to fetch an update for git with `brew install git`. I ran into errors, which was mitigated with `brew upgrade git`.
Please note `brew update` and `brew upgrade` are different functions, and are not compatible with each other in my experience.

I feel that the beginner git user might not understand a `no formulae found in taps` error and I suggest the command should make it easier.

Please revert back to discuss further changes.

twalton83:
Closing, after discussion with the team, Homebrew should be deferring to the latest version on install, and if it doesn’t, it should be fixed with the `brew doctor` command.

PR #22355, opened 2021-06-05
Clarify return value of method #build_tree

Eduardo06sp:
Assignment 3 for binary search trees instructs you to create a build_tree method. It should return the "level-1 root node."

However, in the video "Binary tree traversal" provided in the lesson "Common Data Structure & Algorithms" states the top-most level as "level-0."

So I updated the assignment to fix this minor discrepancy.

**Change:**
> The #build_tree method should return the level-1 root node.

to

> The #build_tree method should return the **level-0** root node.


rlmoser99 (approved):
Great catch! Thanks for taking care of this so quickly, I really appreciate it!

PR #22368, opened 2021-06-09
Fix loop condition

HliebS:
Fix comparison meaning for clearer instructions


twalton83:
This isn’t correct. `i` is already “less than” answer, so we aren’t looping “until” it’s less than `answer`. This loop condition is saying “initialize `i` while `i` is less than OR equal to `answer`, increment `i` until it is greater than `answer`”.

PR #22370, opened 2021-06-11
Change numbering from 1. 1. 1. to 1. 2. 3.

DavidMatalik:

At the moment numbering doesn't work as expected. Instead of 1. 2. 3. on the page 1. 1. 1. is displayed. 
This is because how Kramdown formats stuff. In order that Kramdown formats here correctly, we need to add 3 spaces before each paragraph of a numbering section.


rlmoser99 (approved):
Thank you for taking the time to updating this lesson so that the numbers render correctly.

PR #22374, opened 2021-06-13
replace non-existent "componentDidUnmount"

hasref:
- [ x] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md).

The page on react hooks mentions `componentDidUnmount` as an example of a lifecycle method. As far as I can tell, react does not have a `componentDidUnmount` lifecycle method and there seems to be no mention of it in the react docs. I assume this was meant to be `componentWillUnmount`. See also this issue on the react github: https://github.com/facebook/react/issues/6424

### DIFF
--- a/javascript/react-js/hooks.md
+++ b/javascript/react-js/hooks.md
@@ -43,7 +43,7 @@ Afterwards we are declaring a function, which right now just sets a new count. I
 
 ### useEffect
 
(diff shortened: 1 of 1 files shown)


marvingay (approved):
Good catch.

PR #22375, opened 2021-06-13
Chore: Update number formatting for driver script

rlmoser99:
Because:
* The color code blocks were not working correctly.

This Commit:
* Removes indention for assignment numbering, similar to other lessons.
* Adds "tie it all together" section for driver script.

This screenshot is what it should look like after being rendered.
Screen Shot 2021-06-12 at 22 03 44 PM


wise-king-sullyman (approved):
Looks great!

PR #22379, opened 2021-06-14
Updated install to include bundler gem

Adub648:

In the next lesson (basic data types) bundler is required in order to properly use the practice exercises. While it may seem intuitive for some of us to install it, not having it installed already may confuse some users, so I have added an instruction for users to do it here.

#XXXXX

### DIFF
--- a/ruby_programming/introduction/installing_ruby.md
+++ b/ruby_programming/introduction/installing_ruby.md
@@ -124,6 +124,14 @@ The above command should return something similar to this:
 ruby 2.7.2pxx (20xx-xx-xx revision xxxxx) [x86_64-linux]
 ~~~
 where x represents the version available at the time you installed Ruby.
+ 
+#### Step 2.3: Install Ruby gems
+ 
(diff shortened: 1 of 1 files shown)


CouchofTomato:
thanks @Adub648 

We really appreciate contributions. As Ruby 2.7 comes with Bundler 2+ is there a reason you think it needs to be installed separately? Is it a case of making sure students are on the latest version?

Adub648:
Yeah so i've had a look at Discord and it seems that many people are having the same situation as me, I think that the version of Ruby that is being installed for TOP is not installing the latest version of bundler. 

I think that this will help some people who may get confused if we make sure that it's updated beforehand, as it may confuse some people who are not necessarily familiar with the command line.

wise-king-sullyman:
@Adub648 after trying it out on a clean install it looks like the issue you brought up is more with the ruby-exercises repo so I'm going to close this issue.

Thank you for bringing this to our attention though!

PR #22391, opened 2021-06-17
Make minor punctuation and wording edits to fundamentals-4.md

KaitlinKCheng:

Added missing punctuation.

Changed:
> **Before you write your code the test will fail**, and you should be able to know that your code works correctly when the **tests pass**.

To:
> **The test will fail before you write your code**, and you should be able to know that your code works correctly when the **test passes**.

N/A


CatQueenCodes:

PR #22397, opened 2021-06-23
added a homebrew step for Silicon machines

ManonLef:

Added a command for terminal for silicon machines to avoid 'brew command not found' error.

It seems that home-brew installs in a different directory on silicon machines compared to intel machines. I haven't tried it in bash but since zsh is the default I found this fix from https://stackoverflow.com/questions/20381128/installing-homebrew-on-os-x . I also found this YouTube with a possible explanation but frankly I'm not very experienced yet and not sure if what he says is correct. https://www.youtube.com/watch?v=2St60bsdAbg

#XXXXX

### DIFF
--- a/foundations/installations/setting_up_git.md
+++ b/foundations/installations/setting_up_git.md
@@ -55,6 +55,10 @@ First, you'll need to install Homebrew.  Make sure you have checked the requirem
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 ~~~
 
+Note: On an Apple Silicon Mac you will have an extra step to take.
+If you look at the terminal output after installing Homebrew, you will see "Installation Successful!". Further down in the terminal there will be a section called "Next steps". 
+This may all seem a bit intimidating but this is a great chance at overcoming those feelings. Follow the next steps as stated in your terminal (copy and paste the commands given) and you will make sure you will not encounter a "command not found: brew" error when doing the below Step 1.1: Update Git.
+
 #### Step 1.1: Update Git
 
 MacOS already comes with a version of Git, but you should update to the latest version. In the terminal, type
@@ -64,15 +68,6 @@ brew install git
 ~~~
 
 This will install the latest version of Git. Easy, right?
-
-Note: On an Apple Silicon Mac you might now encounter the following error: "brew command not found". This is due to Homebrew installing in a different directory than it is on an intel machine.
-To fix this issue, copy and paste the following into your terminal:
-
-~~~bash
-echo "export PATH=/opt/homebrew/bin:$PATH" >> ~/.zshrc
-~~~
-
-after reopening the terminal, the brew commands should work.
 
 #### Step 1.2: Verify version
 
--- a/foundations/installations/setting_up_git.md
+++ b/foundations/installations/setting_up_git.md
@@ -57,7 +57,7 @@ First, you'll need to install Homebrew.  Make sure you have checked the requirem
 
 Note: On an Apple Silicon Mac you will have an extra step to take.
 If you look at the terminal output after installing Homebrew, you will see "Installation Successful!". Further down in the terminal there will be a section called "Next steps". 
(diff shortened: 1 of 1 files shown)


xandora:
I'm not against putting this in the instructions, but it should be noted that the Homebrew installation process provides these instructions to the user near the end of the installation process.

Excerpt from https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh
```bash
if [[ "$UNAME_MACHINE" == "arm64" ]] || [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
  cat <<EOS
- Add Homebrew to your ${tty_bold}PATH${tty_reset} in ${tty_underline}${shell_profile}${tty_reset}:
    echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' >> ${shell_profile}
    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
EOS
fi
```

ManonLef:
Yes I agree. I also discussed this in the discord with mdfr4nk. The terminal output was what led me to google the next step since I wasn't sure what to add/replace myself. At that point in foundations, the terminal output will be very daunting to many a user so we should consider how to word it. I'd love to hear input on the best wording to add for emphasizing to read the terminal outputs or find solutions, keeping in mind the 'level' of students at that point in the curriculum.

ZachBaird:
Hey @ManonLef !

I think saying something like the following. It could be cleaned up, so I leave that to you:

> You're going to get a lot of output in the terminal, and it will be a little intimidating to you. However, this is a great exercise in overcoming that intimidation! If you read the terminal output, it will tell you "Installation Successful!". Further down will be a section called "Next steps". Follow those steps, and you'll be all set!

ManonLef:
Hi @ZachBaird !

Would you suggest adding that on top of the command I inserted in the text? Because I don't think there's a lot of output after that. Just a restart and then all set.

If you'd like this instead of what I put in there, I think the main issue would be that the output isn't as straightforward with providing instructions as I recall it. 

Here's a user that logged its output. Basically it's a warning "Warning: /opt/homebrew/bin is not in your PATH." https://github.com/Homebrew/brew/issues/10191
Which personally led me to google and end up at the homebrew support docs. However, those were still too complicated for me personally (not saying I'm a particularly smart cookie so perhaps you can better evaluate the knowledge students at this point should have). 

If needed I can try to uninstall and install again. Maybe even with an on-looker that's more experienced. Pretty sure there were no straight up instructions in the terminal to fix this issue at all.

twalton83:
For clarity @ManonLef, this is what the output looks like.

ManonLef:
@twalton83 thank you. I went ahead and uninstalled and reinstalled just to confirm. I don't think it completely uninstalled since every brew command worked out of the box now. However, it did give me the output exactly as you shared. 

Thank you all for your patience. I will rewrite it.

xandora:
@ManonLef uninstalling Homebrew won't remove the required exports from your `.bashprofile` or `.zshrc` files. Which is why it just worked out of box this time.

foundations/installations/setting_up_git.md
```diff
@@ -55,6 +55,10 @@ First, you'll need to install Homebrew.  Make sure you have checked the requirem
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 ~~~
 
+Note: On an Apple Silicon Mac you will have an extra step to take.
```
twalton83, line 58: I actually wonder if this is even an extra step for silicon devices. I think it might be relevant for all device.

ManonLef: With my limited knowledge, reading the following it seems like it's an extra step in arm64 setups; https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh

I feel like all this goes a bit over my head. But there definitely wasn't an extra step on my older iMac.

xandora: Yea, it's actually sorta weird how they don't handle this automatically for ARM64 macOS users, since the script *does* detect for it in several places already. You are right though, the "Add Homebrew to your PATH..." message will only appear for Apple Silicon users.

twalton83 (approved):

PR #22400, opened 2021-06-24
Fix object description

seoul2045:
Line 82, change mutable to immutable


rlmoser99 (approved):
Oh, wow. Great catch! Thanks for taking the time to make this PR.

PR #22404, opened 2021-06-27
Added how to secure API keys

arbob:
Securing API keys in github using dotenv

Hiding security keys from github.

#XXXXX


xandora:
Thank you for your submission. I have some concerns though, as this isn't really security, it's just obscuring the API key behind another variable. It isn't even that useful to students as their projects won't work when attempting to share a Github Page or submit their work through the Project Submission form.

I certainly think this needs to be communicated to students. However, I am hesitant to recommend this for inclusion in the curriculum as it misrepresent obscurity as security.

I3uckwheat:
I agree, and am closing this for this reason. *Anything* on the frontend is unsafe and exposed no matter how you look at it. 

The changes in this PR are not something we want to communicate in this project.

PR #22407, opened 2021-06-29
Remove firebase extra credit from library JS project

fortypercenttitanium:

I removed the Firebase extra credit from the library project and expanded a tiny bit on the `localStorage` option. I think this would improve the lesson because implementing Firebase at this point is a bit too advanced for where our students are, especially without more explicit instructions. Implementing the NPM package (many who have had issues in the discord try this route) is nearly impossible because webpack hasn't been introduced yet. To see these issues in action, search the discord JS channels for "firebase".

Implementing using the html scripts google gives you is possible, but still a bit too difficult. The student would also need to implement some sort of user database and authentication, and that too could get hairy. In any case, I think it's best we focus the extra credit for this lesson on using the Web Storage API. It's very accessible, teaches our students about serialization, and gives them a quick taste of JSON a little early. I think those are perfect extra credit goals.

n/a


marvingay:
The immediate push back that I'll give here is that the JavaScript course comes last in the Rails path.

Ultimately, while I do think it's viable for the Rails people to just use Rails at that point, I also think using another platform or solution provides a little it of exposure to other ways of doing things, if they wish, as it is currently just "extra credit".

Perhaps, it could even be changed in verbiage to be "optional" for the adventurous.

I agree with your suggestion, generally speaking, but will need to look more into why we carry this option to begin with as well as assessing the utility of it since it is extra credit and not required.

fortypercenttitanium:
> The immediate push back that I'll give here is that the JavaScript course comes last in the Rails path.
> 
> Ultimately, while I do think it's viable for the Rails people to just use Rails at that point, I also think using another platform or solution provides a little it of exposure to other ways of doing things, if they wish, as it is currently just "extra credit".
> 
> Perhaps, it could even be changed in verbiage to be "optional" for the adventurous.
> 
> I agree with your suggestion, generally speaking, but will need to look more into why we carry this option to begin with as well as assessing the utility of it since it is extra credit and not required.

Okay, I'll wait to hear back from you about it. If there is a conversation about it amongst the core team, I'd really like the opportunity to elaborate on my position. Also, in order to better evaluate the difficulty of this task better, it might be a good exercise to have one of the maintainers/core team try to implement this themselves in a toy project using only the resources learned to this point in the JS track (no webpack). That way if it is doable, we could at least provide one or two tips to those students.

marvingay:
@fortypercenttitanium  Ah yeah, I failed to mention that in my comment, this PR will get exposure to all the maintainers, so if they have a formal response, I will encourage them to provide that feedback here.

codyloyd:
I'm in favor of this proposal.  Thanks bendy.

PR #22408, opened 2021-06-29
React intro lesson improvements

fortypercenttitanium:

I cleaned up a lot of the grammar in this lesson and rewrote some content to make the lesson "flow" better. I also provided some more links and additional code examples to better explain some points.

### DIFF
--- a/javascript/react-js/react-introduction.md
+++ b/javascript/react-js/react-introduction.md
@@ -6,14 +6,14 @@ This section aims to provide you with all you need to know to get started with R
 
 React is one of the most powerful, widely used frontend frameworks.
 
-The landscape for frontend frameworks has been changing a lot over the last few years, so it is understandable to be scared of going for the "wrong" one.
+The landscape for frontend frameworks has been changing a lot over the last few years, so it is understandable to be worried about choosing the "wrong" one.
 [This article](https://stackoverflow.blog/2018/01/11/brutal-lifecycle-javascript-frameworks/) shows the recent development of frontend frameworks well. Once you start diving deeper into a framework, you will begin to love it. It makes your code easily scalable, more readable, and possibly a thousand times more efficient (just our modest estimation).
 
 Just to name a few reasons on why to learn React.js:
 
 - Reusability of components
 - Well supported due to its popularity
-- React is unopinionated, which means that it won't force you to follow any specific patterns or logic. It's all up to you.
+- React is not opinionated, which means that it won't force you to follow any specific design patterns, project organizational structure, or logic. It's all up to you.
 - Smaller learning curve, especially when you already have a good grasp of JavaScript and HTML from our previous lessons.
 
 If you need more convincing, the recent [2020 StackOverflow Developer Survey](https://insights.stackoverflow.com/survey/2020#technology-web-frameworks-all-respondents2) should be reason enough to get excited about learning React. Still unsure? Check out [this article](https://laptrinhx.com/top-10-reasons-why-you-should-learn-react-right-now-3718263995/).
@@ -33,17 +33,17 @@ For example, this simple website could be divided into the following components:
 - `MainArticle`, which will be the component that renders your main content.
 - `NewsletterForm`, which is a simple form that lets a user input their email to receive the weekly newsletter.
 
-In React, each component is defined in an ES6 (ECMAScript 2015) module. ES6 introduced the `import` statement which allows you to import code from one module into another module. This allows us to write each component in a separate file and import all components to the parent component like so:
+In React, each component is defined in an ES6 (ECMAScript 2015) module. ES6 introduced the `import` statement which allows you to import code from one module into another module. This allows us to write each component in a separate file and import all components into the parent component like so:
 
 `import ExampleComponent from "./components/ExampleComponent"`
 
 In our case, that parent would be App.js.
 
-In the beginning, it might be a little bit difficult to figure out the best component structure, especially when state and props come into play. This topic will be discussed in the following sections.  For now, don't worry too much about the component structure. Understanding of best practices comes with experience. React components, in general, usually have parent and/or child components. This system of structuring your applications helps to keep your code organized and makes it easy to keep track of your components' relationships with each other.
+In the beginning, it might be a little bit difficult to figure out the best component structure, especially when state and props come into play. This topic will be discussed in the following sections. For now, don't worry too much about the component structure. Understanding of best practices comes with experience. React components, in general, usually have parent and/or child components. This system of structuring your applications helps to keep your code organized and makes it easy to keep track of your components' relationships with each other.
 
 To give you an example of a basic component, see the following code:
 
-~~~javascript
+```javascript
 import React, { Component } from 'react'
 
 class App extends Component {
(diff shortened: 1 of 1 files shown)


javascript/react-js/react-introduction.md
```diff
@@ -6,14 +6,14 @@ This section aims to provide you with all you need to know to get started with R
 
 React is one of the most powerful, widely used frontend frameworks.
 
-The landscape for frontend frameworks has been changing a lot over the last few years, so it is understandable to be "scared" of going for the "wrong" one.
+The landscape for frontend frameworks has been changing a lot over the last few years, so it is understandable to be scared of going for the "wrong" one.
```
marvingay, line 9:
```suggestion
The landscape for frontend frameworks has been changing a lot over the last few years, so it is understandable to be worried about choosing the "wrong" one.
```

This is just my opinion to avoid that term altogether and still convey the point.

javascript/react-js/react-introduction.md
```diff
@@ -6,14 +6,14 @@ This section aims to provide you with all you need to know to get started with R
...
 [This article](https://stackoverflow.blog/2018/01/11/brutal-lifecycle-javascript-frameworks/) shows the recent development of frontend frameworks well. Once you start diving deeper into a framework, you will begin to love it. It makes your code easily scalable, more readable, and possibly a thousand times more efficient (just our modest estimation).
 
 Just to name a few reasons on why to learn React.js:
 
 - Reusability of components
-- Well supported, due to its popularity
-- React is unopinionated, which means that it won't force you to follow any specific patterns or logic, it's all up to you.
+- Well supported due to its popularity
+- React is unopinionated, which means that it won't force you to follow any specific patterns or logic. It's all up to you.
```
marvingay, line 16:
```suggestion
- React is not opinionated, which means that it won't force you to follow any specific design patterns, project organizational structure, or logic. It's all up to you.
```

javascript/react-js/react-introduction.md
```diff
@@ -22,21 +22,24 @@ _Convinced and excited? Great, because it's going to change your life! Let's div
...
+- `App`, which represents your main application and will be the parent of all other components. This would be the larger box surrounding the smaller ones.
+- `Navbar`, which will be the navigation bar.
+- `MainArticle`, which will be the component that renders your main content.
+- `NewsletterForm`, which is a simple form that lets a user input their email to receive the weekly newsletter.
 
-In React, each component is defined in an ES6 module. ES6 (ECMA2015) introduced the `import` statement; you can import code from a module into another module. This allows us to write each component in its own module in a separate file and later import all components to the parent component like so:
-`import ExampleComponent from "./components/ExampleComponent"`. (In our case that parent would be App.js)
+In React, each component is defined in an ES6 (ECMAScript 2015) module. ES6 introduced the `import` statement which allows you to import code from one module into another module. This allows us to write each component in a separate file and import all components to the parent component like so:
```
marvingay, line 36:
```suggestion
In React, each component is defined in an ES6 (ECMAScript 2015) module. ES6 introduced the `import` statement which allows you to import code from one module into another module. This allows us to write each component in a separate file and import all components into the parent component like so:
```

Minor change 'to' => 'into'.

javascript/react-js/react-introduction.md
```diff
@@ -62,7 +65,7 @@ class App extends Component {
 export default App
 ~~~
 
-Does the code look foreign? Don't worry, the structure isn't as difficult as it may look. Let's walk through it step by step.
+Does the code look foreign? Don't worry, it isn't as difficult as it looks. Let's walk through it step by step.
```
marvingay, line 68:
```suggestion
Does the code look foreign? Don't worry, it isn't as complicated as it looks. Let's walk through it step by step.
```

I think complicated (or complex) would probably be more suitable? I understand the original intent, but I wonder if my suggestion would be the more common interpretation of the code.

javascript/react-js/react-introduction.md
```diff
@@ -76,15 +79,15 @@ class App extends Component {
 }
 ~~~
 
-Secondly, we are declaring the class component, which is just a JavaScript class that extends the Component class we imported earlier. One thing to notice is that React components should always be declared with a capital letter at the beginning. This is a naming convention used by most developers and recommended by the React core team at Facebook.
+Secondly, we are declaring the class component App. We do this by extending the React class Component, which we imported at the top of the file. In doing this, we are essentially "Reactifying" our App component by giving it all of the fun methods and properties every react component should have. One thing to notice is that React components, like all classes and factory functions, should always be declared with a capital letter at the beginning ([PascalCase](https://techterms.com/definition/pascalcase)). This is a naming convention used by most developers and recommended by the React core team at Facebook.
```
marvingay, line 82:
```suggestion
Secondly, we are declaring the class component, `App`. We do this by extending the React class Component, which we imported at the top of the file. In doing this, we are essentially "Reactifying" our App component by giving it all of the fun methods and properties every React component should have. One thing to notice is that React components, like all classes and factory functions, should always be declared with a capital letter at the beginning ([PascalCase](https://techterms.com/definition/pascalcase)). This is a naming convention used by most developers and recommended by the React core team at Facebook.
```

Enclosed App in tildes such that it's emphasized in the same way a one line of code would be.
react => React for consistency

javascript/react-js/react-introduction.md
```diff
@@ -102,41 +105,89 @@ render() {
...
+The most unfamiliar code is likely the `render()` function, which returns something that looks like HTML, but is actually **JSX**. JSX is an HTML-like syntax that is "transpiled" (or converted) into JavaScript so a browser is able to process it. One of the primary characteristics and features of React is the ability to combine JavaScript and JSX. One thing you should know about JSX is that you can't use some JavaScript protected words as html attributes anymore, such as `class`, `onchange`, or `for`. Instead of `class` you will need to use `className`, instead of `onchange` you write `onChange`, and instead of `for`, you must use ... wait for it ... `htmlFor`. All attributes in JSX are written in camelCase and some have their names changed completely to avoid the transpiler getting too confused about whether you're assigning a label `for` an input, or instantiating a `for` loop. You should be fairly familiar with the camelCase naming convention from the naming of variables in JavaScript.
```
marvingay, line 108: Looks good.

javascript/react-js/react-introduction.md
```diff
@@ -102,41 +105,89 @@ render() {
...
+    ComponentC
+}
+```
+
+However, when you import them, they will each need to be wrapped in curly braces. If you export a component as a default, you can import it without the curly braces. If you export multiple components, you have to import them inside of curly brackets.
+
+If you're confused about this, take a minute to read up a little bit on ES6 [imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [exports](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export).
 
-So far, so good! We have already learned a lot about components in React, and let me tell you, class components are just one way of defining components. In React, another way is to use functional components.
+So far, so good! We have already learned a lot about components in React. Using class components is one of two ways of defining components in react. The other, more modern, approach is to define the component as a function (like a factory function).
```
marvingay, line 152:
```suggestion
So far, so good! We have already learned a lot about components in React. Using class components is one of two ways of defining components in React. The other, more modern, approach is to define the component as a function (like a factory function).
```
small change 
react => React

marvingay (changes requested):
Left a few suggestions/comments.

fortypercenttitanium:
> Left a few suggestions/comments.

Great suggestions. I implemented them.

marvingay (approved):

PR #22429, opened 2021-07-08
Adds read-only guidance.

bycdiaz:
A learner came by wondering if they should run the commands in the linked npm resources. @leosoaivan brought up that it would make sense for learners to just read through these resources and run commands when they get to their next project.


leosoaivan:
LGTM

PR #22436, opened 2021-07-12
Adds terminal simbology to the text where it tells to install pry

MLGabe:

Adds $ to helps make clear that gem install is supposed to be ran in the terminal.

N/A


ZachBaird:
Hey Gabe, I'm not sure this change is really necessary. The sentence reads "..you'll first need to install it in your terminal by running...", so we do think the intent is there.

That being said, if you find other instances of this where it's a bit more ambiguous, feel free to shoot up a PR. As an FYI, the standard is usually to do `$ gem command` with a space after the `$`.

PR #22444, opened 2021-07-14
Feature/a11y lessons

thatblindgeye:

Created section and lessons pertaining to web accessibility.

#XXXXX


html_css/web_accessibility.md
```diff
@@ -0,0 +1,70 @@
...
+By the end of this lesson, you should be able to:
+
+* Explain what web accessibility is.
+* Explain the purpose of the Web Content Accessibility Guidelines.
+* Name the 4 principles of the WCAG.
+* Know the differences between the WCAG conformity levels.
+
+### What is Web Accessibility?
+
+Web accessibility is when websites don't have barriers that prevent interaction, access, or contribution from users with a disability, situational limitation, or changing abilities due to age.
```
marvingay, line 17:
```suggestion
Web accessibility means that websites, tools, and technologies are designed and developed so that people with disabilities and other circumstantial limitations can use them.
```

Pulled from W3C and adding a portion of what you have here. Accessibility (and their standards) do sit on a spectrum so I believe it's more appropriate to avoid the "don't have barriers" as an absolute of sorts.

html_css/web_accessibility.md
```diff
@@ -0,0 +1,70 @@
...
+
+There are different types of disabilities, including (but not limited to) auditory, physical/motor, cognitive, or visual. A disability can be permanent, such as a user who is completely blind, or it can be temporary, such as a user with a broken arm, and users may have multiple disabilities. Older users with changing abilities may even have similar conditions as those who have a disability.
```
marvingay, line 19:
```suggestion
There are different types of disabilities, including (but not limited to) auditory, physical/motor, cognitive, or visual. A disability can be permanent, such as a user who is completely blind, or it can be temporary, such as a user with a broken arm. Users may have multiple disabilities at any given time. Older users with changing abilities may even have similar conditions as those who have a disability.
```

I broke the 'and users may have multiple disabilities' out into its own statement and added a few words for completeness.

html_css/web_accessibility.md
```diff
@@ -0,0 +1,70 @@
...
+
+Situational limitations are a little different. Using a phone outside on a bright day, browsing a site with one hand while you're busy doing something else with the other, or living in an area where the internet is slow or the bandwidth is expensive are all examples of situational limitations. These limitations, unlike a disability, occur to users only in specific situations, but are still important to keep in mind when developing a website.
+
+### Why Web Accessibility Matters
+
```
marvingay, line 24: I wonder if we should provide a more general, non-web related example of "accessibility" to introduce the point below.

marvingay (changes requested):
I meant to leave these as just normal comments, any further suggestions will do just that moving forward.

html_css/implementing_accessibility.md
```diff
@@ -0,0 +1,221 @@
...
+
+2. A dark theme isn't just taking the colors from the light theme and inverting them. Not only is this not 100% reliable in terms of color contrasts, but while light themes use shadows to simulate elevation for elements, dark themes instead rely on giving elements with a higher elevation a lighter background color. If you simply inverted the light theme colors, then the dark theme would have elements with white shadows, which doesn't simulate the elevation effect you should be going for (and it wouldn't really look good).
+
+### Keyboard Navigation
+
+Some users aren't able to use a mouse to navigate or operate a page and instead rely on using a keyboard or another assistive technology that can simulate keyboard inputs. Other users may even just prefer using a keyboard over a mouse, or may use a mix of both. These users require proper keyboard navigation, something that can easily be overlooked when developing a website.
```
marvingay, line 104: A question that would be raised here is: "What is proper keyboard navigation?"

Other sections are rather descriptive or have an example, can we come up with a tangible example of this to better explain the issue?

html_css/web_accessibility.md
thatblindgeye: I'm gonna reply to these individually just to keep things organized.

Definitely think you're right about the verbiage here, especially since it is mentioned in the lesson that follows that no site will be 100% accessible. What do you think of this, just adding a few words onto what you said:

"Web accessibility means that websites, tools, and technologies are designed and developed so that people with disabilities and other circumstantial limitations can use them with minimal barriers."? To me it comes off less as "you can't have ANYTHING inaccessible" and more "make things AS accessible as possible", but want to hear what you think.

marvingay: Seems fine to me.

html_css/web_accessibility.md
thatblindgeye: Could you clarify on this one a little more? Wasn't sure if you meant the "Why Web Accessibility Matters" or the "situational limitations" part for providing a non-web related example.

marvingay: This comment was intended for the section of "Why Web Accessibility Matters"

thatblindgeye: Ah gotcha. So perhaps instead of listing only web-related examples in the opening paragraph of that section ("text was too difficult to read..."), it might be good to open that section up with non-web examples that could maybe relate more broadly to people?

I will comment that my intention was to keep things web related as much as possible, but if the above is along the lines of what you mean then I could definitely see grounds for making that sort of change.

marvingay: Correct.

The reason being with the given example, the difference between "annoying" and "impossible" isn't yet clear for why the problems mentioned would make it impossible.

thatblindgeye: This is just off the top of my head currently:

"Imagine being in a multi-story building that has no elevator. For some people, this might only be an annoyance. "Huh, no elevator. I guess I'll walk up two flights of stairs, then." A person that requires a wheelchair, however, would find it impossible, or at the very least much more difficult, to go anywhere beyond the first floor. Even if a person in a wheelchair had someone to lift the wheelchair up each step, it would be a much more difficult process. The point is that an elevator would have made this building more accessible."

I3uckwheat (changes requested):
This looks very good, but I think I would like to see some videos of how different people with different disabilities navigate websites. It'd also be nice to see their frustration when they get to an inaccessible website.

html_css/aria.md
```diff
@@ -0,0 +1,276 @@
...
+The `aria-label` attribute overrides any native label and modifies the name property in the accessibility tree, though it's best used when an element doesn't already have a native label. When you add `aria-label` to an element, you pass in a string as the value, which will become that elements accessible name. `aria-label` doesn't work on every HTML element, though. Adding the attribute to a plain `<div>` or a `<span>` will have no effect, unless you give those elements a role of some kind.
+
+A common use for this attribute can be for the "close" button often seen in menus or modals:
+
+~~~html
+<button type='button' aria-label='Close modal'>X</button>
+~~~
+
+Instead of a screen reader announcing, "X, button", which wouldn't make any sense to the user, it would announce, "Close modal, button". The same thing can be done for the hamburger icon commonly used for mobile or hidden menus. Another way you could use `aria-label` is on landmark elements:
+
```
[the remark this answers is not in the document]
thatblindgeye: Did you mean mention the purpose of `id` in the opening paragraph, before the `aria-label` subsection? I do mention using the `id` for the `aria-labelledby` attribute in the subsection that follows, so just want to make sure if that's what you meant.

html_css/implementing_accessibility.md
```diff
@@ -0,0 +1,221 @@
...
+1. **Level AA** (minimum) requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text.
+2. **Level AAA** (enhanced) requires a contrast ration of at least 7:1 for normal text and 4.5:1 for large text.
+
+Both levels have exceptions that don't need to adhere to the required contrast ratio rule:
+
+* Incidental text, such as text that just happens to be within an image that has other significant visual content, or text that is purely decorative.
+* Text that is part of an inactive or disabled user interface component, such as a button that is disabled and has a lowered opacity.
+* Text that is part of a logo or brand name.
+
+[WebAIM Contrast Checker](https://webaim.org/resources/contrastchecker/) is a fantastic tool for checking the contrast ratio. The page also has a link for a link contrast checker, which goes over what the contrast ratio should be if a text link isn't underlined (its default style).
```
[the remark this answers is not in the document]
thatblindgeye: So my reasoning for not mentioning it here was more because I wanted to lump it in with the auditing lesson, since that goes into using the dev tools for checking stuff like this; sort of "here's a tool you can use" in this lesson and then follow up in a later lesson with "hey here's a better/alternative way", and also keep the dev tools stuff together. But I can see your point here. 

If you think quickly mentioning how to use the "element picker" tool in Chrome's dev tools to check contrasts here would make sense I could add that. My concern would kind of be twofold: either leaving Firefox out, or mentioning how to use the Firefox contrast tools as well and having to go over the same steps that are mentioned in the Auditing lesson (not that they're complicated steps or anything).

html_css/aria.md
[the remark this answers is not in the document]
I3uckwheat: I just think we need more information about when to use `id`. In the earlier courses we imply to not use `id` for nearly anything. If we're shaking it up for them here, it makes sense to explain when they should or should not use it. `id` is often abused, so some guidelines to how it relates to A11y would be great.

html_css/implementing_accessibility.md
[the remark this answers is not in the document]
I3uckwheat: No need to go into deep detail about it, but it's worth mentioning here for completeness sake. 

a. We'll get people telling us that it exists in the chat a lot if we don't
b. It's worth letting them do a bit of digging themselves to identify other tools they may not know about. It's OK to explicitly explain it later too.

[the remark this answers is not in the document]
thatblindgeye: Makes sense. Maybe something along the lines of:

"You can also check the contrast ratio in your browser's dev tools. In Chrome, you would click the "element picker" tool in the Elements tab, then hover over an element on the web page. In Firefox, you can do the same thing, except you have to click the "accessible object picker" tool in the Accessibility tab instead."

And of course updating some verbiage in the Auditing lesson.

html_css/aria.md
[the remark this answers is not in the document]
thatblindgeye: Maybe something like the following could be added to the opening paragraph? Might be a bit verbose, though.

"You may remember being warned against using the `id` attribute. Normally, you want to avoid overusing it or even using it when it isn't necessary (which more often than not, it isn't, and you should stick with using classes). There are several ARIA attributes, however, that actually *require* another element to have an `id`. 

When using such ARIA attributes, you would provide an `id` to one element, and you would then pass in that `id` value as another element's ARIA attribute value. This creates a link between the elements, similar to how a `<label>` element's `for` attribute creates a link to an input's `id`. Of the three types of ARIA labels mentioned below, both `aria-labelledby` and 'aria-describedby` are two of the ARIA attributes that require another element to be given an `id`."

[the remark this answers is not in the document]
I3uckwheat: Love it, great work!

html_css/implementing_accessibility.md
```diff
@@ -0,0 +1,221 @@
...
+  <div class='paper' tabindex='0'>Paper</div>
+  <div class='scissors' tabindex='0'>Scissors</div>
+</div>
+~~~
+
+A sighted user would most likely have no trouble playing the game if the elements looked like buttons to interact with. A screen reader user, however, would have no idea what any of these elements mean, nor would the user know that they can interact with them. The screen reader would announce the text contents of the element ("Rock"), and the user might think it's just plain text on the page and move on. There's no context to tell them that they're supposed to click on these elements. This issue can easily be fixed by using semantic HTML:
+
+~~~html
+<div class='button-container'>
+  <button class='rock' type='button'>Rock</button>
```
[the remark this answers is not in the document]
codyloyd: not necessarily 👀 

https://stackoverflow.com/questions/41904199/whats-the-point-of-button-type-button

doesn't really matter _here_... but it's not completely redundant.

codyloyd:
OK so overall this content is _great_.

MY concern: it feels like it might be a little TOO much at once.  There is a LOT to take in here, and it's not exactly clear which bits are the _very important_ ones.

This content IS there in the lessons, but I feel like we need a section of "at the very least check for these things", along with some of the quick tips for making sure they're usually handled right. (i.e. semantic HTML and just use a <button>)
- keyboard navigation
- screenreader announcement (i.e. no icon buttons, inputs without labels or images without alt-text)
- color contrast

I'm definitely not saying all of this isn't important, but there _is_ a line between "this site is completely inaccessible" and "I can definitely make this site better for disabled users".  Would be nice if we had a shorter section somewhere that basically says "if you do these easy things, your site will probably not be completely inaccessible, and all the rest of this stuff will just improve things"

codyloyd:
might be nice to have a section on "common pitfalls and how to avoid them"

Thinking things like:
- keyboard traps
- disappearing focus indicator (i.e. focus going into hidden menus)
- custom dropdowns/etc not being keyboard-accessible

thatblindgeye:
Made updates to the structure and content of lessons. Some notes:

1) Broke out the Implementing Accessibility lesson into 4 separate lessons. Not exactly bite sized chunks, but I think still easier to absorb the information than before.

2) In the Semantic HTML additional resources, I made a note about, "this is the only time where we're saying it's okay to use Windows in the scope of this curriculum" in regards to the NVDA screen reader. I only added this because a) NVDA is free and one of the more popular screen readers so it's one of the better screen reader options, and 2) it's a push to have people test their own projects out using a screen reader. I'm also weary about including this note, because obviously I don't want to give anyone the idea that "hey it's cool to use Windows now!"

3) I tried to make it clearer "this is the stuff you should start checking for/implementing", and anything additional I moved to the additional resources section (like themes).

4) Among the verbiage/content updates, I changed the "Messages" section of the Meaningful Text lesson to just be for Forms, since that was really all it went over in the first place. I think the knowledge from that section can easily be carried over to things like other alerts, status updates, etc. though.

html_css/web_accessibility.md
```diff
@@ -0,0 +1,78 @@
...
+Even though we're only introducing you to some of these a11y concepts, you may still worry that your site isn't accessible enough. The first thing to keep in mind is that **no site will ever be 100% accessible**, so don't try to aim for such an impossible goal. Sometimes the purpose or concept of a site even requires some things to not be accessible in certain ways.
+
+The second thing to keep in mind is that just taking those first few steps towards creating accessible websites matters just as much as the many more steps you'll take in the future. Even if you only add one a11y feature to your websites for now, what you may think is a minor addition could actually be a huge improvement for more users than you realize. So don't feel like you have to make everything perfect or that you need to add everything at once when implementing accessibility, especially when you're just starting out.
+
```
codyloyd, line 64: lets stick an assignment section at the bottom of this (newly broken out) lesson.

maybe have students check out the WCAG site https://www.w3.org/WAI/standards-guidelines/wcag/
and then link them to something like https://webaim.org/standards/wcag/checklist, which is a much nicer checklist.

You go over all this in later lessons, but being generally familiar with the guidelines and checklists might be helpful.

html_css/semantic_html.md
```diff
@@ -0,0 +1,94 @@
...
+* `<form>`
+* `<header>`
+* `<main>`
+* `<nav>`
+* `<section>`
+
+By properly using landmarks and headings, you provide users of assistive technologies a more operable and understandbale page: not only can screen readers navigate a page via landmarks and headings by opening a menu in their screen reader, but these elements also have their roles announced to provide additional context.
+
+If you were to use only `<div>` elements to act as these landmarks and headings, maybe adding in some CSS to visually style them, then a screen reader user would have to go through the entire page just to get to a specific section, and they may not be able to actually tell what is a heading or a landmark on the page. To see how a screen reader interacts with these elements, watch [Why headings and landmarks are so important](https://www.youtube.com/watch?v=vAAzdi1xuUY&list=PLNYkxOF6rcICWx0C9LVWWVqvHlYJyqw7g&index=19). This video also quickly mentions some things you *shouldn't* do with the heading elements.
+
```
codyloyd, line 81: lets add an assignment section here too.. just pull the links from above down here.

I3uckwheat:
@thatblindgeye 

This all looks good, there are a few spelling mistakes that the bot caught, I'll be happy to approve once those are fixed :D

html_css/keyboard_navigation.md
```diff
@@ -0,0 +1,103 @@
...
+<div tabindex="0">This is the second element listed in the HTML.</div>
+~~~
+
+Sometimes you may find it necessary to either change the visual order of elements on a page using CSS (the `float` or `order` properties, for example), or the tab order of elements themselves using the `tabindex` attribute. Regardless of which method you may use, you should make sure the tab order matches the visual order of elements. If the tab order is different from the visual order, users could be left confused or frustrated trying to navigate the page with a keyboard, expecting one element to receive focus based on the visual layout and instead another element receives focus. 
+
+The best way to avoid this issue is to just place elements in your HTMl file in the order that you want them to actually receive focus. Watch [What is Focus?](https://www.youtube.com/watch?v=EFv9ubbZLKw&list=PLNYkxOF6rcICWx0C9LVWWVqvHlYJyqw7g&index=3) to see some issues that occur when trying to change the tab order, then watch [Controlling focus with tabindex](https://www.youtube.com/watch?v=Pe0Ce1WtnUM&list=PLNYkxOF6rcICWx0C9LVWWVqvHlYJyqw7g&index=4) to see how using the `tabindex` attribute can affect the tab order.
```
codyloyd, line 78: stick this stuff down in an 'assignment' section

html_css/accessibility_auditing.md
```diff
@@ -0,0 +1,119 @@
...
+
+![Chrome's accessibility tree](https://user-images.githubusercontent.com/70952936/125674441-44095206-0508-45d0-ab2d-31ed4a2d00cb.jpg)
+
+The first thing you'll notice about the accessibility tree in Chrome is that it's laid out a little differently compared to Firefox. It still lists the role and name property of each object, but it isn't separated into two separate columns. Another difference is how you're able to view the tree itself. In Firefox, regardless of which object you select, you'll still be able to see the entire tree at once. In Chrome, you only see one branch at a time. If you clicked on the `contentinfo` object from the image above, for example, the two `generic` branches would no longer be visible.
+
+It is worth noting that you may be able to go into the Chrome dev tools settings and then into the **Experiments** section to check off the "Enable full accessibility tree view in the Elements panel" option. This will allow you to view an accessibility tree that is similar in style to Firefox's.
+
+### Firefox's Accessibility Properties
```
codyloyd, line 42: I'm a little torn on having both FF and Chrome here.... In general our curriculum just assumes Chrome.

In this case, I think they're different enough that it's worth keeping the FF stuff, BUT can we list chrome first, and then group the FF and Chrome stuff in their own sections instead of alternating? Students will just want to skip to the browser they're using... i.e.:
```
### Chrome
#### Accessibility Tree
#### Accessibility Properties
#### Checks

### Firefox
#### etc....
```

I3uckwheat (approved):

PR #22447, opened 2021-07-15
replaced require 'pry' with require 'pry-byebug'

LaithSaq:
with 'pry' alone I couldn't use the `next` command which is included in the same section.

replaced require 'pry' with require 'pry-byebug' because with 'pry' alone I couldn't use the `next` command which is included in the same section.

#XXXXX


wise-king-sullyman (approved):
Looks great, thank you for fixing this!

PR #22459, opened 2021-07-18
Add video about DOM

rosa-pastel:

I am actually here at TOP right now and I had trouble understanding what really DOM did. This video helped me get to the point where i felt ready to continue with the course, so I thought people like me who have never heard of DOM might also need it.


marvingay (changes requested):
Some elements of the video are covered in what follows where it's placed in this change.
Further, the creator of the video, for convenience conflates an HTMLCollection or Node List as an Array a couple of times which could be potentially confusing.

Upon inspecting our current additional resources, it would best fit there, or not at all. Let me know what you think!

rosa-pastel:
You're right, making people think of node lists as arrays could be misleading. It's better not to use the video at all, I guess.

PR #22490, opened 2021-07-27
Update How it works

mgrigoriev8109:

Added a paragraph to the How It Works section. Since a lot of students seem to be voicing the same concern over forgetting material after having finished FCC, and not feeling ready for the Google Project, it seems appropriate to explain that memorization is not an expectation, and that programming is very different from the educational model most people are used to (school, studying, notes, tests, failure/passing).

#XXXXX

### DIFF
--- a/foundations/introduction/how_this_course_will_work.md
+++ b/foundations/introduction/how_this_course_will_work.md
@@ -20,7 +20,7 @@ This curriculum works by pulling together the best content from across the inter
 
 Most lessons will contain questions that you should be able to answer before moving on.  Some of the lessons will ask you to complete exercises. In addition, we provide several projects throughout the curriculum to help you grow your understanding by actually building things.
 
-Try not too think of The Odin Project, or programming, as a class in school. It's not material you learn all at once, take a test, and then pass or fail. You can think of it as a snowball. You, yourself, are a snowball. You're rolling down a hill full of snow, and the further you roll, the more snow will stick to you. Sure, snow will also fall off of you, especially if it's not too wet and sticky, and you'll forget things often. But the more you keep rolling the more will stick to you, and the bigger and smarter you get.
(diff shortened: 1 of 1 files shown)


foundations/introduction/how_this_course_will_work.md
```diff
@@ -20,6 +20,8 @@ This curriculum works by pulling together the best content from across the inter
 
 Most lessons will contain questions that you should be able to answer before moving on.  Some of the lessons will ask you to complete exercises. In addition, we provide several projects throughout the curriculum to help you grow your understanding by actually building things.
 
+Try not too think of The Odin Project, or programming, as a class in school. It's not material you learn all at once, take a test, and then pass or fail. You can think of it as a snowball. You, yourself, are a snowball. You're rolling down a hill full of snow, and the further you roll, the more snow will stick to you. Sure, snow will also fall off of you, especially if it's not too wet and sticky, and you'll forget things often. But the more you keep rolling the more will stick to you, and the bigger and smarter you get.
```
twalton83, line 23:
```suggestion
Try not to think of The Odin Project, or programming, as a class in school. It's not material you learn all at once, take a test, and then pass or fail. You can think of it as a snowball. You, yourself, are a snowball. You're rolling down a hill full of snow, and the further you roll, the more snow will stick to you. Sure, snow will also fall off of you, especially if it's not too wet and sticky, and you'll forget things often. But the more you keep rolling the more will stick to you, and the bigger and smarter you get.
```

twalton83: Love the metaphor, but can we bring it back to literal terms and emphasize that memorization is not a goal?

twalton83 (changes requested):

foundations/introduction/how_this_course_will_work.md
ZachBaird: I agree with @twalton83 tbh, we have a lot of students think they need to memorize everything before moving on, and addressing that this isn't true could be really helpful.

twalton83: My suggestion was dismissed, but you have a typo still. It should be “to” not “too”.

twalton83 (approved):

PR #22504, opened 2021-08-04
Add additional content to Private Events instructions

msespos:

Add one line to the Private Events instructions indicating that allowing a user to become an event attendee will not only require a controller and corresponding routes, but also an interface in the view(s) for the user to indicate that they want to attend an event.

I think that this line was a bit of a missing piece, and that adding it adds more specificity to the instructions while hopefully not giving too much away still.

Not related to any open issues


rlmoser99 (commented):
I'm not sure I love adding on a second sentence because I think it reiterates the same info.

Do you think the following sentence would get the point across?
`Create a Controller with corresponding routes and views for the "through" table that will allow a user to become an "attendee" of an event.`

msespos:
Sure, that works too. The main thing for me was to indicate that the user will need to add a view / add to an existing view to complete this part. So in the interest of keeping it succinct, I think your revision works. Thanks!

rlmoser99:
@msespos I have reviewed all the instructions for this project and I am not sure if my suggestion is the best move. I am going to review this PR with other maintainers before we move forward.

rlmoser99 (approved):
After discussing this with other maintainers, we have decided to go with your original working. Thank you so much for taking the time to submit this PR and improve these instructions.

PR #22505, opened 2021-08-05
Update fundamentals-1.md

twalton83:

Changes the order of content so students can test examples they will read about more easily.

#XXXXX

### DIFF
--- a/foundations/javascript_basics/fundamentals-1.md
+++ b/foundations/javascript_basics/fundamentals-1.md
@@ -19,7 +19,7 @@ Look through these now and then use them to test yourself after doing the assign
 
 ### How to Run JavaScript Code
 
-All JavaScript we will be writing for the forseeable future will be ran via the browser. Later lessons in Foundations and the NodeJS path will show you how to run JavaScript outside of the browser environment. Keep in mind that due to differences in the NodeJS environment and the browser environment, for now you should always default to running your JavaScript in the browser unless otherwise specified.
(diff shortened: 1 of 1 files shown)


foundations/javascript_basics/fundamentals-1.md
```diff
@@ -17,25 +17,11 @@ Look through these now and then use them to test yourself after doing the assign
...
-2. [This MDN article](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Math) covers the same info from a slightly different point of view. There's much more that you can do with numbers, but this is all you need at the moment.
-3. Read through \(and code along with!\) [this article](http://javascript.info/operators) about operators in Javascript.  Don't forget to do the "Tasks" at the bottom of the page!  It will give you a pretty good idea of what you can accomplish with numbers (among other things!) in JavaScript.
+### How to Run JavaScript Code
 
-### Practice!
+All JavaScript we will be writing for the forseeable future will be ran via the browser. Later lessons in Foundations and the NodeJS path will show you how to run JavaScript outside of the browser environment. Keep in mind that due to differences in the NodeJS environment and the browser environment, for now you should always default to running your JavaScript in the browser unless otherwise specified.
```
twalton83, line 22: As a note, I added this bit because some people do try to use Node to run browser code via Code Runner and don't understand why they can't use `prompt()` or similar browser methods.

I3uckwheat: I'm concerned that the jest section will be confusing for people after reading this.

I3uckwheat (changes requested):
I left one comment, there's also a few typos that github actions caught

twalton83:
The check seems to be successful? 

Also, as far as the comment, that's why I added that part, because I knew shortly after we would be taking them to a Node environment, and the Jest stuff is "unless otherwise specified".

wise-king-sullyman (approved):
LGTM!

I3uckwheat (approved):

PR #22522, opened 2021-08-12
Delete React imports from router.md

DavidMatalik:
- [x ] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md).

For the demonstration purposes in this chapter React import is not necessary in every file. JSX gets also transformed without it, since React 17.

#XXXXX


wise-king-sullyman (approved):
I think the intent of these is to be followed along with rather than just for demonstration, but since students will be making a new CRA app for this and as you said importing React is no longer needed I think this is a good change. Thanks for doing it!

PR #22546, opened 2021-08-20
Add clarifying sentence with link on first step 6

dm-murphy:

The link in step 6 goes to the start of Chapter 4 which immediately shows the `belongs_to` association and `#build_association` method. This method will not work and students need to go down and find the `#build` association method from the `has_many` association section. 

This fix adds another line about finding the right `#build` association method from this Rails Guide link. Along with that reminder it also moves the link to this new sentence to help students think about checking for the right method and hopefully avoid confusion. By keeping the exact same link to the start of chapter 4 instead of linking to the `has_many` section it prevents spoilers on which association students should be using for this model.

#XXXXX


rlmoser99 (approved):
Thanks for taking the time to clarify these instructions!

PR #22549, opened 2021-08-20
Feature/css foundations

thatblindgeye:

Wrote the CSS Foundations lesson for [TheOdinProject/top-meta/issues/19](https://github.com/TheOdinProject/top-meta/issues/19). All criteria are met.

#XXXXX


html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+* Add CSS to an HTML file
+* Add a class and an ID to an HTML element
+* Use element, class, and ID selectors
+* Group multiple selectors and use combinators
+* Understand what the cascade is
+* Understand how specificity, inheritance, and rule order work
+
+### Basic Syntax
+
+At the most basic level, CSS is made up of various rules. These rules are made up of a selector (more on these in a bit) and a semi-colon separated list of declarations, with each of those declarations being made up of a property:value pair. The actual syntax will be slightly different depending on which method is used for adding CSS, but generally each method will all follow this structure.
```
codyloyd, line 18: IMO an image here that shows the basic syntax would be helpful.

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+* Understand what the cascade is
+* Understand how specificity, inheritance, and rule order work
+
+### Basic Syntax
+
+At the most basic level, CSS is made up of various rules. These rules are made up of a selector (more on these in a bit) and a semi-colon separated list of declarations, with each of those declarations being made up of a property:value pair. The actual syntax will be slightly different depending on which method is used for adding CSS, but generally each method will all follow this structure. 
+
+So, what exactly are these methods?
+
+#### Inline CSS
```
codyloyd, line 22: two things here:
1. I think I might prefer going straight to `selectors` and saving "how do you get your CSS into your HTML" until the end.  I'm unsure about this though so take it or leave it.

2. Put the Inline CSS option last. I don't like starting the first lesson straight away with something we're going to tell them not to use LOL.

I'd word it something like this:
> #### Inline CSS
> it _is_ technically possible to stick styles directly in your html, but this approach is not recommended..... (followed by your actual content)

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+
+Let's go over the differences here compared to inline CSS. 
+
+First, this is where we have to declare a selector (the `div` and `p` bits). Then we have opening and closing curly braces after the selector, which create a "declaration block". Finally, we place any declarations inside of the declaration block. If we wanted to add other rules, we would just continue adding them inside of the opening and closing `<style>` tags.
+
+This method can be useful for adding unique styles to a *single page* of a website, but it still isn't what many consider to be the preferred method.
+
+#### External CSS
+
+External CSS, aka the preferred method, involves creating a separate file for the CSS and linking it inside of an HTML's opening and closing `<head>` tags:
```
codyloyd, line 68: > External CSS, aka the preferred method
seems kinda snarky lol

maybe try something like
> The most common method of writing CSS is to put it in a separate file and then link that file to your HTML.

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+First, we add a self-closing `<link>` element inside of the opening and closing `<head>` tags of the HTML file. The `href` attribute is the location of the CSS file, either an absolute URL or, what you'll be utilizing, a URL relative to the location of the HTML file.
+
+Then inside of the newly created `styles.css` file, you'll notice the syntax is basically the same as the internal CSS method. The difference here being we no longer need those `<style>` tags (in fact, there should be no HTML tags within an external CSS file). `styles.css` is just what we went with as the file name here. You can name the file whatever you want as long as the file type is `.css`, though "style" or "styles" is most commonly used.
+
+Why is this the preferred method?
+
+1. It keeps our HTML and CSS separated, which results in the HTML file being smaller and makes things look cleaner
+2. We only need to edit the CSS in *one* place, which is especially handy for websites with multiple pages that all have similar styles.
+
+### Selectors
```
codyloyd, line 99: as i mentioned above... it might be nice to go straight to 'selectors' after the intro.  Seems more in line with the first paragraph that mentions syntax details (i.e. css is made up of a selector and a property and a value leads well into "so now let's learn about selectors, and then properties and values)

but i won't put up a fight either way.. starting with how to link CSS to HTML is also valid.

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+div {
+  color: white;
+}
+~~~
+
+Here, all three `<div>` elements would be selected, while the `<p>` element wouldn't be.
+
+#### Class Selectors
+
+Class selectors will select all elements with the given class, which is just an attribute you place on an HTML element. Here's how you would add a class in an HTML file and select it in CSS:
```
codyloyd, line 137: super nit: remove the word `would`

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+/* styles.css */
+
+.alert-text {
+  color: red;
+}
+~~~
+
+Note the syntax for class selectors: a period immediately followed by the case-sensitive value of the class attribute. Classes aren't required to be unique, so you can use the same class on as many elements as you want.
+
+Another thing you may have noticed is the kebab-case style. We separate a multi-worded class with hyphens because whitespace is reserved for separating a list of several classes on a single element, such as `class="alert-text severe-alert"`. Kebab-case isn't exactly required, but it tends to be the convention that developers opt for.
```
codyloyd, line 156: I think i'd leave out the kebab-case mention here.. this is the very first coding our students have done so they probably aren't ready for kebabs or snakes or camels or whatever lol.  I'd just say something like "don't put spaces in there, use hyphens" and leave it at that for the moment.

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+When it comes to internal and external CSS, each type of selector has its own specificity level (there are other selectors that make up the following list, but we're focusing on the ones mentioned in this lesson):
+
+1. ID selectors (most specific, after inline CSS)
+2. Class selectors
+3. Type selectors
+4. Universal selector and combinators (no specificity)
+
+Specificity will only be taken into account when an element has multiple rules targeting it, sort of like a tie-breaker. Ignoring universal selectors and combinators, the other types of selectors will increase the specificity of a rule for that selector's level.
+
+Specificity is represented by a comma-separated list of numbers: `0, 0, 0`. Starting from the left, each number represents the three specificity levels (id, class, and type), with each number having a higher precedence than the next (ID beats all, class beats type and no specificity, type beats anything left over).
```
codyloyd, line 289: I'm trying to decide if the scorecard is too far here.  On one hand, it might be more appropriate to give a couple of easy-to-remember rules like "ids beat classes beat types" and "more classes is more specific"

BUT: knowing exactly how it works is also valuable and I'm torn lol.

I think when it comes down to it... I personally never refer to or try to calculate the actual specificity score when i'm doing dev so I'd probably leave it out here.

I think if i had written this I might go with the simpler rules, and then say something like:
> if you want to know _exactly_ how specificity is calculated check this link here.

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+#parent {
+  color: red;
+}
+
+.child {
+  color: blue
+}
+~~~
+
+Despite the `parent` element having a higher specificity with an ID, the `child` element would have the `color: blue` style applied since that declaration directly targets it, while `color: red` from the parent is only inherited.
```
codyloyd, line 336: I'd add a note that says something like 
> typography related properties (font, color, letter-spacing etc.) usually inherit, but most other things do not.

Just so people have some idea of what to expect when it comes to inheritance.

html_css/v2/foundations/css-foundations.md
```diff
@@ -0,0 +1,425 @@
...
+
+### The Cascade of CSS
+
+Sometimes we may have rules that conflict with one another, and we end up with some unexpected results. "But I wanted *these* paragraphs to be blue, why are they red like these other paragraphs?!" As frustrating as this can be, it's important to understand that CSS doesn't just *do* things against our wishes. CSS only does what we tell it to do. One exception to this is the default styles that are provided by a browser. These default styles vary from browser to browser, and they are why some elements create a large "gap" between itself and other elements, or why buttons look the way they do, despite us not writing any CSS rules to style them that way.
+
+So if you end up with some unexpected behavior like this it's either because of these default styles, or due to not understanding how a property works or not understanding this little thing called the cascade. 
+
+The cascade is what determines which rules actually get applied to our HTML. There are different factors that the cascade uses to determine this, three of which we'll go over to hopefully help you avoid (as many of) those frustrating "I hate CSS" moments.
+
+#### Specificity
```
codyloyd, line 276: I'd probably put specificity and inheritance after the 'properties' section... like start off with learning the selectors and properties then go into this deeper stuff.

codyloyd (changes requested):
lots of little nitpicky stuff here, but overall it's good content.  Thanks for taking this on 💪🏽

html_css/v2/foundations/css-foundations.md
thatblindgeye: 1. I get what you mean. I'll mull that over because it can also make sense to order things in that way. It's kind of a "you should know how to add CSS to your HTML before learning what you can do with CSS" vs "how can you add CSS to your HTML when you don't know what you can do with it" sort of thing.

2. Would you think the order should just be reversed completely? External -> Internal -> Inline, so it's more "Here's the most common method. Here's another way you could do it, and also a third way but it's not recommended."

html_css/v2/foundations/css-foundations.md
thatblindgeye: The main reason I included the whole scorecard thing was an attempt to maybe simplify specificty for people at first. "100 is higher than 10, so ID wins", then once they get further along they won't need to use that anymore. Though at the same time yeah I do get that introducing something like this that people will most likely not need after long may not make much sense.

"X beats Y and Y, even if there are 100 Y's and only 1 X" or something would get the same point across.

codyloyd (approved):

PR #22553, opened 2021-08-22
Update developer_tools_2.md

Zufallszahl:

In the _Developer Tools 2_ lesson I added an option for people who primarily use Firefox. Chrome DevTools docs are great, but as someone who uses Firefox I had to figure out where some of the things described there were, because FF DevTools are slightly different.

It's not related to any open issues.


wise-king-sullyman:
After talking it over with the team we want to push students towards learning the dev tool basics with Chrome here at the moment.

Whether we want to keep the focus on Chrome long term or try and go in a more browser agnostic direction in the future is something we will need to discuss more; however, I will be closing this PR, as if we decide to do so that will require more extensive changes. 

Thank you for submitting this PR though! You've brought up a good point.

PR #22561, opened 2021-08-26
add addtional resume builder link

superspike7:
#### 1.Describe the changes made: Add new additional link to a awesome (totally free) resume builder

I found this gem https://flowcv.io/ and I think this will be a great resume builder.


rlmoser99 (approved):
That resource looks very interesting. I like that it has tips that you can look at as you write your resume. Thanks for taking the time to add it to this lesson.

PR #22570, opened 2021-08-31
Lesson/inspecting html css

twalton83:
closes https://github.com/TheOdinProject/top-meta/issues/20


html_css/v2/foundations/inspecting-html-and-css/inspecting-html-and-css.md
```diff
@@ -0,0 +1,62 @@
+# Inspecting HTML and CSS
```
codyloyd, line 1: remove this top level heading.. it's inserted in the page from the seeds file

codyloyd (approved):

html_css/v2/foundations/inspecting-html-and-css/inspecting-html-and-css.md
```diff
@@ -0,0 +1,60 @@
...
+### Testing Styles In The Inspector
+
+The Styles pane also allows you to edit styles directly in the browser. You can click inside of any individual selector to add a new rule, or click on an existing attribute or value to alter it.
+
+In the below image, we have altered the value of `margin-bottom` in the `.hero__main-heading` class, and the webpage responds with the changes in real time. This won't affect your source code in your text editor, but it is extremely useful to quickly test out various attributes and values without needing to reload the page over and over again.
+
+### Assignment
+
+<div class="lesson-content__panel" markdown="1">
+1. Skim through the [official Chrome DevTools docs](https://developers.google.com/web/tools/chrome-devtools):
```
thatblindgeye, line 38: Really minor thing, but I think the verbiage might be better as "Go through the following sections of [official Google DevTools docs]...", then add further comments like the last bullet point. Just to make it explicitly clear that it's important for students to take this info in. Maybe something like:

```suggestion
1. Go through the following sections of the [official Chrome DevTools docs](...):

- "Overview": don't navigate to any other pages linked here, just get familiar with *what* tools are available in the DevTools rather than how to use all of them right now.
- "Open Chrome DevTools": similar to what we went over above, but with some nice extras.
- "CSS": be sure to follow along with any interactive instructions!
- "Get Started With Viewing And Changing The DOM": skim through any part that uses the JavaScript console.
```

Regardless, I think the lesson looks good! It's really just a nitpicky semantic argument over "skimming", which if the point *is* for students to skim through stuff for now and revisit as they begin using devtools more, then ignore everything except "looks good!"

KevinMulhern (approved):
Looks great 🔥

PR #22587, opened 2021-09-05
update router review article 

mattdimicelli:
…vers React Router 4

 - [Yes ] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md).

I updated the link to a "review article" for React Router to a newer article from the same author, which covers React Router v4.  The previous article was re: React Router v3.  The new article compares and contrasts the differences between the two versions, which are important.  

As a beginner I feel that this new article is didactic and useful.  The only part of the article that I see as a bit "out-of-scope" compared to the rest of the content in the TOP Router lesson is the paragraph on "Authorized Route", but this doesn't detract from the usefulness of the article.

#XXXXX

### DIFF
--- a/javascript/react-js/router.md
+++ b/javascript/react-js/router.md
@@ -109,4 +109,4 @@ You should now have enough basics to get started with React routing. There are a
 ### Additional Resources
 
 - [This video by The Net Ninja](https://www.youtube.com/watch?v=QUz3k2O3ZJU&ab_channel=TheNetNinja) is an excellent review of what we learned here.
-- For some extra practice/review, check out [this article](https://css-tricks.com/learning-react-router/).
+- For some extra practice/review, check out [this article](https://https://css-tricks.com/react-router-4/).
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (commented):
It doesn't seem like this really covers the same material that the article you're replacing does.

Additionally, I'm not sure that a resource comparing one version of a tool to another, when we just introduced said tool, is for the best.

I will need to discuss with the team and I'll get back with you.

mattdimicelli:
Hi Austin,

Understood.  What prompted me to read this article was the author's own
warning at the top of the previous article.  The warning dissuaded me from
reading the first article, because it suggested that it was outdated.

wise-king-sullyman (changes requested):
If you don't know of a resource that covers the same content for the current version of React Router I think it would be best to link to the React Router docs.

mattdimicelli:
No, I don't know of any other resource.

But no need to link to the React Router docs bc in the Router article it's
already linked to as a primary learning source.  This article was just for
the "supplementary learning".

wise-king-sullyman:
@mattdimicelli would you like to change your PR to just remove the linked V3 resource then?

mattdimicelli:
Well, no, bc at least for me the article was useful.

Since I don't have another resource to suggest, and the content of the new
article has deemed non-equivalent/innapropriate.... maybe just delete the
pull request?

wise-king-sullyman:
Ah alright, you replace the existing resource with the V4 vs V3 comparison made me think you didn't get much value from the existing resource, but understood. I'll close this PR.

PR #22619, opened 2021-09-09
Making context of sharing screenshots clearer.

linkonsat:
Changing it to clarify that if your sharing a screenshot that you should also share the respective code in a repository. Rather than just solely having a screenshot when asking for help. Might make it clearer that if there is correlating code to share that in a repository rather than just a screenshot?

Changed the text on line 42 to basically state that if your sharing a screenshot of the output of your code that putting the code in repl.it is more beneficial than just the screenshot and that it makes it easier for people helping if they can see the correlating code from the screenshot in a repository.

#XXXXX


foundations/introduction/join_the_odin_community.md
```diff
@@ -39,7 +39,7 @@ When asking your question, please remember to include the context:
 
 If you can't pinpoint the problem, you can share a **screenshot**. This is especially useful for showing the output of commands in the command line. In Discord, drag and drop your screenshot image file into the chat box to upload it or simply use the PrtScn and paste keyboard shortcuts. If you don't know how to take a screenshot on your computer, this is a good time to ask Google.
 
-If the puzzle still can't be solved, push your project to GitHub or put your code in a [replit](https://replit.com) so that others can comb through and debug the code. You'll learn all about GitHub very soon.
+If you do share a screenshot of the output of your code, make sure to push your project to GitHub or put your corresponding code in a [replit](https://replit.com) so that others can comb through and debug the code. The screenshot of the output and the correlating code will help make it easier to understand the problem for people helping you. You'll learn all about GitHub very soon.
```
CouchofTomato, line 42: I think the wording here is a little confusing about when is a good time to use a screenshot. The aim of this section should be to make clear that when seeking help, the person should use all of the tools at their disposal to give anyone helping them enough information. That may include a screenshot for things like terminal outputs or visual design problems but it will often need corresponding code file(s) and if possible code that is runnable and reproduces the problem.

linkonsat: I updated it with the situations where a screenshot could be useful. I can't think of any other situations where a screenshot would be useful. I did include it to state to always include the corresponding files with the respective error. I think it will help make it clearer when to use the screenshot.

twalton83 (approved):

PR #22624, opened 2021-09-09
Update username++ text

Maheshkumar-novice:
There is a space between `username` and `++`  => `username ++` , This won't work with the bot because we are using `space` after `username`. So it's updated to `username++` , so new users won't wonder why it's not working sometimes.

### DIFF
--- a/foundations/introduction/join_the_odin_community.md
+++ b/foundations/introduction/join_the_odin_community.md
@@ -75,7 +75,7 @@ Your Multiple Lines of Colorful Code
 
 * Have fun with giphys: type `/giphy hi` to say hi to everyone.
 * Type `/help` for more information on chat commands.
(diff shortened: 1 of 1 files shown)


CouchofTomato:
Thanks @Maheshkumar-novice 

The points command works with or without one space between the username as `++`

Maheshkumar-novice:
> Thanks @Maheshkumar-novice
> 
> The points command works with or without one space between the username as `++`

Yeah it works but with the default space inserted by discord after we type the username. But if intentionally put the space like `username` `space` `++` it won't work.

Maheshkumar-novice:
> Thanks @Maheshkumar-novice
> 
> The points command works with or without one space between the username as `++`

Got it Thanks!

PR #22642, opened 2021-09-12
Update DOM-manipulation.md

C19H3R:

Improved listing , using GitHub markdown 
from " a. b. c.  "    >>   to " 1. 2. 3. 4. 5. "
In the practices section , Point no.s  : 2, 3, 4


twalton83:
The alternating between numbers and letters is typical with this type of notation as it infers that the alternate notation is a sub point of the “higher level” notation.

C19H3R:
ok got it but my concern was :
Github Markdown doesn't support a list with letters/alphabets, so there's a difference in the markdown output.

Example one:-

1. head one
    1. sub-one
    sub-one continued
    2. sub-two
  
2. head-two
   a. sub-one
   sub-one continued
   b. sub-two


Example two:-

Screenshot 2021-09-12 191928
Screenshot 2021-09-12 192818

twalton83:
I think in this case our major concern will always be the rendering on the site, which you can always check at theodinproject.com/lessons/preview ! :) we do have quite a few files where github doesn’t play nice with our markdown (links, for example), so we usually defer to what renders well on the site.

PR #22645, opened 2021-09-12
fix typo in "working with text"

jktham:

The html comment syntax explanation used an em dash instead of a hyphen, causing issues when copy-pasting, as well as an incorrect indefinite article.

Edit: Turns out the a/an thing is mostly up to pronunciation, but an seems to be more popular.

#XXXXX


thatblindgeye:
Thanks for the fix, @jktham !

PR #22652, opened 2021-09-14
fix typo

YiannisRoss:
'will be ran' -> 'will be run'

 - [ X] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md).

fixed typo on 'run'

#XXXXX


rlmoser99 (approved):
@YiannisRoss Thanks for taking the time to make this change. You are right. I had to research this change and learned that when using future tense, it should be "run" instead of "ran".

PR #22670, opened 2021-09-17
Add new resource and remove unnecessary ones from the coding test questions resources

hkrsmk:
Added LeetCode, Kattis, and HackerEarth to the list of coding test questions resources.

Added new resources to the list of coding test questions.

Not related.


rlmoser99:
@hkrsmk Thank you for taking the time to submit this PR. However, I am not sure that we want to add all 3 of these to our curriculum. I do not think the purpose of this list is to provide comprehensive list of options, but rather ones that we recommend. For each one that you would like added, I would like to know why you recommend it and what is different about it than the ones that are already on the list.

hkrsmk:
Of course!

1. For LeetCode I really like that you don't have to sign in to view the code, submit answers or see the solution, which isn't found in Code Wars or Hacker Rank.
2. I would say Kattis is most like Project Euler, but with more interesting questions alongside CPU and memory limits, which focuses coders on efficiency more so than what I've seen in other programming problem websites.
3. HackerEarth - offers support for bash, visual basic, and swift, which can be hard to find elsewhere. Though Code Wars is covering pretty good ground here too.

To be very honest, I'm also not particularly familiar with any of these platforms; these are the differences I can see so far without creating an account on all of these sites.

rlmoser99 (changes requested):
Thanks for taking the time to submit an overview of these resources. I took a quick look at them, as well as the ones that are currently recommended. 

I will only approve adding Leetcode at this time. I do not want to add HackerEarth or Kattis, because people going through our curriculum will know JavaScript and/or Ruby, so recommending sites that do not offer those languages seems a little weird to me. Along those same lines, I think we should remove the Coding Bat resource.

hkrsmk:
I believe both [Kattis](https://open.kattis.com/help) and [HackerEarth](https://help.hackerearth.com/hc/en-us/articles/4402703137945-Supported-browsers-and-languages) support Nodejs and Ruby.

I agree with you about removing Coding Bats, though - maybe I can do so when I update this PR.

rlmoser99:
@hkrsmk I see, thanks for providing those links. In the few minutes that I browsed their sites, I did not see those options. However, the decision to not include them is only in part because I did not think they supported JS or Ruby. The other reasoning is that I do not think we need to provide a comprehensive list, rather ones that our team of maintainers can personally recommend. 

In addition, you stated that you are "'not particularly familiar with any of these platforms", so I take it that you can not personally recommend them either. However, I have personally used leetcode and although I do not recommend it because during none of my interviews were I given a leetcode type problem (more like codewars type problems), I am part of a interview prep group (in another online community) that utilizes this platform and they strongly recommend this site.

hkrsmk:
@rlmoser99 Have updated the PR, could you take a look? Removed the other two resources and also coding bats.

rlmoser99 (approved):
Thank you for taking the time to make these additional changes!

PR #22675, opened 2021-09-18
Update inspecting-html-and-css.md

godsarmor:

fix to a very minor typo

#XXXXX


twalton83:
I believe “skim” was intended here, but “skip” may be appropriate. I’ll let others chime in.

godsarmor:
@twalton83, it may be somewhat confusing for those who are non-native English speakers.

thatblindgeye:
If the intention is for people to not read parts regarding the JavaScript console, then "Skip" would make more sense, though I would actually suggest saying "Ignore any part..." instead of "Skip through..."

If the intention is for people to read over the JS parts but not pay as much attention to it, then "Skim" makes more sense (maybe tweak the verbiage to make it clearer; "The JavaScript console is also covered, but that will be brought up in a later lesson and the sections regarding HTML and CSS are more important." or something). Although, I would say telling people to ignore the JS parts right now might be better than skimming through them if the main point of the lessons is HTML and CSS inspection.

twalton83 (approved):

PR #22699, opened 2021-09-22
Feature/css rewrite selectors

ChargrilledChook:
closes https://github.com/TheOdinProject/top-meta/issues/31 with a new lesson


html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+  img[src="puppy.jpg"] {
+    /* This will target img elements with a src attribute that is exactly "puppy.jpg" */
+  }
+~~~
+
+Sometimes we need to be more general in how we access these attributes. For example, perhaps we're only interested in `img` elements where the `src` attribute's value ends in `.jpg`. For cases like this we have some attribute selectors that allow us to match a part of the attribute's value. If you've ever come across [regular expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) before, these attributes use a similar syntax.
+
+
+* `[attribute^="value"]` - `^=` Will match strings from the start.
+* `[attribute$="value"]` - `$=` Will strings from the end.
```
Zadag, line 207: I think line 207 should say "Will match strings from the start"

Zadag (commented):
I think there's a typo on line 207.  Overall I like the lesson

html_css/v2/main-course/selectors.md
ChargrilledChook: Good catch, thanks!

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+  /* Our cool CSS */
+}
+
+/* This rule will only select divs with a class of grand-child */
+main > div > div {
+  /* More cool CSS */
+}
+~~~
+
+Phrased another way, the child selector will select an element that is one level of indentation down. In order to select an element that is adjacent to our target, or on the same level of indentation, we can use the adjacent sibling combinator `+`.
```
mgrigoriev8109, line 61: Two proposed edits to line 61:

First, it currently says the child selector will select an element that is one level of indentation down. Unless I misunderstand, child selector will select *all* elements matched by the second selector, that are one level of indentation down. 

Second, I think that phonetically the second sentence on this line would flow better with the first comma removed. So it could be:

In order to select an element that is adjacent to our target or on the same level of indentation, we can use the adjacent sibling combinator `+`.

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+
+~~~css
+/* This rule will only select the div with the classes child group2 */
```
mgrigoriev8109, line 64: Might sound better and be more descriptive (selecting class, not classes) if worded as. 

/* This rule will only select the div with the class of child group2 */

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+.group1 + div {
+  /* Our cool CSS */
+}
+
+/* This rule will only select the div with the classes child group3 */
```
mgrigoriev8109, line 69: Might sound better and be more descriptive (selecting class, not classes) if worded as. 

/* This rule will only select the div with the class of child group3 */

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+</style>
+
+<body>
+  <div> Let's <span class="emojify">emojify</span>this span!</div>
+</body>
+~~~
+
+Using these pseudo-elements this way would give us this result:
+
+Let's 😎 🥸 🤓 emojify 🤓 🥸 😎 this span!
```
mgrigoriev8109, line 170: Not quite sure what's going on here but I can look into it, the image between the two smiley faces isn't loading for me and it states disguised_face instead.

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+
+[`::selection`](https://css-tricks.com/almanac/selectors/s/selection/) allows you to change the highlighting when a user selects text on the page.
+
+ There are lots more! Have a quick browse through the [pseudo-element docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) to see a fuller list of what's possible.
```
mgrigoriev8109, line 174: In my opinion "fuller" is one of those words in the English language which sounds strange phonetically, so could be replaced with:
"...to see a more substantial list of what's possible."

mgrigoriev8109 (changes requested):
4 sentence changes that have more to do with the flow of the sentence.
1 change for images that for whatever reason did not load for me, though it could be because I'm not using github correctly.

.mikhail

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Pseudo-classes
+Pseudo-classes offer us different ways to target elements in our HTML. There are quite a lot of them, and they come in a couple of different flavors. Some are based on their position or structure within the HTML. Others are based on the state of a particular element, or how the user is currently interacting with it. There are too many to cover in detail here but we'll have a look at some of the most useful ones. Pseudo-classes share the same specificity as regular classes (0, 0, 1, 0). Just like regular classes, most can be chained together
+
+As always don't forget to check the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) to see a complete picture of what's available.
+
+## Dynamic and User Action Pseudo-classes
```
codyloyd, line 94: I think we should specifically add :focus in this section.. it's a big one IMO.

html_css/v2/main-course/selectors.md
```diff
@@ -0,0 +1,264 @@
...
+  */
+}
+~~~
+
+To see what other things you can achieve with attribute selectors, such as searching case insensitively, or sub-strings separated by hyphens have a browse through the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors).
+
+
+### Assignment REVIEW
+<div class="lesson-content__panel" markdown="1">
+1. Complete [CSS Diner](https://flukeout.github.io/). You should be familiar with most of the content in the first couple of exercises, but practice and review never hurt! Don't forget to read the examples and explanations on the right.
```
codyloyd, line 241: lol this was too fun... nice find.

html_css/v2/main-course/selectors.md
codyloyd: In addition to this comment, I feel a little iffy about the indentation statement.. it's not _technically_ true, because you don't _have_ to use proper indentation in HTML.

I'd feel better about it if we had a little disclaimer.. something like:

...will select an element that is one level of indentation down (assuming you're indenting your code properly).

or

Phrased another way, if you have your code indented properly, the child selector will select an element that is one level of indentation down.

codyloyd:
fantastic work here!

html_css/v2/main-course/selectors.md
shiroinegai: I think that it is probably confusing for the two sentences to be together as a learner would be reading the first sentence which relates to the code above, while the second sentence refers to the code below.

I believe it'd be better to have the first sentence be put at line 47 or completely omitted to encourage learning the actual terminology.

html_css/v2/main-course/selectors.md
ChargrilledChook: changed to 'complete' which is I think more concise and more accurate

PR #22703, opened 2021-09-24
fix/external-link-API-FullStackJS

Ishmam156:
 - [X] The PR title summarizes the change and where it happened, for example: "Fixes punctuation in Clean Code lesson".
 
Current external link is landing learners on the SDK page of GIPHY docs and might create some difficulty to navigate to the API section and thus changed the link to directly take a learner to the API section.

#XXXXX


wise-king-sullyman (approved):
It looks like that link has changed over time and this should bring it back in line with the original intent. Good catch and thanks for your contribution!

PR #22725, opened 2021-09-28
Adds Custom Properties lesson

thatblindgeye:
 
closes https://github.com/TheOdinProject/top-meta/issues/28 with 1 new lesson

#XXXXX


html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+
+Besides just adding custom properties to the CSS file, we need to add in some JavaScript to handle changing the theme, as well as some interactive element for the user to actually change it (such as a toggle or a switch). At this point, we trust that you can easily find out how to do both of these since that's a little out of scope for this lesson.
+
+#### Media Queries
+
+Giving users the ability to toggle a theme themselves is great, but there's another option for setting a theme that you may have noticed on certain sites or applications: using the user's theme setting from their operating system or user agent (like a browser).
+
+Instead of adding custom properties to the `:root` element depending on which class it has, we would add our custom properties on the `:root` element, without any class, inside of a `prefers-color-scheme` media query. Let's tweak our previous example a little bit to use this media query:
+
+~~~css
```
codyloyd, line 163: i think best practice here is to first define a default theme outside the media-query and then put the other theme inside a media-query.

this will avoid any issues that might arise if `prefers-color-scheme` is unset, has a different value, or if the browser does not support it.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
+### Introduction
+
+Custom properties (also known as CSS variables) can be a really useful and powerful tool when writing our CSS files. In short, they allows us to reference a CSS value however many times we want throughout a file. How is this any different or better than just using the value, you ask?
```
codyloyd, line 3: I'd leave out the rhetorical question here ("how is this different..") and just be direct 😄

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+
+By using custom properties, instead of having to update every single instance of a specific value ("This shade of red is too light, let's change the shade on all seven of these selectors"), we only need to update a single instance: the custom property itself.
```
codyloyd, line 5: also helps keeping colors consistent throughout a project, which becomes very useful in larger projects.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+
+Not only that, but we can even use the same custom property in different selectors, giving each instance of the property a different value. This is incredibly useful for creating themes, such as the dark and light themes you see on many websites these days.
```
codyloyd, line 7: the wording is confusing here.

I think it might be clearer to say something like:

> It is even possible to redefine custom properties in different contexts, which is useful for creating themes.... etc.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+  --color-text: rgb(18, 18, 18);
+}
+
+.themed-element {
+  background-color: var(--color-background);
+  color: var(--color-text);
+}
+~~~
+
+We just created some quick themes! So what is actually going on here? First, we added a `class` attribute on our `html` element in the HTML file so that we can have a more specific selector to add custom properties onto, and so that our page has a default theme.
```
codyloyd, line 151: it would be pretty wicked to throw in a codepen embed with a button to toggle the themes.

codyloyd (changes requested):

html_css/v2/main-course/custom-properties.md
thatblindgeye: Was gonna ask about the permissions for TOP's Codepen, since I think at least one Codepen embed would work really well on this lesson.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+.cool-div {
+  --main-bg: red;
+}
+
+.cool-paragraph {
+  background-color: var(--main-bg);
+}
+~~~
+
+Even though we declared our custom property on our `.cool-div` selector, we can still access it on its descendant `.cool-paragraph` selector. Thanks, dad! Er, thanks `cool-div`!
```
I3uckwheat, line 93: I think this joke may miss a bit too hard for our ESL folks.
> Thanks, dad! Er, thanks `cool-div`!

html_css/v2/main-course/custom-properties.md
I3uckwheat: We should avoid "not only that" as well, if it's referencing a previous paragraph, it likely should be part of that paragraph. Otherwise, we should be very specific in what "that" is.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+
+.boring-paragraph {
+  background-color: var(--main-bg);
+}
+~~~
+
+In the above example, only the element with the `cool-paragraph` class gets styled with a red background, while the `boring-paragraph` selector doesn't since it's not within the scope of where the custom property was declared. If you're familiar with how scope works in JavaScript, this sort of behavior should feel a little similar.
+
+That doesn't mean a custom property can *only* be accessed on the selector it was declared for, though. The scope includes the selector the custom property was declared for as well as the selector's descendants. Let's take a look at a modified version of the above example:
+
```
I3uckwheat, line 77: I think it could help to explain the example before demonstrating it. This way we can reduce people being confused when trying to follow along expecting this to work, just to tell them in the next paragraph it won't work.

It's alright to have supporting text under the example, but it may not be a good idea to put an example that a person may expect to work before telling them it's not going to work.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+.cool-paragraph {
+  background-color: var(--main-bg);
+}
+~~~
+
+Even though we declared our custom property on our `.cool-div` selector, we can still access it on its descendant `.cool-paragraph` selector. Thanks, dad! Er, thanks `cool-div`!
+
+#### The `:root:` Selector
+
+While there may be times where you will want to limit the scope of a custom property, you may want to be able to use other custom properties on many, unrelated selectors. One work around would be declaring the same custom property on a bunch of selectors, but that defeats one of the purposes of using custom properties in the first place (the ease of changing multiple instances of a value at once).
```
I3uckwheat, line 97: `work around` -> workaround

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+  background-color: var(--color-background);
+  color: var(--color-text);
+}
+~~~
+
+We just created some quick themes! So what is actually going on here? First, we added a `class` attribute on our `html` element in the HTML file so that we can have a more specific selector to add custom properties onto, and so that our page has a default theme.
+
+In our CSS file we're just creating two scopes for our custom properties on the `:root` selector, one for when our `html` (or root) element has a class of `dark`, and another for when it has a class of `light`. Then our `.themed-element` selector will use the values of the custom properties depending on which class is currently present on our root element.
+
+Besides just adding custom properties to the CSS file, we need to add in some JavaScript to handle changing the theme, as well as some interactive element for the user to actually change it (such as a toggle or a switch). At this point, we trust that you can easily find out how to do both of these since that's a little out of scope for this lesson.
```
I3uckwheat, line 155: > At this point, we trust that you can easily find out how to do both of these since that's a little out of scope for this lesson.

I don't think we need this part, the reader should understand that.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+  background-color: var(--main-color);
+}
+~~~
+
+By declaring our custom property on the `:root` selector in the example above, we can access it on *any* other valid selector within our CSS file, since any other selector would be considered a descendant of the `:root` selector.
+
+### Themes
+
+Beyond allowing us to access custom properties more globally, the `:root` selector gives us one way to add themes to our pages. Keep in mind that we're not covering every way to add theming in this lesson, as we just want to introduce you to a simple way that you can get a lot of use out of.
+
```
I3uckwheat, line 125: I would consider renaming, and reframing this section from "Themes" to "Case Study" or something. This way we don't need to explain to the reader that there are more ways to do themes.

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+### Additional Resources
+
+This section contains helpful links to other content. It isn’t required, so consider it supplemental for if you need to dive deeper into something.
+
+### Knowledge Check
+
+* How would you declare a custom property with a name of `text-color`?
+* How would you access a custom property with a name of `background-color`?
+* Where would you declare a custom property to have its scope be global and accessible by all other selectors?
+* Where would you declare a custom property so that a user's theme setting from their OS or browser was taken into account?
```
I3uckwheat, line 205: Please add a newline here :)

I3uckwheat (changes requested):
There are a lot of rhetorical devices in this lesson, but overall has great content. With a few tweaks this is for sure going to be a great addition!

Edit: By rhetorical devices, I mean: rhetorical questions,and presenting broken examples before explaining it.

html_css/v2/main-course/custom-properties.md
codyloyd: 🤔 I like "themes" here actually... it's one of the big wins for custom properties.

if we need to be more specific lets do "creating themes with custom properties"

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,205 @@
...
+### Themes
+
+Beyond allowing us to access custom properties more globally, the `:root` selector gives us one way to add themes to our pages. Keep in mind that we're not covering every way to add theming in this lesson, as we just want to introduce you to a simple way that you can get a lot of use out of.
+
+~~~html
+<html class='dark'>
+<!-- our boilerplate... -->
+  <body>
+    <div class='themed-element'>Our themed text</div>
+  </body>
```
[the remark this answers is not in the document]
thatblindgeye: I only included it here since I'm including the actual `<html>` tags, whereas the other ones that tag isn't really needed. But the comment probably isn't necessary thinking about it, since at the point these lessons are people *really* shouldn't need to be reminded "make sure you include this stuff"

html_css/v2/main-course/custom-properties.md
I3uckwheat: Id happy with some more specificity here, I just don't want to imply that it's the ONLY thing you can do with this, or the ONLY way you can do themes. 

However, our readers aren't dumb, they can figure that out.

thatblindgeye: True. I can change the section header to "Creating Themes with Custom Properties" like Cody said (or something similar),and just remove the second sentence.

"Beyond allowing us to access custom properties more globally, the `:root` selector gives us one way to add themes to our pages." Saying "one way to..." could definitely be enough to clue in readers that this isn't the only way to create themes. The second sentence is just me being very explicit (admittedly at times when really it's not needed).

html_css/v2/main-course/custom-properties.md
thatblindgeye: So, one example I came across that fit your description here (others I came across just used two of these queries for both dark and light), mentioned that the `light` value is both used for when "light" is the preference *and* when there is no preference/an umatched preference. I tested this out on Ubuntu (putting the "dark" colors outside the query as the default and using the query with the "light" value) and it seems like this might be true, but the "Standard" theme might just be considered the light theme underneath it all.

I want to test this out a little more before comitting to changing this piece, but I do have it updated locally based on your suggestion for now.

html_css/v2/main-course/custom-properties.md
codyloyd: >  Saying "one way to..." could definitely be enough to clue
 100% this

html_css/v2/main-course/custom-properties.md
```diff
@@ -0,0 +1,134 @@
+### Introduction
+
+Custom properties (also known as CSS variables) can be a really useful and powerful tool when writing our CSS files. In short, they allows us to reference a CSS value however many times we want throughout a file. By using custom properties, instead of having to update every single instance of a specific value ("This shade of red is too light, let's change the shade on all seven of these selectors"), we only need to update a single instance: the custom property itself. Not only that, but custom properties can help us keep colors consistent throughout a project, something that will be really helpful as projects get larger.
```
codyloyd, line 3: `they allows` > `they allow`

codyloyd:
I just remembered something that i omitted from the original outline:

FALLBACK VALUES.

the syntax is: `var(--some-variable, 10px)`

you can even nest other variables inside them: `var(--some-var, var(--some-other-var, 10px)`

the fallback value is used when the given variable is unset.

PR #22728, opened 2021-09-29
Update objects-constructors.md

JuanDCeballos:
the 'additional resources' appears twice. At his original position and then at the bottom of the 'Knowledge Check', I already delete it.

the 'additional resources' appears twice. At his original position and then at the bottom of the 'Knowledge Check', they are necessary because we don't want repetition of information.


twalton83 (approved):

twalton83:
Thank you!

PR #22755, opened 2021-10-01
add section about the viewport meta tag

codyloyd:
just adds a short bit that got left out of this lesson originally.


twalton83 (approved):
LGTM

PR #22757, opened 2021-10-01
Lesson: HTML/CSS - Positioning #30

LeoKirasic:
 
I have written the first draft of the positioning lesson.

TheOdinProject/top-meta/issues/30

### DIFF
--- a/html_css/v2/main-course/positioning.md
+++ b/html_css/v2/main-course/positioning.md
@@ -15,8 +15,9 @@ The difference between static and relative is fairly simple. Static is the defau
 Absolute positioning allows you to _absolutely_ position elements on the screen. Absolutely positioned elements will be relative to the next absolute or relative parent. Absolutely positioned elements are also removed from the normal flow of the document. To put it in simple terms: elements that are removed from the normal flow of the document don't affect other elements and are also not affected by other elements. Using absolute positioning allows you to position elements anywhere on the screen using top, right bottom and left properties. This property is really useful when you want to position something at an exact point on the screen, without disturbing any of the other elements.
 
 ### Fixed Positioning 
-### Sticky Positioning
+Fixed elements are also removed from the normal flow of the document and are positioned relative to the _viewport_. You basically use top, right bottom and left properties to position it, and it will stay there as the user scrolls. This is especially useful for things like navigation bars.
 
+### Sticky Positioning
 ### Assignment
 
 ### Additional Resources
--- a/html_css/v2/main-course/positioning.md
+++ b/html_css/v2/main-course/positioning.md
@@ -9,13 +9,13 @@ You have learned various ways how to position elements on the screen and even th
  - You'll know the difference between each property and how to combine them.
 
 ### Static and Relative Positioning
-The difference between static and relative is fairly simple. Static is the default position of every element, and properties top, right, bottom, and left have no affect on statically positioned elements. Relative on the other hand is pretty much the same as static, but properties top, right bottom, and left affect the specific element relative to itself.
+The difference between static and relative is fairly simple. Static is the default position of every element, and properties _top_, _right_, _bottom_, and _left_ have no affect on statically positioned elements. Relative on the other hand is pretty much the same as static, but properties  _top_, _right_, _bottom_, and _left_ affect the specific element relative to itself.
 
 ### Absolute Positioning
-Absolute positioning allows you to _absolutely_ position elements on the screen. Absolutely positioned elements will be relative to the next absolute or relative parent. Absolutely positioned elements are also removed from the normal flow of the document. To put it in simple terms: elements that are removed from the normal flow of the document don't affect other elements and are also not affected by other elements. Using absolute positioning allows you to position elements anywhere on the screen using top, right, bottom, and left properties. This property is really useful when you want to position something at an exact point on the screen, without disturbing any of the other elements.
+Absolute positioning allows you to _absolutely_ position elements on the screen. Absolutely positioned elements will be relative to the next absolute or relative parent. Absolutely positioned elements are also removed from the normal flow of the document. To put it in simple terms: elements that are removed from the normal flow of the document don't affect other elements and are also not affected by other elements. Using absolute positioning allows you to position elements anywhere on the screen using  _top_, _right_, _bottom_, and _left_ properties. This property is really useful when you want to position something at an exact point on the screen, without disturbing any of the other elements.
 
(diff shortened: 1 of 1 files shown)


html_css/v2/main-course/positioning.md
```diff
@@ -0,0 +1,38 @@
+### Positioning
+
+You have learned various ways how to position elements on the screen and even though the methods you know so far are sufficient in most cases, these CSS rules are still very powerful and will certainly help you step up your CSS game!
```
codyloyd, line 3: I think people are going to be confused about what we actually mean by "position" here because it _sounds_ like something they already know how to do (move stuff around a page with margin/padding/flexbox etc)... I think we could clear this up with how it's described in the introduction.... maybe something like this:

By now you have had quite a bit of practice moving elements around the screen using things like margin, padding, flexbox and grid. These techniques have all relied on CSS's default "positioning-mode".  This default positioning-mode is intuitive, and you'll continue using it for almost all of your layout needs, _however_, there are other methods at your disposal that can be very useful in some situations.

I3uckwheat (changes requested):
Because CSS and HTML are very visual, I think we really need to add some concrete examples and even use-cases where this could be used in real life. We really want to avoid making this seem like documentation, we can just send people to MDN for that.

A good example for `absolute` would be those chat boxes that a lot of websites have, or even a sticky header. Modals are also a good use-case for absolute positioning.

It could be very helpful to talk about the differences, in context as well. It's not clear when you should use a specific positioning property as it stands in the lesson. 

For examples, it can be very useful to introduce the example, then analyze it after displaying it. This is a repeatable way to solidify things into people's heads as it helps them to see it, make hypothesis, do experiments, and verify hypothesis. Then by doing an analysis, after they play around with it, we can help bring more awareness of the topics.

One thing I used to hear in school a lot was, "When will we ever need to use this in real life?", I think that's an important thing to answer, by giving concrete examples and such, we can achieve that answer.

Of course, defer to Cody's suggestions first.

html_css/v2/main-course/positioning.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Learning Outcomes
+ - You'll learn how to use absolute positioning.
+ - You'll learn how to use fixed positioning.
+ - You'll learn how to use sticky positioning.
+ - You'll know the difference between each property and how to combine them.
+
+### Static and Relative Positioning
+The difference between static and relative is fairly simple. Static is the default position of every element, and properties _top_, _right_, _bottom_, and _left_ have no affect on statically positioned elements. Relative on the other hand is pretty much the same as static, but properties  _top_, _right_, _bottom_, and _left_ affect the specific element relative to itself.
```
codyloyd, line 12: If you change the intro above, I'd edit the wording here to reflect the fact that we've already mentioned that there is a default positioning method... so you can start with something like:

> The default positioning mode that you've gotten used to is `position: static`.....

and then continue from there.

also: I don't think  you need to mention `top, right, bottom, left` twice here.... you can simplify by just saying something like:

> Relative, on the other hand, is pretty much the same as static except that you can move the element around using the properties top, ri....(etc.)

unrelated note: feel free to put all of these suggestions into your own words.  Or quote them.  I don't care either way 💜

html_css/v2/main-course/positioning.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Absolute Positioning
+Absolute positioning allows you to _absolutely_ position elements on the screen. Absolutely positioned elements will be relative to the next absolute or relative parent. Absolutely positioned elements are also removed from the normal flow of the document. To put it in simple terms: elements that are removed from the normal flow of the document don't affect other elements and are also not affected by other elements. Using absolute positioning allows you to position elements anywhere on the screen using  _top_, _right_, _bottom_, and _left_ properties. This property is really useful when you want to position something at an exact point on the screen, without disturbing any of the other elements.
```
codyloyd, line 15: I think i'd reword this one to lead with the simple explanation and then give more details later.... Like I'd lead with

> `position: absolute` allows you to position something at an exact point on the screen without disturbing the other elements around it. More specifically, using absolute positioning on an element will remove that element from the normal document flow ....... etc.

I think some practical usage ideas would be welcome here. Off the top of my head, absolute positioning is primarily useful for elements that stack on top of each other:
  - an image with a caption on it
  - icons on top of other elements
  - visual design tricks like a logo that overhangs the header

  I'd also add a little disclaimer about not relying on absolute positioning to do entire layouts... it's not an everyday kind of tool and if something can be accomplished using flex/grid etc. do that!

html_css/v2/main-course/positioning.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Fixed Positioning
```
codyloyd, line 17: nothing wrong with this section.. another common example though: those little floating chat-buttons (we have one on TOP, might be worth pointing that out.)

html_css/v2/main-course/positioning.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Fixed Positioning 
+Fixed elements are also removed from the normal flow of the document and are positioned relative to the _viewport_. You basically use  _top_, _right_, _bottom_, and _left_ properties to position it, and it will stay there as the user scrolls. This is especially useful for things like navigation bars.
+
+### Sticky Positioning
+Sticky elements will act like normal elements until you scroll past them, then they start behaving like fixed elements. They are also not taken out of the normal flow of the document. It might sound confusing, but checking out the behavior of <a href="https://codepen.io/theanam/pen/MPLBYy">this</a> example might clear things up for you.
```
codyloyd, line 21: useful for things like section-headings on lists or tables that can scroll off the screen.  for example it's nice to be able to see what 'category' you're looking at on a long scrolling list, even tho the header should have scrolled off the screen.

html_css/v2/main-course/positioning.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Assignment
+<div class="lesson-content__panel" markdown="1">
+1. <a href="https://www.youtube.com/watch?v=jx5jmI0UlXU">This</a> video is fast-paced but provides a good visual representation of different positioning behaviors. Go ahead and watch it.
```
codyloyd, line 25: this video is perfect... great find!

codyloyd:
nice work overall :D  left a few comments.

codyloyd (changes requested):

PR #22810, opened 2021-10-04
Made sentence logical on line 123

NueMe:
Fixed line 123 to say "less than or equal to" instead of "greater than" the answer which makes more sense.

 - [ x The PR title summarizes the change and where it happened, for example: "Fixes punctuation in Clean Code lesson".
 
The sentence should state that the comparison operator used is "less than or equal to" and not "greater than"
fixed sentence to make it logical

#XXXXX

### DIFF
--- a/foundations/javascript_basics/problem_solving.md
+++ b/foundations/javascript_basics/problem_solving.md
@@ -117,7 +117,7 @@ for (let i = 1; i <= answer; i++) {
 }
 ~~~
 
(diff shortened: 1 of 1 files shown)


twalton83 (commented):
This actually isn't accurate. The second clause of a `for` loop indicates that we should loop while this condition is true. In this case, we loop and increment `i` while `i` is less than or equal to `answer`, and the loop ends when `i` is greater than `answer`. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Loops_and_iteration#for_statement
> The conditionExpression expression is evaluated. If the value of conditionExpression is true, the loop statements execute. If the value of condition is false, the for loop terminates. (If the condition expression is omitted entirely, the condition is assumed to be true.)

NueMe:
Thanks for the explanation!

PR #22830, opened 2021-10-08
Add Node.js to popular server-side languages

BLooDBRothER:
 
Node.js is becoming popular nowadays so I thought of adding it to the popular server-side languages. Since it's mentioned as "Some of the most popular......" in the lesson I thought it would be good to have it in that para.


twalton83:
I don’t think the change is necessary especially given we mention Node in the context of backend. Thanks though!

PR #22838, opened 2021-10-09
Fixes npm init command in Project: Restaurant Page

thehashmap:
 
Updated 'npm init' command to 'npm init -y' (which is the same as used in the webpack tutorial), to instantly initialize the project because 'npm init' command prompts the user to input a few parameters which can be overwhelming for beginners.

#XXXXX


thatblindgeye:
Agreed with the above. I think a better change would be to explicitly tell users that they can leave the default values or something, but those parameters were already covered in a resource in the ES6 Modules lesson (the npm section). And at that point, users should know that they can look into what a specific parameter is for/what the value should be.

PR #22967, opened 2021-10-25
Add `lang="en" to the <html> tags

SiameseFeline:
Add a lang attribute explanation to the html tag so the boilerplate code passes the W3C validator, as well as a link to the different language codes.

 - [done] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md).
 - [done] The PR title summarizes the change and where it happened, for example: "Fixes punctuation in Clean Code lesson".
 
Adds the lang attribute to the html tag that is needed to pass W3C validation in the HTML boilerplate foundation lesson, and an explanation of it as well as a link to language codes for the lang tag.

### DIFF
--- a/foundations/html_css/html-foundations/html-boilerplate.md
+++ b/foundations/html_css/html-foundations/html-boilerplate.md
@@ -47,6 +47,20 @@ Back in the `index.html` lets add the `<html>` element by typing out its opening
 </html>
 ~~~
 
+### Lang Attribute
+
+We need to add a lang attribute to the `<html>` tag. So let's do it now.
+
+Back in the `index.html` let's add `lang="en"`(language is English) to the `<html>` tag.
+
+You can use other language codes for other languages [here](https://www.w3schools.com/tags/ref_language_codes.asp), too.
+
+~~~html
+<!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


thatblindgeye:
I will say that this is both a validation *and* accessibility issue, at least according to WCAG. That said, I don't believe it is mentioned in the accessibility lessons.

That said, I agree with Marvin. This should not get its own section. The original mention of the `html` tag on line 46 should be updated to just have this attribute from the start, and from there your lines 50-63 should be replaced with a Heading 4 `####` section **briefly** explaining why the attribute is needed, including a quick mention that it helps assistive technologies announce the correct pronunciation of words.

Please let us know if you would like to continue making these changes, @SiameseFeline .

twalton83:
Would we like to follow through on these changes? Otherwise, I will close because this is stale.

thatblindgeye:
@SiameseFeline I am closing this PR due to inactivity and a lack of reply. I have opened an official issue at #23256. If you wish to claim it and make the updates as requested please comment on the issue itself, otherwise the issue is open for anyone to claim.

PR #23043, opened 2021-11-09
Improves explanation of Array.new's block form in Ruby Nested Colelctions lesson

themetar:
Previous explanation conflated mutability with multiple references to one object.
 
Previous explanation was wrong and misleading. It kinda implied that mutability in and of itself was the reason to use a block to set initial values for an array.


ruby_programming/basic_ruby/nested_collections.md
```diff
@@ -79,7 +79,7 @@ teacher_mailboxes.dig(0, 4)
...
 
-To create an array of immutable objects, you will need to pass the default value to `Array.new` in a block, using curly braces, instead of the second optional argument. This creates *distinct copies* of the default value in the array, rather than references to the same default value.
+To create an array of *distinct objects*, you will need to pass the default value to `Array.new` in a block, using curly braces, instead of the second optional argument. The code in the block gets evaluated for every slot in the array, instead of just once when executing the arguments to `Array.new(size, some_default_value)`. This creates *distinct values* to populate the array with, rather than references to the same default value.
```
rlmoser99, line 82: What do you think of this wording to clarify the confusion?

```suggestion
To create an array of mutable objects (string, array, hash, etc), you will need to pass the default value to `Array.new` in a block, using curly braces, instead of the second optional argument. The code in the block gets evaluated for every slot in the array, creating multiple objects of the default value, rather than references to the same object.
```

rlmoser99 (changes requested):
Thank you for taking the time to make this PR. In reading your suggested changes, I see where the confusion can stem from. However, I do not think this completely captures what is meant to be said here, so let me know what you think of my proposed change.

ruby_programming/basic_ruby/nested_collections.md
themetar: First off, someone changed your original "mutable" to immutable at the beginning of line 82  - I guess because they wrongfully thought of mutability as the change to all elements of the array in the first case - and their change was accepted; I don't understand why.

Anyway, the reason I shy away from "default value" is because in the block form there is no default value. The array gets populated with the returned values from the block. You could write this:

```
arr = Array.new(5) do |index|
  if index % 2 == 0
    index
  else
    Array.new
  end
end
```
Or any custom complex code. In the above example the array will be `[0, [], 2, [], 4]`. What's *the* default value? They aren't copies of something nor objects of a single class.

rlmoser99: Let's not worry about what has changed in the past and focus on the present - I can see now that this wording has been confusing and want to use this opportunity to clarify this confusion. 

The reason that I changed to "mutable object" is to clarify that it is indicating the type of objects inside an array. An array of strings, an array or arrays, an array of hashes. This goes along with the wording in the paragraph above.

> According to the documentation the second optional argument, for the default value, should only be used with an immutable (unable to be changed) object such as a number, boolean value, or symbol. Using a string, array, hash, or other mutable object may result in confusing behavior...

I do not think think we need to worry about your example of a complex block that could be passed for this lesson. However, if you want to further clarify my suggested change you are welcome to edit it, but I will not approve your change as is.

themetar: I made some changes to make it more to the point (e.g. "To create an array of mutable objects (string, array, hash, etc)"). If you could take a look.

rlmoser99: > initialize the array with

Great wording! Thank you so much for helping clarify the confusing language in this paragraph!

rlmoser99 (approved):

PR #23051, opened 2021-11-11
Update TOP style guide and template files

thatblindgeye:
 
Updates have been made to our style guide and template files in order to bring them up to date with current practices, with some suggested changes. This will make it a lot easier for editing old lessons and adding new ones by having a source of truth, and users can just make a copy of a template file to begin writing new lessons.

In addition to somewhat minor changes, some of the more notable changes include:

1) Added short sections regarding adding links and images to lessons/projects, including a link to our Wiki page on adding images to the curriculum. Verbiage is taken from the Google style guide, but can be rewritten to be more “in house” if needed.

2) Removed the text explaining what each section was for from the layout code block and added it to a numbered list immediately following the code block, more similarly to the Google style guide. This makes it easier for users to just copy+paste and makes it clearer what is meant to be part of the template  vs explanations that shouldn’t be part of the lessons.

3) Replaced the previous verbiage plus the verbiage currently used on some lessons ("By the end of this lesson, you should be able to do the following:") for Learning Outcomes. I think it may help to kind of of take the LO's from "do this thing" to "you'll learn about this thing". I'm also fine with removing any verbiage completely so that it's just the heading and the bulleted list, but wanted to get thoughts on this.

4) Tweaked the purpose of the Assignment section. Instead of just stating that the section includes resources, I opted to include both resources and exercises/practice/other tasks. I feel assignments should be viewed as “this is what you need to do in order to fully complete this lesson,” and exercises/practice should be included in that rather than given their own section *after* the assignment if it's actually required.

5) Included a default bullet item for when the Additional Resources section doesn’t have any actual resources, and telling users that they can expand the section by contributing. This should help prevent people asking whether the section should be removed when a lesson has no resources, and may help people want to get more involved.

    I also moved this section back to the bottom of the template, after Knowledge Checks (instead of before). My rationale for this is because additional resources may include content that is related, but not entirely necessary for the user to read, and thus most likely won’t have the answers for or be linked to by the knowledge checks. I feel it’s better to have users answer the knowledge checks before inviting them to go down a rabbithole of additional resources that aren't pertinent to the actual lesson completion.

6) The project template basically follows the same as above where applicable, but I also added in optional pre and post assignment sections to the template to cover situations where sections are needed prior to a project assignment (e.g. in the Landing Page project where users are shown how to setup a GitHub repo ) or afterwards (e.g. in the Todo List project where users are linked to a few coding exercise links).

    We may also want to decide whether project assignments should follow the Recipes project in Foundations (with several “iterations” and each iteration having its own numbered list) or keep the layout as is (a single numbered list). The Iteration approach looks a bit nicer as it spreads things out a little more, preventing that single, giant block of text some projects can run into, but I'm also fine with just having project assignments cleaned up a bit as is.

#XXXXX


templates/TOPStyleGuide.md
```diff
@@ -24,131 +25,146 @@ TOP uses Markdown for its lessons to be easily read and edited, and get properly
...
+1. A resource/task.
 </div>
 
-### Exercises
+### Knowledge Check
+
+This section contains questions for you to check your understanding of this lesson on your own. If you’re having trouble answering a question, click it and review the material it links to.
 
-A group of exercises (If Applicable) for the student to complete in relation to the topic taught in the lesson. 
+*   <a class="knowledge-check-link" href="A knowledge check URL">A knowledge check question.</a>
```
fortypercenttitanium, line 64: Could the href example mirror a markdown header link and maybe use all caps to imply it needs to be replaced? Just a thought. `href="#REPLACE-ME"` or `href="#KNOWLEDGE-CHECK-LINK"` .This would make it easier when copying and pasting the template when writing a lesson because it "jumps out" when it's not replaced.

templates/TOPStyleGuide.md
```diff
@@ -24,131 +25,146 @@ TOP uses Markdown for its lessons to be easily read and edited, and get properly
...
 
-  *   Learning outcome 2 answer
+5. `### Knowledge Checks`: A bulleted list of specific questions that a user should be able to answer on their own after reading the lesson and completing the assignment. Each knowledge check should link either to a Heading 3 `###` section within the lesson or a resource previously linked to in the lesson. This link should help users review the necessary material in order to answer the knowledge check. 
 
-</details>
-~~~
-## Project Layout
+    Replace the `A knowledge check URL` text with the actual link to the section/resource and the `A knowledge check question.` text with your own question/problem that the user should be able to solve. Then add any additional bulleted knowledge check items. The lesson should ideally have no more than 7 knowledge checks, but this number might vary by lesson **If the lesson does not have any knowledge checks, remove this entire section from the lesson.**
```
fortypercenttitanium, line 86: If you make the change suggested above, it should be reflected here as well.

fortypercenttitanium (changes requested):
Awesome work man. I'd like to see one more section about CodePen embeds. Basically, we should use the recommended HTML embed codepen provides when you click "embed". Some lesson-writers have been confused about this in the past because there are issues rendering them in the lesson preview tool.

templates/TOPStyleGuide.md
thatblindgeye: Not using either type of link as the default text for the `href` attribute was purposeful, but I do see your point. Unlike the rest of the default text used in the template, the KC instance does have that default text kind of crammed in with other text which I could see being easy to gloss over.

Won't make any change jus tyet in case anyone else has any thoughts, but I'm gonna keep this in mind.

thatblindgeye: I ended up going for more of a middle-ground approach. I wanted to refrain from defaulting to any type of link for the KC's, but I do agree it was difficult to pick out what text should be edited within that chunk of text.

I opted to update both layout files so that any text that *should* be replaced by the author is in all caps throughout.

templates/TOPStyleGuide.md
```diff
@@ -1,13 +1,18 @@
+<!-- Learning Outcomes: Keep old verbiage and bullets; discuss with Bender and Sully -->
+<!-- Project Iterations/Steps: Keep; contact James -->
+<!-- Assignment vs Practice: remove links scattered throughout the lesson; keep Practice and Assignment separated -->
```
wise-king-sullyman, line 3: I assume these aren't intended to be kept?

wise-king-sullyman (changes requested):
This looks absolutely 🔥 🔥 🔥. I see one thing that looks like it needs to be removed then I'll be ready to approve.

templates/lesson-template.md
```diff
@@ -0,0 +1,41 @@
+### Introduction
+
+A BRIEF INTRODUCTION.
+
+### Learning Outcomes
```
01zulfi, line 5: In the lesson layout in `TOPStyleGuide.md`, this heading is called "Learning Overview". While here and in the `lesson-example.md`, this heading is called "Learning Outcomes". 

Is this intentional? Which heading should be used in the lesson?

thatblindgeye: Good catch, thank you! Lesson Overview is the new verbiage we're going for.

wise-king-sullyman (approved):

thatblindgeye:
I updated the name of the main style guide file and moved it to live with other MD files at the root. I think style guides like this should be more easily findable rather than tucked away in a "templates" folder.

wise-king-sullyman (approved):

PR #23054, opened 2021-11-13
Create asset_pipline.md

Reed-SolomonE:
### The Asset Pipeline

Assets in your application are additional files that get called by the browser after your initial job of HTML is received. Assets is used to mean things like CSS stylesheets, Javascript files, images, videos etc... basically anything that requires an additional request to grab it. It used to be that the Asset Pipeline handled all assets but since the inclusion of Webpack and the webpacker gem to give Webpack some Rails type conventions, it is now better practice to handle Javascript assets through there. Webpacker now comes as standard in Rails 6 applications. If you ever work on legacy applications that haven't included the webpacker gem you may well find Javascript is still handled by the asset pipeline.

**Changes**
Contextual change in verbose to gob changed to "job"

 - [ ] The PR title summarizes the change and where it happened, for example: "Fixes punctuation in Clean Code lesson".
 
...your text here

#XXXXX


Reed-SolomonE:
1. Ruby on Rails> Rails Basics> Lesson 5> The Asset Pipeline
2. "Fixes punctuation in The Asset Pipeline lesson".

KevinMulhern (changes requested):
I think gob was used correctly here since it can mean a large amount of of something
https://www.merriam-webster.com/dictionary/gob

But I can see how it would confusing, but "job" doesn't fit so maybe just changing that sentence to this would be better:
> Assets in your application are additional files that get called by the browser after your initial HTML is received.

What's your thoughts?

twalton83:
Closing because it is stale, and "gob" is correct here.

PR #23070, opened 2021-11-16
fix typo in factory function

osamamammar:

I found the letter "s" here is unimportant and unnecessary, especially since it has a strange shape during reading.

#### 2. Related Issue

fix typo in factory function #23070

### DIFF
--- a/javascript/organizing-js/factory-functions.md
+++ b/javascript/organizing-js/factory-functions.md
@@ -160,7 +160,7 @@ counter(); // 2
 counter(); // 3
 ~~~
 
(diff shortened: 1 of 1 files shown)


DevlinRocha (changes requested):
Instead of removing the "S" I think it would be better to use an apostrophe to make it appear like this:

`console.log`'s

osamamammar:
It's okay. thank you. @DevlinRocha

fortypercenttitanium:
> Instead of removing the "S" I think it would be better to use an apostrophe to make it appear like this:
> 
> `console.log`'s

As I mentioned earlier, this is actually not technically correct grammar. Thanks for the suggestions everyone, but I'm closing this.

DevlinRocha:
> As I mentioned earlier, this is actually not technically correct grammar. Thanks for the suggestions everyone, but I'm closing this.

I'm sorry, I never saw another mention of this issue anywhere. I must have missed it.

PR #23090, opened 2021-11-19
Update setting_up_git to configure default editor

crespire:

Adds a small section on configuring VSCode as the default editor for git.

#### 2. Related Issue
none

### DIFF
--- a/foundations/installations/setting_up_git.md
+++ b/foundations/installations/setting_up_git.md
@@ -125,6 +125,11 @@ git config --get user.name
 git config --get user.email
 ~~~
 
+You can also set up a default editor that `git` will use for more advanced commands. We will set up `git` to use VSCode as its default editor by typing
+~~~bash
+git config --global core.editor "code -w"
+~~~
+
 #### Step 2.2: Create a GitHub Account or Sign In
 
 Go to [GitHub.com](https://github.com/) and create an account! If you already have an account, sign in. You do not need to use the same email address you used before, but it might be a good idea to use the same one to keep things simple.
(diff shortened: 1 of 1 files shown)


marvingay:
Why is this change necessary or important?

I question if we should enforce an editor for `git`. The "advanced commands" that are being mentioned are unknown / ambiguous at the time and the utility of imposing this upon the user is highly subjective.

Additionally, there's a minor argument to be made about how using not-VSCode (and consequently not enforcing this settings change) is beneficial long-term as it provides exposure and usage of an editor that you may have to use in the real world anyway (e.g. SSH'ing into some server and only having nano or vi at your disposal).

twalton83:
@marvingay i think we get a non insignificant amount of users stuck in vim and nano when they forget the -m flag in a commit who get confused and I think just using VSC here is beneficial.

crespire:
I think @marvingay makes a lot of good points, but I do feel that, given the _target_ audience of the course in general, setting up VSCode as the default editor would be a somewhat prudent choice.

At this point in the course, folks may not even be comfortable in command line, so this change may help to ease people into "doing the dev thing." As people grow in their depth of knowledge, folks will figure out they can change the setting to whatever their preference is.

That being said, I think Marvin's point is valid, and perhaps changing the language so that it's clear this setting is a user preference might be a good change. That way, it isn't so much as enforcing VSCode as a default editor, but rather setting up users for success in a new endeavour.

crespire:
I'm going to close this pull request, as I think there are changes in the git lessons that have made this obsolete, or at least misplaced.

Thanks for your input @marvingay and @twalton83!

PR #23097, opened 2021-11-20
Change COUNT(posts.*) to COUNT(*)  in 2nd and 3rd example of DATABASES AND SQL lesson

nguyenlekhtn:

I believe COUNT(posts.\*) is incorrect because we joined post and users table. The correct one should be COUNT(*). In the W3 example in the same lesson https://www.w3schools.com/sql/trysql.asp?filename=trysql_select_groupby,  if we do something similar like
```sql
SELECT  Country, COUNT(Orders.*)
FROM Customers JOIN Orders by Customers.CustomerID = Orders.CustomerID 
GROUP BY Country
```
and press Run, error popup will be shown

### DIFF
--- a/databases/databases.md
+++ b/databases/databases.md
@@ -129,7 +129,7 @@ You often see aliases (`AS`) used to rename columns or aggregate functions so yo
 Now we're getting into the fun stuff.  Aggregate functions like `COUNT` which return just a single value for your whole dataset are nice, but they become really useful when you want to use them on very specific chunks of your data and then group them together, e.g. displaying the `COUNT` of posts for EACH user (as opposed to the count of all posts by all users).  That would look like:
 
 ~~~sql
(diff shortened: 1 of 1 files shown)


CouchofTomato (commented):
Admittedly my SQL game isn't the strongest but I wonder if it would be better to count on the posts.user_id to be more explicit and make the example clearer. What do you think?

nguyenlekhtn:
> Admittedly my SQL game isn't the strongest but I wonder if it would be better to count on the posts.user_id to be more explicit and make the example clearer. What do you think?

I think you mean `posts.id` ? Because we are counting "posts_written"
I agree that `posts.id` is more clearer. I pushed a new commit to reflect this suggestion. What do you think?

CouchofTomato (approved):
Thanks @nguyenlekhtn

PR #23102, opened 2021-11-22
Edit Altering Elements section

linh15:

Section 'Altering Elements' (H4) is about the changes we can make to an element. The following sections are its subsections (Adding inline style, Editing Attributes, Working with classes, Adding text content and Adding HTML content), but they are presented in H4, which is equal to its parent. This causes confusion for learners, so I suggest changing these subsections to H5. 
I also elaborated more on the content of 'Altering Elements' section so it is clear that what the example is used for. Originally, the createElement example is identical to the one in 'Element Creation' without any further explanation.


fortypercenttitanium (approved):
Great contribution, thank you!

PR #23106, opened 2021-11-23
Replace calculator example in "factory functions and module pattern" lesson

EmilioAK:

This replaces the calculator module example with one from the resources linked on this page. The calculator example has no private properties, which leaves you wondering why you couldn't just use a simple object there instead. This example clearly shows the benefits of a module.


javascript/organizing-js/factory-functions.md
```diff
@@ -250,22 +250,21 @@ Modules are actually very similar to factory functions. The main difference is h
...
-  const mul = (a, b) => a * b;
-  const div = (a, b) => a / b;
-  return {
-    add,
-    sub,
-    mul,
-    div,
-  };
-})();
+var myModule = (function() {
```
twalton83, line 253: We probably don’t want to use var here, and we should probably just extended the calculator instead of using an abstract example.

twalton83 (changes requested):

javascript/organizing-js/factory-functions.md
EmilioAK: I can change the vars, but I don't know what you would have in mind for the calculator

twalton83:
So “let” shouldn’t be used here, const should be. That said I’ll close this but please think of ways to add a private method to this calculator module instead. We usually want to stick to more practical examples.

PR #23140, opened 2021-12-01
Adjusted content in additional resources paragraph within predicate-enumerable-methods lesson

Conor-Developer:

The wording in "Additional Resources" states there are more methods than covered in the lesson (with the example of #one). However, the lesson does in-fact cover **#one?** enumerable method. It's within the exercise and within the article you read as part of the assignment.

Here is the text:

"Additional Resources
This section contains helpful links to other content. It isn’t required, so consider it supplemental.

There are many more enumerable methods than are covered in this lesson (e.g., #member? and **#one?**). For a full listing, you can check out the Ruby Docs."

My solution is to remove this example.

#### 2. Related Issue

N/A

### DIFF
--- a/ruby_programming/basic_ruby/predicate_enumerable_methods.md
+++ b/ruby_programming/basic_ruby/predicate_enumerable_methods.md
@@ -221,7 +221,7 @@ fruits.none? { |fruit| fruit.length > 6 }
 ### Additional Resources
 This section contains helpful links to other content. It isn't required, so consider it supplemental.
 
-* There are many more enumerable methods than are covered in this lesson (e.g., `#member?` and `#one?`). For a full listing, you can check out the [Ruby Docs](https://ruby-doc.org/core-2.5.0/Enumerable.html).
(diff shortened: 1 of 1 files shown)


CouchofTomato:
Do you think it would be better to maybe just replace it with one other example? I would prefer to offer up more than one example to learners.

Conor-Developer:
> Do you think it would be better to maybe just replace it with one other example? I would prefer to offer up more than one example to learners.

Morning CouchofTomato,

I absolutely agree, in-fact I researched for an alternative (https://ruby-doc.org/core-3.0.2/Enumerable.html). The problem is, I couldn't find another example.

Predicate enumerable methods:

* all?
* any?
* include?
* none?
* one?
and finally... member?

All of the above are discussed in the lesson (except member?)

Do you know of any other predicate enumerable methods that are perhaps not included within the ruby documentation?

CouchofTomato:
haha that's fair enough. If they aren't in the docs they don't exist. Thanks.

PR #23149, opened 2021-12-05
Change Welcome to Confirmation in Rails Mailer Project Title

msespos:
 - [] If the PR is related to an open issue, use a [relevant keyword](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) and reference it with the `#` sign and the issue number.
- [] If changes are requested, please make the changes in a timely manner. After you submit the changes, request another review from the maintainer (top right).

Change Project: Sending Welcome Emails to Project: Sending Confirmation Emails to line up with the header and reflect the project better

#### 2. Related Issue

None


msespos:
Looking over the project again, I can see where "Welcome Emails" could be a description of the general type of emails used in the project, so I may be mistaken in requesting this change. My thinking was that "Confirmation Emails" is probably more precise a description, and also the header for the project does also say "Confirmation Emails" not "Welcome Emails."

msespos:
Looking at it one more time, I'm realizing that the term "thank you email" is used a few times in the project description. I guess my question is whether it would be a good idea or not to use the same term (confirmation, thank you or other) each time the emails are described, or whether this is not that important. If it is, my vote would be for "confirmation email" everywhere and I can make that change.

CouchofTomato:
I think since the title of the project is "Sending Confirmation Emails" we should ensure that is reflected throughout.

msespos:
The edits for consistency are done. I just have one other concern now.

When going through the project, I looked at the extra credit. In order to do the extra credit, you need to give Heroku your credit card. While this is not technically paying (they just need your info), it seems potentially problematic. If nothing else, I think a warning should be put in the project description. But I don't know if that should be removed entirely for that reason. Thoughts?

Otherwise this is done.

msespos:
This PR is now done.

zachmmeyer:
Looks great! Thanks @msespos !!

PR #23179, opened 2021-12-13
Update rails_backend.md

xari4808:
 - [ ] The title summarizes the change and where it happened, for example: "Fixes punctuation in Clean Code lesson".
 - [ ] If the PR is related to an open issue, use a [relevant keyword](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) and reference it with the `#` sign and the issue number.

- Replace the words "a different path" with a link to the full-stack JavaScript path
- Update wording of the Firebase assignments. The way it's currently worded sounds as if learners might've encountered Firebase in previous TOP lessons/projects
- Add [this video](https://www.youtube.com/watch?v=zQyrwxMPm88) as an additional resource for Firebase

#### 2. Related Issue

Closes #XXXXX

### DIFF
--- a/javascript/js-rails/rails_backend.md
+++ b/javascript/js-rails/rails_backend.md
@@ -59,8 +59,14 @@ By the end of this lesson, you should be able to:
 ### Additional Resources
 This section contains helpful links to other content. It isn't required, so consider it supplemental.
 
+#### With Ruby on Rails
+
 * [Grabbing your Rails form CSRF token with Javascript so Rails doesn't yell at you with "Warning, can't verify CSRF token authenticity", via SO](http://stackoverflow.com/questions/7203304/warning-cant-verify-csrf-token-authenticity-rails)
 * [... and another SO post on the CSRF token](http://stackoverflow.com/questions/8503447/rails-how-to-add-csrf-protection-to-forms-created-in-javascript)
+
+#### With Firebase
+
+* This [video](https://www.youtube.com/watch?v=zQyrwxMPm88) to reinforce the learnings from [Google Codelab](https://codelabs.developers.google.com/codelabs/firebase-web/#0).
 
 ### Knowledge Check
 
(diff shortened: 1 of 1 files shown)


javascript/js-rails/rails_backend.md
```diff
@@ -51,16 +51,23 @@ By the end of this lesson, you should be able to:
...
+2. Complete [this Google Codelab](https://codelabs.developers.google.com/codelabs/firebase-web/#0), which will walk you through setting up a sample app on Firebase step by step.
+3. Go back to one of the previous projects (e.g. the library project or the to-do app), and [follow this guide](https://firebase.google.com/docs/web/setup?hl=en) to prepare it for interaction with a Firebase backend. 
+    1. **Caution:** If you're *not* going to be hosting your app on Firebase Hosting, but prefer to leave it where it is (e.g. on GitHub Pages), take a closer look at the section below Step 4, there should be a link, which directs you to the "Available Libraries" page. Then, use what you learned in the codelab to make your own app actually interact with Firebase.
```
CouchofTomato, line 56: Is there a reason why this needs to be numbered?

Also it makes reference to a step 4 which I believe doesn't exist any longer. Probably a good time to handle this if you can.

CouchofTomato (changes requested):
Thanks @xari4808 

Just a small comment

javascript/js-rails/rails_backend.md
xari4808: > Is there a reason why this needs to be numbered?

That's an attempt to increase readability. We could use a bullet point here instead of numbering it, what do you think?

>  it makes reference to a step 4 which I believe doesn't exist any longer

Not quite sure what this means. There seems to exist a step 4 with a link to the "available libraries" page

image

CouchofTomato: lol ok. Yeah step 4 in the linked guide. I assumed it was step 4 in the lesson. My bad

I think a bullet point would work better than a numbered list. Can you make that change?

CouchofTomato (approved):
Thanks @xari4808

PR #23202, opened 2021-12-17
Update head element section in html-boilerplate.md

yogeshwarbari:
There is a statement "Anything included within the head element will not be displayed to the user." which may be misinterpreted by beginners that "whatever you write inside <head> and </head> tags even the plain text should not be displayed on webpage by browser." which is wrong. I think the statement might create this misconception which will remain in their mind until they try stuff like putting random text in head tag and then they can see the output on screen.

...your text here

#### 2. Related Issue

Closes #XXXXX

### DIFF
--- a/foundations/html_css/html-foundations/html-boilerplate.md
+++ b/foundations/html_css/html-foundations/html-boilerplate.md
@@ -50,6 +50,7 @@ Back in the `index.html` lets add the `<html>` element by typing out its opening
 ### Head Element
 
 The `<head>` element is where we put important meta-information **about** our webpages, and stuff required for our webpages to render correctly in the browser.
+Inside the `<head>`, we **should not** use any element that displays content on the webpage.
 
 #### Title Element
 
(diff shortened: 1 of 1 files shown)


foundations/html_css/html-foundations/html-boilerplate.md
```diff
@@ -49,7 +49,7 @@ Back in the `index.html` lets add the `<html>` element by typing out its opening
 
 ### Head Element
 
-The `<head>` element is where we put important meta-information **about** our webpages, and stuff required for our webpages to render correctly in the browser. Anything included within the head element will **not** be displayed to the user.
+The `<head>` element is where we put important meta-information **about** our webpages, and stuff required for our webpages to render correctly in the browser.
```
fortypercenttitanium, line 52: I like the idea behind this change, but I don't like simply removing this sentence. Do you feel like you could reword it to be more accurate, while also emphasizing that elements within the `head` element *should not* be displayed to the user?

fortypercenttitanium (changes requested):

foundations/html_css/html-foundations/html-boilerplate.md
yogeshwarbari: When I raised the PR I was not able to think proper replacement for the removed statement but now after thinking some more time I can reword the correct statement.

fortypercenttitanium (approved):
Thank you!

PR #23206, opened 2021-12-17
Added Link to bcrypt implementation for Secure Password

don-hue:

...your text here

#### 2. Related Issue

Closes #XXXXX

### DIFF
--- a/nodeJS/authentication/Authentication.md
+++ b/nodeJS/authentication/Authentication.md
@@ -297,7 +297,7 @@ Once it's installed you need to require it at the top of your app.js and then we
 
 Password hashes are the result of passing the user's password through a one-way hash function, which maps variable sized inputs to fixed size pseudo-random outputs. 
 
-Edit your `app.post("/sign-up")` to use the bcrypt.hash function which works like this:
+Follow this [article](https://coderrocketfuel.com/article/store-passwords-in-mongodb-with-node-js-mongoose-and-bcrypt) along and edit your `app.post("/sign-up")` to use the bcrypt.hash function which works like this:
 
 ~~~javascript
 bcrypt.hash("somePassword", 10, (err, hashedPassword) => {


don-hue:
Following the NodeJS curriculum, I noticed a missing link during the authentication lesson (Securing passwords with bcrypt).
The description/instructions were unclear, so I found an article on how to implement bcrypt hashing/comparing. I added said article in the text as a hyperlink.

fortypercenttitanium (commented):
It seems to me like the description of how to use bcrypt is already sufficient. I skimmed through the article and it goes through a lot of procedure that the learner should have done already. Can you explain what exactly was unclear about the lesson instructions that would necessitate reading this rather lengthy article to perform this step?

don-hue:
Hey man, 
you are right. there is some overlap in the article and the curriculum. however, there is one major step which was pretty unclear to me: 

the part were I had to edit my ('/sign-up) route to use bcrypt.hash. 
-> in the article, you have to define a so called pre-middleware for mongoose: 

UserScheme.pre("save", function.......)
-> this is basically a function which will be called BEFORE we save our user and password to our MongoDB. In this step we take the user password, hash it and than we proceed to save the hashed password in our MongoDB.

fortypercenttitanium:
> Hey man, you are right. there is some overlap in the article and the curriculum. however, there is one major step which was pretty unclear to me:
> 
> the part were I had to edit my ('/sign-up) route to use bcrypt.hash. -> in the article, you have to define a so called pre-middleware for mongoose:
> 
> UserScheme.pre("save", function.......) -> this is basically a function which will be called BEFORE we save our user and password to our MongoDB. In this step we take the user password, hash it and than we proceed to save the hashed password in our MongoDB.

That is certainly one way to do it, but it is not the only way. Alternatively, you could handle the hashing in the route handler before you call `User.save()`. I think that is what we are expecting the learner to do here since it is basically the path of least resistance. The `.pre` is a handy function but it's not absolutely necessary to complete this step.

My concern is that by adding this article, we are lengthening the time it will take for someone to complete this lesson for little benefit in the end. If you feel like this part of the lesson is unclear or could be better without this article, we'd be open to that.

Before I close, I'll let you respond in case I'm missing something that you'd like to point out.

zachmmeyer:
Closed due to staleness.

PR #23255, opened 2021-12-31
Update CSS Foundations lesson wording for clarity in Chaining Selectors section

bryansoftdev:
 - [ ] If the PR is related to an open issue, use a [relevant keyword](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) and reference it with the `#` sign and the issue number.

If you lookup "multiple type selectors" on the Discord you can see some confusion about what is meant by "multiple type selectors". I reordered the sentences, included the word "syntax", and added some more detail to make it clear that the paragraph is talking about how this syntax wouldn't work for type selectors.

#### 2. Related Issue

Closes #XXXXX

### DIFF
--- a/foundations/html_css/css-foundations.md
+++ b/foundations/html_css/css-foundations.md
@@ -157,9 +157,7 @@ We have two elements with the `subsection` class that have some sort of unique s
 }
 ~~~
 
-What `.subsection.header` does is it selects any element that has both the `subsection` *and* `header` classes.
-
-Notice how there isn't any space between the `.subsection` and `.header` selectors. This syntax basically works for chaining any combination of selectors except for multiple type selectors. This is because `divp` would try to select a literal `<divp>` element which doesn’t exist, and either way, an element can’t be two different types at once.
+What `.subsection.header` does is it selects any element that has both the `subsection` *and* `header` classes. Notice how there isn't any space between the `.subsection` and `.header` selectors. This syntax basically works for chaining any combination of selectors except for multiple type selectors. For example, if you tried to chain the `div` and `p` type selectors like `divp`, then the selector would try to select a literal `<divp>` element which doesn’t exist. In general, a type selector can't be chained since an element can’t be two different types at once.
 
 #### Descendant Combinator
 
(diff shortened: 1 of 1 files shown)


foundations/html_css/css-foundations.md
```diff
@@ -157,7 +157,9 @@ We have two elements with the `subsection` class that have some sort of unique s
 }
 ~~~
 
-Notice how there isn't any space between the two selectors here. What `.subsection.header` does is it selects any element that has both the `subsection` *and* `header` classes. This basically works with any combination of selectors except multiple type selectors (since an element can't be two different types at once, and `divp` would try to select a literal `<divp>` element which doesn't exist).
+What `.subsection.header` does is it selects any element that has both the `subsection` *and* `header` classes. Notice how there isn't any space between the `.subsection` and `.header` selectors. This syntax basically works for chaining any combination of selectors, with the exception of type selectors.
+
+In general, type selectors can't be chained since an element can’t be two different types at once. Also, chaining two type selectors like `div` and `p`, would give us the selector `divp`, which wouldn't work since the selector would try to find a literal `<divp>` element, which doesn’t exist.
```
thatblindgeye, line 162: I agree the verbiage could be made a little clearer, however I think this could be tweaked a little further. Do you feel something like the below suggestion would ease some of that confusion, as it does seem like the confusion may stem from the "type" word being misconstrued as "you can't use more than one **type of** selector".

```suggestion
What `.subsection.header` does is it selects any element that has both the `subsection` *and* `header` classes. Notice how there isn't any space between the `.subsection` and `.header` class selectors. This syntax basically works for chaining any combination of selectors, with the exception of chaining more than one type (or element) selector.

In general, you can't chain more than one type/element selector since an element can’t be two different types at once. For example, chaining two type selectors like `div` and `p`, would give us the selector `divp`, which wouldn't work since the selector would try to find a literal `<divp>` element, which doesn’t exist. 
```

bryansoftdev: I committed your suggestion, then made another commit where I just made a link to the "Type Selectors" section. Let me know what you think :)

PR #23263, opened 2022-01-02
Add notice about VirtualBox not supporting Win 11 as host OS to install guide lessons

MchaseCov:
 - [X] You have previewed your markdown (if any) in the [Odin Markdown Preview Tool](https://www.theodinproject.com/lessons/preview)

Windows 11 is not currently supported or listed on the [Host OS](https://www.virtualbox.org/manual/UserManual.html#hostossupport) list for VirtualBox. According to various forum & Reddit threads, users are having headaches and unable to run Vbox on Windows 11 at all. There *are* some community-made compatibility patches and settings and tweaks to make it run. I didn't mention those because I felt that it may be read as TOP encouraging this route and being responsible if there is a failure. 

I felt like this was important to add since Windows 11 is only going to be more popular with time and somebody on Win 11 may not realize what is going wrong until they've already tried to install and it failed; Or it runs, crashes, and they potentially corrupt or lose data entirely.

I added this notice to both the Vbox install guide section as well as the preceding lesson that states Windows is not supported. The latter notice may not be needed, but I felt like it was because Win 11 users should know ahead of time that Vbox is not supported before getting attached to that method and preparing to do it, only to read on the next page that it isn't an option.

#### 2. Related Issue

N/A no related issue

### DIFF
--- a/foundations/installations/installation_overview.md
+++ b/foundations/installations/installation_overview.md
@@ -46,6 +46,16 @@ A **virtual machine** is an emulation of a computer that runs within your existi
 
 Dual-booting can be as easy as inserting a flash drive and clicking a few buttons. The benefits of dual-booting cannot be overstated. You will allow Linux to access the full capabilities of your hardware, have a clean and distraction-free environment for coding, and learn the platform used by many senior developers and servers around the world.
 
+#### NOTICE FOR WINDOWS 11 USERS:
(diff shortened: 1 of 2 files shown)


zachmmeyer (approved):
Looks good to me! Thanks for adding this.

PR #23317, opened 2022-01-13
Changed Git setup to reflect the latest SSH ed25519 recommendations from Github Docs

ManonLef:
 - [X] You have previewed your markdown (if any) in the [Odin Markdown Preview Tool](https://www.theodinproject.com/lessons/preview)

Changed the lesson to setup an SSH with an Ed25519 algorithm. This is the latest method in the Github Docs. https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

In the current version it's RSA focused. It also states that if after the command `ls ~/.ssh/id_rsa.pub` the terminal shows: “No such file or directory”, there's no SSH key present. This is incorrect as it merely means there's no id_rsa.pub, we don't know with that command if the directory is present, and we only searched for an id_rsa.pub file, not the other options like Ecdsa or Ed25519. Changing this to specify exactly what we were looking for, namely an id_ed25519.pub version in this change, makes this part correct. 

Currently there are quite a few help requests on discord from people who state that they already have an SSH key but who still get the  “No such file or directory”. These threads and proposed solutions often end up in generating a new RSA key through TOP instructions while they might already have an Ed25519 installed from different tutorials for example.

Long story short, I think this update is more current and more correct with the Ed25519 focus. Plus it might lead to fewer help requests with the incorrect conslusion from the “No such file or directory” reworded. 

This issue and the proposed changes have been discussed in discord with BriggsE here: https://discord.com/channels/505093832157691914/931168767281610832/931170082674065459

Some decisions were made in the linked Discord thread regarding the choice for Ed25519 specifically and thus diverting a bit from the Github Docs for simplicity's sake.

#### 2. Related Issue
N/A

Closes #XXXXX

### DIFF
--- a/foundations/installations/setting_up_git.md
+++ b/foundations/installations/setting_up_git.md
@@ -136,10 +136,10 @@ An SSH key is a cryptographically secure identifier. It's like a really long pas
 First, we need to see if you have an ed25519 SSH key already installed. Type this into the terminal and check the output with the information below:
 
 ~~~bash
-ls ~/.ssh/id_rsa.pub
+ls ~/.ssh/id_ed25519.pub
 ~~~
 
-If a message appears in the console containing the text "No such file or directory", then you do not yet have an SSH key, and you will need to create one. If no such message has appeared in the console output, you already have a key and can proceed to step 2.4.
+If a message appears in the console containing the text "No such file or directory", then you do not yet have an ed25519 SSH key, and you will need to create one. If no such message has appeared in the console output, you can proceed to step 2.4.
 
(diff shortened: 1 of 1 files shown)


I3uckwheat (approved):
Looks good to me, thanks for taking this on!

I'll defer to @zachmmeyer and @xandora for feedback though

zachmmeyer (approved):
Just tested this out and it works for me! Thanks for getting this out!

xandora:
Looks good to me.

PR #23397, opened 2022-01-25
Explained the Regular expressions pattern used in the Pattern Validation section and provided a link to an external resource in the Additional Resources section of form_validation.md

Artymanprince:
- [✅ ] You have read [The Odin Projects Contributing Guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md).
 - [✅ ] The title summarizes the change and where it happened, for example: "Fixes punctuation in Clean Code lesson".
 - [✅ ] If the PR is related to an open issue, use a relevant keyword and reference it with the # sign and the issue number.
 - [✅ ] You have previewed your markdown (if any) in the [Odin Markdown Preview Tool](https://www.theodinproject.com/lessons/preview)
 - [✅ ] If changes are requested, please make the changes in a timely manner. After you submit the changes, request another review from the maintainer (top right).

I have explained the Regex pattern used in this section "(\\d{5}([\\-]\\d{4})?)" and provided a link to an article that clearly explains how to apply Regex to most sought after data in a database which also includes the zip code(Number) format used.

Closes #XXXXX

### DIFF
--- a/html_css/intermediate_html/form-validations.md
+++ b/html_css/intermediate_html/form-validations.md
@@ -113,15 +113,7 @@ Try submitting the form with seven passengers to see the validation in action.
 To ensure we get the correct information from users, we will often want to ensure data matches a particular pattern.
 Real-world applications would be checking a credit card number or a zipcode is in the correct format.
 
-To add a pattern validation, we give the form control a `pattern` attribute with a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) as the value. In our example we are using the pattern validation to ensure a US zip code is in the correct format. 5 numbers followed by an optional dash and 4 more numbers: Using the Regular expression pattern `(\\d{5}([\\-]\\d{4})?)`. How exactly do we interpret this expression? Let’s take this step by step, going from left to right:
-<ul> 
-  <li> Capturing group: ( ) </li>
-  <li> Escape character: \ </li>
-  <li> Digit: \d </li>
-  <li> Number of characters: { } </li>
-  <li> Matches characters in brackets: [ ] </li>
-  <li> 0 or 1 character: ? </li>
-</ul>
+To add a pattern validation, we give the form control a `pattern` attribute with a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) as the value. In our example we are using the pattern validation to ensure a US zip code is in the correct format. 5 numbers followed by an optional dash and 4 more numbers:
 
 <p class="codepen" data-height="300" data-theme-id="dark" data-default-tab="html,result" data-slug-hash="YzrQqRK" data-preview="true" data-user="TheOdinProjectExamples" style="height: 300px; box-sizing: border-box; display: flex; align-items: center; justify-content: center; border: 2px solid; margin: 1em 0; padding: 1em;">
   <span>See the Pen <a href="https://codepen.io/TheOdinProjectExamples/pen/YzrQqRK">
--- a/html_css/intermediate_html/form-validations.md
+++ b/html_css/intermediate_html/form-validations.md
@@ -113,7 +113,15 @@ Try submitting the form with seven passengers to see the validation in action.
 To ensure we get the correct information from users, we will often want to ensure data matches a particular pattern.
 Real-world applications would be checking a credit card number or a zipcode is in the correct format.
 
-To add a pattern validation, we give the form control a `pattern` attribute with a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) as the value. In our example we are using the pattern validation to ensure a US zip code is in the correct format. 5 numbers followed by an optional dash and 4 more numbers:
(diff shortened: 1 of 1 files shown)


thatblindgeye (commented):
First of all, thank you for wanting to make your first contribution! It's always a big step. I just wanted to give a few comments on your issue and PR.

In regards to the issue you had opened up, it read more as you explaining changes you already made rather than requesting/suggesting an improvement. Usually an issue should be more along the lines of suggesting a new feature or pointing out a bug or something of the like. Issues should also be opened (and a user assigned) prior to any work being done, if it's not a smaller change like a typo fix (which don't really need an issue and a PR can just be made).

In regards to this PR, in the future I would try to be a little more brief with the title. Something along the lines of "Form Validation lesson: explain regular expressions".

Now, that said, I don't think we need to go out of the way to go over the different parts of the regex in this lesson or provide another regex link. The lesson already links to the MDN docs for regex for users to visit if they want to learn more about it, as well as a site with common regex patterns. I don't think we should put the focus on regex so much.

I want to keep this PR open for now so you can respond, as I would like to know whether you feel either of the two regex links in this lesson are lacking in any way, keeping in mind that _understanding_ regex isn't really the point of the lesson.

Artymanprince:
Thank you for the review. I only did that with the mindset of "improving" the lesson, although I understand _understanding_  regex is not the point of the lesson, hence why I felt it's better to at least have a glimpse of the pattern attribute used as other attributes were explained(min, max, minlength etc) briefly. And the Regex link provided to MDN is not as simplified as the link I posted and could be a whole lot of information for beginners even though it's provided as an _additional resource._
I hope I make sense as I myself am a learner.

Artymanprince:
"In regards to the issue you had opened up, it read more as you explaining changes you already made rather than requesting/suggesting an improvement.". Thank you. I'll be precise next time.

html_css/intermediate_html/form-validations.md
```diff
@@ -113,7 +113,15 @@ Try submitting the form with seven passengers to see the validation in action.
...
-To add a pattern validation, we give the form control a `pattern` attribute with a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) as the value. In our example we are using the pattern validation to ensure a US zip code is in the correct format. 5 numbers followed by an optional dash and 4 more numbers:
+To add a pattern validation, we give the form control a `pattern` attribute with a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) as the value. In our example we are using the pattern validation to ensure a US zip code is in the correct format. 5 numbers followed by an optional dash and 4 more numbers: Using the Regular expression pattern `(\\d{5}([\\-]\\d{4})?)`. How exactly do we interpret this expression? Let’s take this step by step, going from left to right:
+<ul> 
+  <li> Capturing group: ( ) </li>
+  <li> Escape character: \ </li>
+  <li> Digit: \d </li>
+  <li> Number of characters: { } </li>
+  <li> Matches characters in brackets: [ ] </li>
+  <li> 0 or 1 character: ? </li>
+</ul>
```
thatblindgeye, line 124: I don't think we should include this information at all. What is stated here is already covered in the link you provided, and this doesn't really help explain *what* exactly the characters in the regex are doing. Revert this back to the original text:

```suggestion
To add a pattern validation, we give the form control a `pattern` attribute with a [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) as the value. In our example we are using the pattern validation to ensure a US zip code is in the correct format. 5 numbers followed by an optional dash and 4 more numbers:
```

html_css/intermediate_html/form-validations.md
```diff
@@ -197,6 +205,7 @@ It's also worth noting client-side validations are not a silver bullet for ensur
 - Check out [html5pattern](https://www.html5pattern.com/) for a list of commonly used pattern regular expressions you may find helpful.
 - Look through [this Twitter thread](https://twitter.com/vponamariov/status/1400388896136040454) of the do and don'ts for form validation UX.
 - [10 Guidelines](https://www.nngroup.com/articles/errors-forms-design-guidelines/) for form validation design
+- Have a look through [regular expressions](https://towardsdatascience.com/regular-expressions-clearly-explained-with-examples-822d76b037b4) clearly explained in this article.
```
thatblindgeye, line 208: After looking over this link, I think it could be beneficial for users who wish to look into it to see a regex pattern and have each piece walked through. That said, I think we should also include an MDN article to further explain things. Let's change this to:

```suggestion
- [Regular expressions clearly explained](https://towardsdatascience.com/regular-expressions-clearly-explained-with-examples-822d76b037b4) goes over several real world examples of regular expression patterns and walks through what they're doing step by step. This resource can be useful if you don't just want to copy + paste popular patterns, and instead want to know what a pattern is actually doing.
- One last resource on regular expressions that can be helpful is MDN's [regular expression syntax cheatsheet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet). This cheatsheet does a great job explaining the syntax of regular expressions in more detail.
```

thatblindgeye (changes requested):
Sorry for the delay in response. After thinking this over some more I think there is some compromise to be had. Please see my suggested changes below.

Artymanprince:
> Sorry for the delay in response. After thinking this over some more I think there is some compromise to be had. Please see my suggested changes below.

Thank you for the review, again :). I support the suggested changes.

thatblindgeye (approved):

PR #23458, opened 2022-01-31
line 61 - article issues

fgostev:

When you open the link on line 61 - "What is the difference between a block element and an inline element? " the article has issues. The article doesn't display the examples properly. The suggestion is to replace the article with another one.

#### 2. Related Issue

Not related issue


zachmmeyer:
- Your PR title does not match contributing guidelines
- What is wrong with the article? It looks weird, but it appears to be correct.

fgostev:
> 

Sorry, doing for the first time.

The wrong part of the article is the examples. Instead of visual examples (which it looks it is supposed to show them) - it has HTML code showing on the page. I found it to be confusing, but that's perhaps is just my perspective.

twalton83:
You should probably check your extensions or check with another browser. I don't see any HTML code.

fgostev:
Shows up now, perhaps was a glitch. Thank for checking!

PR #23494, opened 2022-02-05
 Semantic updates to iteration 0 in example 2 of reduce method section in basic_enumerable_methods.md

tunsmart:
Update example 2 of the reduce method section, under Iteration 0, to result["Bob's Dirty Burger Shack"] == 0 and result["St. Mark's Bistro"] == 0 since the idea is to make the point that the hash already has a default value of 0. The new codes will return true while the previous codes simply 'reasigns' the value 0 to those keys.

 - [ ] If the PR is related to an open issue, use a [relevant keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) and reference it with the `#` sign and the issue number.

This might be splitting hairs but I am suggesting a potential update to the reduce method section, in example 2 under Iteration 0. Since the idea is to make the point that the hash already has a default value of 0, isn't it a tad more correct in terms of the semantics to say that `result["Bob's Dirty Burger Shack"] == 0` and `result["St. Mark's Bistro"] == 0` and not `result["Bob's Dirty Burger Shack"] = 0` and `result["St. Mark's Bistro"] = 0`. The suggested codes will return `true` while the existing codes in essence simply 'reassign' the value 0 to those keys.

#### 2. Related Issue

Closes #XXXXX

### DIFF
--- a/ruby_programming/basic_ruby/basic_enumerable_methods.md
+++ b/ruby_programming/basic_ruby/basic_enumerable_methods.md
@@ -333,7 +333,7 @@ Now that we know that this new hash with a default value of `0` is our accumulat
 
  1. Iteration 0:
     * result = {}
-    * Remember, this hash already has default values of `0`, so `result["Bob's Dirty Burger Shack"] = 0` and `result["St. Mark's Bistro"] = 0`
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (approved):
I really like this change, thank you for contributing it!

tunsmart:
Thanks, @wise-king-sullyman. Happy to be helpful.

PR #23504, opened 2022-02-07
Refactor instructions in Project: Rock Paper Scissors

### LINKED ISSUE #23500: Refactor Rock Paper Scissors Instructions
Many learners have trouble interpreting the RPS instructions, which leads to poor outcomes in their code and anti-patterns. 

We should: 
- Refactor the instructions to change the language so `game()` does not hold declarations for `playRound` and other functions inside of it
- Show boilerplate for a simple `for` loop so they can have a better `game()` function instead of punting loops to a later lesson
- Update "Important note: you want to return the results of this function call, not console.log() them." to explain that we want to return the results so we may use them elsewhere

scheals:
 - [x] If the PR is related to an open issue, use a [relevant keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) and reference it with the `#` sign and the issue number.

Due to problems with users declaring the `playRound()` function in the `game()` function instead of calling it I have changed the wording and included a separate point in the instructions to make sure this does not happen.

The important note has been expanded to make it clear that the `return` of the function is the most important and that `console.log()` is there to make sure that the function is returning the expected result. This is due to users not using the returned values.

The instructions now feature a boilerplate for a `for` loop instead of a note about them to prevent poor code and to solidify the concept of loops learned in Problem Solving lesson.

#### 2. Related Issue

Closes #23500

### DIFF
--- a/foundations/javascript_basics/project_rock_paper_scissors.md
+++ b/foundations/javascript_basics/project_rock_paper_scissors.md
@@ -34,12 +34,11 @@ Finally, this is your first JavaScript program built from scratch, so don't forg
    1. Remember loops? This is a great opportunity to use one to play those five rounds:
 
       ~~~javascript
-      for (let i = 1; i <= 5; i++) {
+      for (let i = 0; i < 5; i++) {
          // your code here!
       }
       ~~~
 
-   1. Make sure there are no functions declared inside of the `game()` function.
    1. At this point you should be using `console.log()` to display the results of each round and the winner at the end.
    1. Use `prompt()` to get input from the user. [Read the docs here if you need to.](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt)
    1. Feel free to re-work your previous functions if you need to. Specifically, you might want to change the return value to something more useful.
--- a/foundations/javascript_basics/project_rock_paper_scissors.md
+++ b/foundations/javascript_basics/project_rock_paper_scissors.md
@@ -30,7 +30,7 @@ Finally, this is your first JavaScript program built from scratch, so don't forg
    console.log(playRound(playerSelection, computerSelection));
    ~~~
 
-6. Write a NEW function called `game()`. Call the previous function _inside_ of this one to play a 5 round game that keeps score and reports a winner or loser at the end.
+6. Write a NEW function called `game()`. Call the `playRound` function _inside_ of this one to play a 5 round game that keeps score and reports a winner or loser at the end.
    1. Remember loops? This is a great opportunity to use one to play those five rounds:
 
       ~~~javascript
--- a/foundations/javascript_basics/project_rock_paper_scissors.md
+++ b/foundations/javascript_basics/project_rock_paper_scissors.md
@@ -18,7 +18,7 @@ Finally, this is your first JavaScript program built from scratch, so don't forg
 4. Write a function that plays a single round of Rock Paper Scissors.  The function should take two parameters - the `playerSelection` and `computerSelection` - and then return a string that declares the winner of the round like so: `"You Lose! Paper beats Rock"`
    1. Make your function's playerSelection parameter case-insensitive (so users can input `rock`, `ROCK`, `RocK` or any other variation).
 
-5. **Important note:** you want to `return` the results of this function call, _not_ `console.log()` them. To test this function console.log the results:
+5. **Important note:** you want to `return` the results of this function call, _not_ `console.log()` them. You're going to use what you `return` later on, so let's test this function by using console.log to see the results:
 
    ~~~javascript
    function playRound(playerSelection, computerSelection) {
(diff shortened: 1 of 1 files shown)


foundations/javascript_basics/project_rock_paper_scissors.md
```diff
@@ -30,8 +30,16 @@ Finally, this is your first JavaScript program built from scratch, so don't forg
    console.log(playRound(playerSelection, computerSelection));
    ~~~
 
-6. Write a NEW function called `game()`. Use the previous function _inside_ of this one to play a 5 round game that keeps score and reports a winner or loser at the end.
-   1. You have not officially learned how to "loop" over code to repeat function calls... if you already know about loops from somewhere else (or if you feel like doing some more learning) feel free to use them. If not, don't worry! Just call your `playRound` function 5 times in a row. Loops are covered in the next lesson.
+6. Write a NEW function called `game()`. Call the previous function _inside_ of this one to play a 5 round game that keeps score and reports a winner or loser at the end.
```
twalton83, line 33: Can we say playRound instead of previous function?

foundations/javascript_basics/project_rock_paper_scissors.md
```diff
@@ -30,8 +30,16 @@ Finally, this is your first JavaScript program built from scratch, so don't forg
...
+   1. Remember loops? This is a great opportunity to use one to play those five rounds:
```
twalton83, line 34: Did we actually introduce loops by now? Also, can we start at 0 not 1?

twalton83 (changes requested):

foundations/javascript_basics/project_rock_paper_scissors.md
scheals: [Problem solving](https://www.theodinproject.com/paths/foundations/courses/foundations/lessons/problem-solving) lesson covers FizzBuzz and explains everything step-by-step, loop included. It is true that loops are covered in [Fundamentals Part 4](https://www.theodinproject.com/paths/foundations/courses/foundations/lessons/fundamentals-part-4) that come after the Project - after `Clean Code` and `Installing Node.js` lessons to be precise - so the users are expected to learn loops without picking up new concepts besides Arrays afterwards. They're ready and the loop they've seen is enough for their needs - they're supposed to call their `playRound()` function five times so it's only matter of putting it there. The only thing left to figure out is where to put win variables and how to interact with them.


We can start at 0 not 1, I'll just make a small case for this to make sure we're not losing some potential upside.
Problem Solving lesson also uses start at 1 and gives this reasoning while explaining how the loop works:
>If you haven’t seen this before and it looks strange, it’s actually straightforward. We declare a variable i and assign it 1: the initial value of the variable i in our loop (NOTE: Most of the time, programmers find themselves looping from 0. Due to the needs of our program, we’re starting from 1). The second clause, i <= answer is our condition. We want to loop until i is greater than answer. The third clause, i++, tells our loop to increment i by 1 every iteration. As a result, if the user inputs 10, this loop would print numbers 1 - 10 to the console.

Obviously, there's no *need* for it to start at 1 here but it might feel more clear thanks to familiarity and it might feel "more right" to them, since it goes 1-5 as if it were Round 1, Round 2 and so on. Might prevent some unwanted tinkering with how the loop is constructed.

foundations/javascript_basics/project_rock_paper_scissors.md
scheals: Absolutely.

foundations/javascript_basics/project_rock_paper_scissors.md
twalton83: > Obviously, there's no need for it to start at 1 here
Since we've established both that and convention tends to be starting at 0, I think we should go with 0. I also don't want people to default to starting at 1.

twalton83: Some things are just as much about the ability to follow the instructions (AKA, not "tinkering"), you know?

foundations/javascript_basics/project_rock_paper_scissors.md
```diff
@@ -30,8 +30,16 @@ Finally, this is your first JavaScript program built from scratch, so don't forg
...
+6. Write a NEW function called `game()`. Call the `playRound` function _inside_ of this one to play a 5 round game that keeps score and reports a winner or loser at the end.
+   1. Remember loops? This is a great opportunity to use one to play those five rounds:
+
+      ~~~javascript
+      for (let i = 1; i <= 5; i++) {
+         // your code here!
+      }
+      ~~~
+
+   1. Make sure there are no functions declared inside of the `game()` function.
```
twalton83, line 42: I actually think this may serve to confuse since the instructions earlier now make it clear one should only *call* the function, and there isn't language that would lend one to declare `playRound` inside of `game`.

foundations/javascript_basics/project_rock_paper_scissors.md
scheals: Makes absolute sense all you've written. Going to change it so it starts at 0. I assume that using a loop is a go, correct?

foundations/javascript_basics/project_rock_paper_scissors.md
scheals: I checked couple of recent RPS projects and none of them had any functions declared inside of them. As such, I am going to delete this line.

foundations/javascript_basics/project_rock_paper_scissors.md
twalton83: Yes, let's please use a loop.

twalton83 (approved):

PR #23523, opened 2022-02-09
Fixes transform order in animations lesson.

florianbuehler:
 - [x] If the PR is related to an open issue, use a [relevant keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) and reference it with the `#` sign and the issue number.

The described order for multiple transforms in the _Advanced HTML and CSS_ course (animations section) is incorrect. It is not applied from right to left, but from left to right. 

This can be observed in the codepen or see e.g. [https://riptutorial.com/css/example/4130/multiple-transforms](https://riptutorial.com/css/example/4130/multiple-transforms).

#### 2. Related Issue

No issue relating to that bug exists.

### DIFF
--- a/html_css/animation/transforms.md
+++ b/html_css/animation/transforms.md
@@ -142,7 +142,7 @@ There are two boxes located at the same position. We chained `rotate` and `trans
 </p>
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-If you guessed correctly, congratulations! Otherwise, you've now learned that when chaining multiple transforms, each transform function is applied from right to left. 
+If you guessed correctly, congratulations! Otherwise, you've now learned that when chaining multiple transforms, each transform function is applied from left to right. 
 
 While you can generally chain multiple transforms in any order for various results, there is one exception: `perspective`. This brings us nicely to the next section where `perspective` is involved.
 
--- a/html_css/animation/transforms.md
+++ b/html_css/animation/transforms.md
@@ -144,7 +144,7 @@ There are two boxes located at the same position. We chained `rotate` and `trans
 
 If you guessed correctly, congratulations! But this is a tricky concept. There is a bit of debate on how to read a chain of transform functions. According to [MDN's transform docs](https://developer.mozilla.org/en-US/docs/Web/CSS/transform#values): "The transform functions are multiplied in order from left to right, meaning that composite transforms are effectively applied in order from right to left."
 
-To learn more about the "left to right" vs "right to left" interpretations, check out this visual article on [chaining transforms.](https://codepen.io/bali_balo/post/chaining-transforms).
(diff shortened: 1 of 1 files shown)


html_css/animation/transforms.md
```diff
@@ -142,7 +142,7 @@ There are two boxes located at the same position. We chained `rotate` and `trans
 </p>
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-If you guessed correctly, congratulations! Otherwise, you've now learned that when chaining multiple transforms, each transform function is applied from right to left. 
+If you guessed correctly, congratulations! Otherwise, you've now learned that when chaining multiple transforms, each transform function is applied from left to right.
```
dm-murphy, line 145:
```suggestion
If you guessed correctly, congratulations! But this is a tricky concept. There is a bit of debate on how to read a chain of transform functions. According to [MDN's transform docs](https://developer.mozilla.org/en-US/docs/Web/CSS/transform#values): "The transform functions are multiplied in order from left to right, meaning that composite transforms are effectively applied in order from right to left."

To learn more about the "left to right" vs "right to left" interpretations, check out this visual article on [Chaining transforms.](https://codepen.io/bali_balo/post/chaining-transforms) 
```

dm-murphy: Thanks for bringing this issue up! It's actually a bit more nuanced than it first appears. If your interested in reading up more about this "left to right" vs "right to left" debate you can check out this Stack Overflow discussion here: https://stackoverflow.com/questions/27635272/css3-transform-order-matters-rightmost-operation-first

But to clarify this lesson I've added some suggested changes to your PR that would include the MDN docs explanation as well as a visual article that showcases a good example on the interpretation.

If you agree with my suggestions you can commit these changes. Thanks again for pointing out this issue.

dm-murphy (changes requested):

html_css/animation/transforms.md
florianbuehler: Thanks, a lot for the input! I didn't know that - it's super interesting. Seems you learn something new every day :) 

I added your suggestions to the PR.

zachmmeyer: @florianbuehler If you are still able to, I would hit the commit suggestion button. It looks like an extra period was added at the end of line 147.

florianbuehler: @zachmmeyer unfortunately I'm not able to hit the button anymore (it says it's already outdated). I'm sorry for that - I never used the suggestion option of GitHub before. However, I will remember it for the next time.

florianbuehler: I just noticed that I added one period too much - I didn't see the one in `[Chaining transforms.]`. Fixed it.

dm-murphy (approved):
Thanks!

PR #23630, opened 2022-02-18
Project: Recursion: add example input/output for fibonacci challenge

JoshDevHub:
[Link to Recursion Project](https://www.theodinproject.com/paths/full-stack-ruby-on-rails/courses/ruby-programming/lessons/recursion)

I'm requesting this change because of noticing how often people miss what exactly is being asked in the Fibonacci challenge. It wants the student to write a method that can build an array with n digits of the Fibonacci sequence, but people often interpret it to mean returning only the nth number in the sequence. I think just a short example input/output is likely to clear potential misunderstandings up.

I also think this is an important distinction to make because the 'return the nth digit of the sequence' is a problem that's already covered when [these exercises](https://www.codequizzes.com/computer-science/beginner/recursion) are included in the [lesson prior to this project](https://www.theodinproject.com/paths/full-stack-ruby-on-rails/courses/ruby-programming/lessons/recursive-methods). It provides a better challenge to learners to tackle the problem with the new wrinkle of building an array recursively versus just solving the same problem twice back-to-back.

- [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
- [x] The title of this PR follows the `where the change occurs: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`

### DIFF
--- a/ruby_programming/computer_science/project_recursion.md
+++ b/ruby_programming/computer_science/project_recursion.md
@@ -5,7 +5,7 @@ The [Fibonacci Sequence](http://en.wikipedia.org/wiki/Fibonacci_number), which s
 ### Assignment 1
 
 <div class="lesson-content__panel" markdown="1">
-  1. Using iteration, write a method `#fibs` which takes a number and returns an array containing that many numbers from the fibonacci sequence.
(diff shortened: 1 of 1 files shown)


zachmmeyer (approved):
Thanks for adding this!!!

PR #23636, opened 2022-02-20
Template Files: Update based on main TOP repo

thatblindgeye:
- [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
- [x] The title of this PR follows the `where the change occurs: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`

- [ ] I have previewed all lesson file(s) included in this PR with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure the Markdown content is formatted correctly.
- [ ] If one exists, I have linked a related open issue to this PR in Step 2 below
- [ ] If changes were requested, I have made them and re-requested a review from the maintainer (top of the right sidebar)

**1. Description of the Changes**
 
Updated the template files to align with the main [TOP repo templates](https://github.com/TheOdinProject/theodinproject/tree/main/.github), which is currently the format we should be using moving forward (as closely as possible)

**2. Related Issue**

Closes #XXXXX

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,38 +1,54 @@
 ---
 name: Bug Report
-about: Create a report to help us improve
-title: ''
-labels: 'Status: Needs Review, Type: Bug'
-assignees: ''
-
+about: Create a report to help us improve something that is not working correctly
+title: "Bug - :"
+labels: "Status: Needs Review, Type: Bug"
+assignees: ""
 ---
+
 <!-- Thank you for taking the time to submit a bug report to The Odin Project. In order to get issues closed in a reasonable amount of time, you must include a baseline of information about the bug in question. Please read this template in its entirety before filling it out to ensure that it is filled out correctly. -->
 
-<!-- Complete the following REQUIRED checkboxes by replacing the whitespace between the square brackets with an 'x', e.g. [x]. -->
-- [ ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-- [ ] The title of this issue follows the `where the change occurs: brief description of bug` format, e.g. `React section: Knowledge Checks don't link to resource`
+Complete the following REQUIRED checkboxes:
+-   [ ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
+-   [ ] The title of this issue follows the `Bug - location of bug: brief description of bug` format, e.g. `Bug - React section: Knowledge Checks don't link to resource`
 
-<!-- The following checkbox is OPTIONAL. Completing it does not guarantee you will be assigned this issue, but rather lets us know you are interested in working on it. -->
-- [ ] I would like to be assigned this issue to work on it
+The following checkbox is OPTIONAL:
+<!-- Completing this checkbox does not guarantee you will be assigned this issue, but rather lets us know you are interested in working on it. -->
+-   [ ] I would like to be assigned this issue to work on it
 
-**1. Description of the Bug**
+<hr>
+
+**1. Description of the Bug:**
 <!-- A clear and concise description of what the bug is. Include any screenshots that may help show the bug in action. -->
 
 
-**2. How To Reproduce**
-<!-- What steps one might need to take in order to reproduce this bug. -->
+**2. How To Reproduce:**
+<!--
+What steps one might need to take in order to reproduce this bug, e.g.:
+1. Log in
+2. Visit a lesson page
+3. Click the complete button
+4. The complete button does not update
+-->
 
 
-**3. Expected Behavior**
-<!-- A brief description of what you expected to happen. -->
+**3. Expected Behavior:**
+<!--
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,20 +1,31 @@
...
 
+**1. Because:**
+<!--
+If this PR closes an open issue, replace the XXXXX below with the issue number, e.g. Closes #2013. Or if the issue is in another TOP repo replace the #XXXXX with the URL of the issue, e.g. Closes https://github.com/TheOdinProject/curriculum/issues/XXXXX
 
-**2. Related Issue**
-<!-- If the PR is not related to any open issue, skip this step. 
-
-Otherwise, replace the XXXXX with the issue number, e.g. Closes #2013, or if the issue is in another TOP repo replace #XXXXX with the URL of the issue, e.g. Closes https://github.com/TheOdinProject/curriculum/issues/XXXXX -->
+Otherwise, provide a clear and concise reason for your pull request, e.g. what problem it solves or what benefit it provides. If this PR is related to, but does not close, another issue or PR, you can also link it as above without the 'Closes' keyword, e.g. "Related to #2013".
```
dm-murphy, line 18: The use of "otherwise" seems to imply an either or situation here. Is that the intent?

If the PR is related to an open issue and the contributor marks the closing issue number, should they not include a brief description as well?

dm-murphy (approved):
Everything looks good! I left one question on the PR template for the Because section.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,38 +1,54 @@
...
+<hr>
 
-**1. Description of the Bug**
+**1. Description of the Bug:**
 <!-- A clear and concise description of what the bug is. Include any screenshots that may help show the bug in action. -->
 
 
-**2. How To Reproduce**
-<!-- What steps one might need to take in order to reproduce this bug. -->
+**2. How To Reproduce:**
```
linkonsat, line 25: I think "All steps to produce" would be a bit more clearer. I could imagine someone reading this and leaving out a detail thinking they can just give a generic overview. Think it would be a bit helpful if we made it explicit in case they ignore the commented out parts.

linkonsat (changes requested):
Would we possibly want to include a desired behavior section? Just for formatting purposes so it's even more organized.

.github/PULL_REQUEST_TEMPLATE.md
thatblindgeye: I don't necessarily think a description is needed if the PR closes an open issue. The open issue would technically be the description as to the purpose/benefit of the PR or what problems are solved.

dm-murphy: Right, I wasn't thinking it was necessary but I was thinking more along the lines of: if I am making a PR that closes an open issue, am I not allowed to write a very brief description here as well? Most of the PRs I've seen are usually doing both the issue link and the brief description. Wondering if this change might throw them off.

One other thought on this section. Since the first line was removed: "A clear and concise description of your changes." And now the first line begins: "If this PR closes an open issue..." Do we think people might skip right over this section if it's not related to an issue? Therefore running the risk of no linked issue and no description?

thatblindgeye: If a description is included in addition to the issue being linked to close, that wouldn't really be a problem. The change may throw people off initially, but this is all in an attempt to make things more consistent across repos; the TOP repo has successfully used a format somewhat similar to this for a while, and after some discussion with Kevin we came to the decision to go with this format.

I think no matter what people will find an excuse to not read, though, and that there's only so much we can do to ensure they read things properly. Between telling users to read the template in its entirety and to read the Contributing Guide thoroughly, if users still don't fill things out correctly then that'll have happened no matter what. That's not to say I don't get where you're coming from, it's just I've seen others try to ensure users read things and they still find a way not to 😆

.github/ISSUE_TEMPLATE/bug_report.md
thatblindgeye: For the desired behavior, I would say the Expected Behavior section covers that.

For the Reproduce section, I could see this being an issue on a repo like the main TOP one. For this repo I don't think there would be as much need for super detailed steps to reproduce something since a lot of bugs would probably be able to be summed up in a single sentence (e.g. "Click on this link" if a link is broken, "Open this lesson in Chrome" if there's some issue with content rendering in Chrome, etc).

.github/PULL_REQUEST_TEMPLATE.md
dm-murphy: Gotcha! Hopefully all goes smoothly. If we do start getting a bunch of blank responses in that section, another idea would be adding in the word "REQUIRED" the way the checkbox sections do at the top. But people might choose not to see that either!

.github/ISSUE_TEMPLATE/bug_report.md
linkonsat: That's a fair point. I suppose it makes sense that a vast majority are going to be minor things. We can see how it goes and if it feels like this is an issue it can always be revisited. In that case it looks good otherwise! 👍

linkonsat (approved):

PR #23786, opened 2022-03-14
Updated Intermediate HTML and CSS/Creating a Grid/Gap Section

### LINKED ISSUE #23784: Creating a Grid lesson: update Gap section
**1. Description:**

The [Gap section of Creating a Grid lesson](https://www.theodinproject.com/paths/full-stack-javascript/courses/intermediate-html-and-css/lessons/creating-a-grid#gap) alludes that `gap` can _only_ be used to set `column-gap` and `row-gap` to the same value. In fact, `gap` is a shorthand that also allows us to set different values for `column-gap` and `row-gap`. The [MDN doc on gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) provides further clarification as well as live examples.

**2. Acceptance Criteria:**

- [ ] Replace the last sentence of the first paragraph with "Furthermore, we can use a shorthand property called `gap` to set both `row-gap` and `column-gap`."
- [ ] Add a paragraph after the third CodePen embed which tells users that they could "set both `row-gap` and `column-gap` using `gap`".
- [ ] The above paragraph should also mention the syntax of `gap` as well as encourage users to try it out in the embed.
- [ ] Depending on how the above changes are applied and if it becomes necessary, reword the last paragraph of this section.

If anyone wants to work on this issue, please comment to be assigned!

iawong:

-   [ x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)

-   [x ] I have previewed all lesson files included in this PR with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure the Markdown content is formatted correctly
-   [x ] I have ensured all lesson files included in this PR follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

<hr>

**1. Because:**

Closes #23784

**2. This PR:**

- Replaced the last sentence in the first paragraph with "Furthermore, we can use a shorthand property called gap to set both row-gap and column-gap."
- Added a paragraph after the 3rd CodePen telling users that they can set the row-gap and column-gap using gap, gave an example explaining the syntax of using gap, and encouraged users to try it out.
- Changed the last paragraph to explain how gap can be used to set both the row-gap and column-gap to the same value. 


html_css/grid-lessons/creating-a-grid.md
```diff
@@ -158,7 +158,9 @@ Finally we’ll add a lot of gap to our rows to highlight the difference:
 </p>
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-If we wanted the columns and rows to share the same gap value we could replace the `column-gap` and `row-gap` properties with something like `gap: 25px;`.
+Alternatively, we can set both the `row-gap` and `column-gap` by just using `gap`. For example: `gap: 10px 20px;` where the first value contains the `row-gap` and the second value contains the `column-gap`. Try using `gap` to change the `row-gap` and `column-gap` values with the example syntax.
```
01zulfi, line 161: Wording suggestion:
```suggestion
Alternatively, we can set both the `row-gap` and `column-gap` by just using `gap`. For example: `gap: 10px 20px;` where the first value contains the `row-gap` and the second value contains the `column-gap`. Try using `gap` to change the `row-gap` and `column-gap` values in the above CodePen embed. 
```

01zulfi (changes requested):
The changes look great. Just one requested change below.

html_css/grid-lessons/creating-a-grid.md
```diff
@@ -158,7 +158,9 @@ Finally we’ll add a lot of gap to our rows to highlight the difference:
 </p>
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-If we wanted the columns and rows to share the same gap value we could replace the `column-gap` and `row-gap` properties with something like `gap: 25px;`.
+Alternatively, we can set both the `row-gap` and `column-gap` by just using `gap`. For example: `gap: 10px 20px;` where the first value contains the `row-gap` and the second value contains the `column-gap`. Try using `gap` to change the `row-gap` and `column-gap` values with the example syntax. 
+
+If we wanted the columns and rows to share the same gap value we can omit the `column-gap` property so it looks something like `gap: 25px;`. When a `column-gap` isn't specified, the value of the `column-gap` is assumed to be the same value as the `row-gap`.
```
dm-murphy, line 163:
```suggestion
If we wanted the columns and rows to share the same gap value we can omit the `column-gap` property so it looks something like `gap: 25px;`. When a `column-gap` isn't specified, the value of the `column-gap` is assumed to be the same value as the `row-gap`.
```

dm-murphy (changes requested):
Although the changes here highlight some of the shorthand usage of the `gap` property, these were intentionally left out of the lesson when it was written. The point of the Creating a Grid lesson is to give a very simple walkthrough on the use of `grid` without overwhelming a beginner. Remember this is a lesson on creating a grid, not a lesson on `gap`.

While sometimes we touch upon shorthand properties, like this lesson's mention of `gap`, we also do our best not to overwhelm learners when learning a new concept with too much information or shorthand variations. It is easier for a beginner to visually separate `row-gap` and `column-gap` as two separate properties.

I do think the change on the first paragraph to this section is good (line 132) and further explains what the shorthand `gap` is doing : "Furthermore, we can use a shorthand..."

Perhaps the final paragraph could be reworded to imply that `gap` has other uses or learners could play around with it as a shorthand.

Regardless, the CSS Tricks article that is linked throughout the Assignments in this Grid curriculum series goes over all of the shorthand uses for `gap`.

01zulfi (changes requested):
Based on @dm-murphy's comment, we should update the final paragraphs.

@iawong Can you replace the final two paragraphs of the "Gap" section that you've added with:

"You can also try playing with the shorthand `gap` to set both the `row-gap` and `column-gap` in the above CodePen."

iawong:
Hey, I've removed the last 2 paragraphs and replaced them.

01zulfi (approved):
Looks great! Thanks for doing this 🚀

PR #23796, opened 2022-03-14
DOM Manipulation and Events: Add some consistent naming and expand a hint

danielyor:

<hr>

**1. Because:**

Based on my own experience as a learner of this lesson, I propose the following improvements.

**2. This PR:**

- Adds consistent naming for arguments in the code examples in "Append Elements" and "Remove Elements" sections.
- Expands on the reasoning of a given hint in the lesson exercise section.

### DIFF
--- a/foundations/javascript_basics/DOM_manipulation_and_events.md
+++ b/foundations/javascript_basics/DOM_manipulation_and_events.md
@@ -235,7 +235,7 @@ Add the following elements to the container using ONLY JavaScript and the DOM me
 3. a `<div>` with a black border and pink background color with the following elements inside of it:
    1. another `<h1>` that says "I'm in a div"
    2. a `<p>` that says "ME TOO!"
-   3. Hint for this one: after creating the `<div>` with createElement, append the `<h1>` and `<p>` to it before adding it to the container.
+   3. Hint for this one: after creating the `<div>` with createElement, append the `<h1>` and `<p>` to it before adding it to the container. Inserting into the DOM is a slow operation, so it's better to work in the memory first and then append the whole thing all at once.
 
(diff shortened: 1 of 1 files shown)


marvingay:
I would consider `newNode` to be more semantic/instructive for the method in question for the first change. The use of `childNode` to be a redundancy semantically, given the method name, in addition to the fact that in a tree structure, `child` is fairly commonplace / standard with intent being conveyed.

The anecdote on line 238 is somewhat ambiguous and doesn't add anything to the exercise, given where the learner's expertise might lie at this point.

Closing.

danielyor:
Thanks for the feedback to my PR! You've brought up some good points that I hadn't thought of so I'll try to keep those in mind in case of future contributions.

I understand your reasoning for the choices in naming and I can't disagree with them, it makes sense when you look at every single method individually. In the lesson however the methods are introduced one right after the other, so I would argue that they are perceived as a group of methods (in the learner's mind). Because of that, it can cause some confusion or a break of lesson flow when the naming for the node to be added/removed is different in every single one of the examples (childNode, newNode, child). That was the case in my own experience at least.

As for the addition in line 238, I proposed it because the hint given in the exercise section raised more questions to me than it gave answers/explanations. It suggested I do something (finalize changes inside the <div> before appending it to the DOM) but without explaining the reasoning behind it, even if the program would still be "correct" if I were to do it otherwise (ex. append the <div>, then append the <h1>, and then the <p>).

Lastly, I realize now that my explaining above should've had been in my initial PR so I apologize for how long this comment has become. If I come up with better proposals in the future I hope that you can welcome a new PR on this. Thanks again.

PR #23917, opened 2022-03-30
Block And Inline lesson: Fix grammar, make sentences easy to read, and format to follow the Layout Style Guide

lloydlobo:

<hr>

**1. Because:**

Closes N.A.

- This PR hopes to make the lesson easier to read, understand and digest. 
- Fixes minor grammatical and punctuation errors.
- It also formats the file to follow TOP's Layout Style Guide.

**2. This PR:**

- 489c57596a2795f8823e08faaee8776dd5260d75 & 3b35f959cf64a482b8daa0cbcc718c100c766700
   - Fixes punctuations in sentences. 
   - Adds and removes commas where it seemed necessary. 
- 51c984b3be8ca438fcffc536a19e753a2c9977d0
   - Adds hyphens to the word "in depth" making it "in-depth".
- 67e4b68a447136058548a7630fe8b96a0985e428
  - Provides simpler alternatives to long sentences.
  - Splits them up into easy to understand atomic sentences.
- 919008f4ad4bcb8a04a5f51179ed4464941681aa and 4282eb143005d60b46e4939b8b9fb8ae8408d48e
  - Fixes past errors made during the commits (editing).
  - Formats the markdown file to follow TOP's Layout Style Guide.

N.A.


thatblindgeye (changes requested):
It looks like you have some merge conflicts. Because there are currently dedicated issues for style guide updates going out right now, this could be a reason why. Depending what the changes look like afterwards, I will have to look into whether an already open issue exists to fix something trying to be fixed in this PR.

lloydlobo:
> It looks like you have some merge conflicts. Because there are currently dedicated issues for style guide updates going out right now, this could be a reason why. Depending what the changes look like afterwards, I will have to look into whether an already open issue exists to fix something trying to be fixed in this PR.

Hey there. It seemed like a word was different than that in main. LGTM now.

When do you figure the style guide will be updated?

foundations/html_css/block-and-inline.md
```diff
@@ -28,7 +32,7 @@ Div is a block-level element by default. It is commonly used as a container elem
 </p>
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-Span is an inline-level element by default. It is commonly used to group text content and inline HTML elements so we can apply styling to them.
+Span is an inline-level element by default. It is generally used to group text content and inline HTML elements to apply styling to them.
```
thatblindgeye, line 35: I'm kinda feeling like "commonly" makes more sense here, partly because that's the verbiage used for the div line further up, and because "generally" seems *too* restrictive is the best I can describe the feeling 😆 

I'm gonna leave this open for another day to mull it over. If anyone else has thoughts feel free to comment, or if you have any persuasive argument for the change definitely feel free to comment as well, @lloydlobo

thatblindgeye (commented):
The style guide updates I meant were issues created to bring lessons inline with the style guide, not updates to the style guide itself. 

Only have one other comment right now below.

foundations/html_css/block-and-inline.md
lloydlobo: The word _"commonly"_ does sound and feels to flow better with the overall context. Especially as you mentioned, it is being used along with `<div>` in the prior sentences.

After doing a quick research about the `<span>` element, I want to propose a new alternative. This would mean that neither _"commonly"_ nor _"generally"_ will be necessary in this case.

Newer Proposed Change:
>Span is an inline-level element by default. It groups text content and inline HTML elements for styling and is best used when no other HTML semantic element is available.

Since it's used as a last resort basis, it seems that its use could be kept to a bare minimum. As newer semantic elements are being introduced frequently, we may one day be able to do away with the good ol' `<span>`. :)

One downside that seems to be with the newer alternative is if the reader isn't familiar with semantic elements. What do you feel about that? 

We could go back with _"commonly"_ anyway. I feel this newer change may contribute to more complexity.

thatblindgeye: Ah no I was just trying to separate out the two options 😆  I don't think we should link to anything in that verbiage.

lloydlobo: Okay that makes sense and clears my doubt :). I feel we should go with the `[Option B]`. 
It's more appropriate between the two options (pun intended). 

For now this would look like:
>Span is an inline-level element by default. It can be used to group text content and inline HTML elements for styling, and should only be used when no other HTML semantic element is appropriate.

<sub>Note: Added _"semantic"_ as an Easter Egg for the user.</sub>

Do you have any suggestions to add to this?

thatblindgeye: I think the only thing I can think of would be a minor nitpick in reversing "HTML semantic" so that it reads "semantic HTML elements...", but this looks good! 💪🏼 Once you get a new commit pushed just re-request a review from me and I'll take another look over things.

lloydlobo: Done!

thatblindgeye (changes requested):
Can you resolve the merge conflicts? Should just have to update some things that have been merged into our main branch.

lloydlobo:
> Can you resolve the merge conflicts? Should just have to update some things that have been merged into our main branch.

Sure thing. It seems like descriptions of headings like "Knowledge Check" and "Lesson Overview" are being updated. Looking good 👍🏽

thatblindgeye (approved):

PR #23921, opened 2022-03-30
Advanced Selectors:  changed description of :only-child

Tzvi-H:

<hr>

**1. Because:**

Closes #XXXXX
Currently, :only-child is described as matching elements that have just one child.
I believe this is incorrect. :only-child matches elements that are an only child (or that don't have any siblings)

**2. This PR:**

- Updates the way `only-child` and `empty` are described

### DIFF
--- a/html_css/intermediate_css/selectors.md
+++ b/html_css/intermediate_css/selectors.md
@@ -130,7 +130,7 @@ Structural pseudo-classes are a powerful way to select elements based on their p
 
 <span id="first-child-knowledge-check">[`:first-child`](https://css-tricks.com/almanac/selectors/f/first-child/)</span> and [`:last-child`](https://css-tricks.com/almanac/selectors/l/last-child/) will match elements that are the first or last sibling.
 
-Similarly [`:empty`](https://css-tricks.com/almanac/selectors/e/empty/) and [`:only-child`](https://css-tricks.com/almanac/selectors/o/only-child/) will match elements that have no children at all, or that don't have any siblings.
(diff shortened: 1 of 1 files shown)


html_css/intermediate_css/selectors.md
```diff
@@ -130,7 +130,7 @@ Structural pseudo-classes are a powerful way to select elements based on their p
 
 <span id="first-child-knowledge-check">[`:first-child`](https://css-tricks.com/almanac/selectors/f/first-child/)</span> and [`:last-child`](https://css-tricks.com/almanac/selectors/l/last-child/) will match elements that are the first or last sibling.
 
-Similarly [`:empty`](https://css-tricks.com/almanac/selectors/e/empty/) and [`:only-child`](https://css-tricks.com/almanac/selectors/o/only-child/) will match elements that have no children at all, or just one.
+Similarly [`:empty`](https://css-tricks.com/almanac/selectors/e/empty/) and [`:only-child`](https://css-tricks.com/almanac/selectors/o/only-child/) will match elements that have no children at all, or that don't have any siblings.
```
thatblindgeye, line 133: I think we should also separate out what each of these do. Right now it could come off as _both_ of these selectors will either match element with no children or that don't have any siblings, rather than each matching different things.

So more along the lines of "Similarly empty does this, and only-child does this."

thatblindgeye (changes requested):

thatblindgeye (approved):

thatblindgeye:
Just an FYI, I updated your original comment to be more inline with how the PR template should be filled out. Otherwise thank you for you making this update! 🙂

PR #23927, opened 2022-03-31
Ruby - How this course will work: Improvements in grammar and punctuation

JM-Sully:

<hr>

**1. Because:**

The communication on the page could clearer.

**2. This PR:**

Makes small grammar and punctuation improvements.


ruby/introduction/how_this_course_will_work.md
```diff
@@ -13,7 +13,7 @@ You'll do a healthy chunk of learning in this course, but by the end of it all,
...
 
 ### The Path Ahead
-Ruby is a big language with lots of syntax, so we've broken the concepts up into smaller chunks to make it more digestible. Some of the early material will be fairly straightforward and will rely on simple exercises to help reinforce your understanding. As you progress through the course, the goal is for you to get a much deeper and more practical understanding of the material. As you move into some of the more advanced topics, you'll be reading less and building more--just the way it should be!
+Ruby is a big language with lots of syntax, so we've broken the concepts up into smaller chunks to make it more digestible. Some of the early material will be fairly straightforward and will rely on simple exercises to help reinforce your understanding. As you progress through the course, the goal is for you to get a much deeper and more practical understanding of the material. As you move into some of the more advanced topics, you'll be reading less and building more-- just the way it should be!
```
thatblindgeye, line 16: For the change that was made here, do you think it might be better to remove the double hyphens and use one with whitespace on either side (or an em/endash; can't remember which one might be applicable if either would be)? E.g.:

```suggestion
Ruby is a big language with lots of syntax, so we've broken the concepts up into smaller chunks to make it more digestible. Some of the early material will be fairly straightforward and will rely on simple exercises to help reinforce your understanding. As you progress through the course, the goal is for you to get a much deeper and more practical understanding of the material. As you move into some of the more advanced topics, you'll be reading less and building more - just the way it should be!
```

thatblindgeye (commented):
Thank you for wishing to contribute! Normally changes like these may not really get merged in due to the significance factor, but the grammar update I think is worth merging in. I only have one comment that requires a reply below.

ruby/introduction/how_this_course_will_work.md
JM-Sully: > For the change that was made here, do you think it might be better to remove the double hyphens and use one with whitespace on either side (or an em/endash; can't remember which one might be applicable if either would be)? E.g.:

Thanks for your feedback! I think the best option would actually be a single dash with no spaces on either side. Feels a bit odd but a bit of punctuation research tells me this is right.

thatblindgeye (approved):

PR #24009, opened 2022-04-12
CSS Functions lesson: Reword the linear-gradient example

themetar:

<hr>

**1. Because:**
Saying (for example) "the equation calculates the sides of the hypotenuse" sounds weird to me.

**2. This PR:**

* In css_math_functions.md, changed "calculates <arguments> of <result>" to  "use <arguments> to get <result>"

### DIFF
--- a/html_css/intermediate_css/css_math_functions.md
+++ b/html_css/intermediate_css/css_math_functions.md
@@ -18,7 +18,7 @@ color: rgb(0, 42, 255);
 background: linear-gradient(90deg, blue, red);
 ~~~
 
-Here, the value of `color` is the function `rgb()`, which accepts arguments in the form of numbers. It processes those numbers to calculate the rgb color corresponding to the three values given. Similarly, the `background` property has a value of `linear-gradient(90deg, blue, red)`, which calculates the angle, starting color, and ending color of a gradient.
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (changes requested):
I think I understand what you're saying, and I agreed with it at first, but after some further consideration, I think we both initially read this differently than it's written.

Another way this could be written is that the `linear-gradient` function *uses* arguments (90deg, blue, red in this case) to calculate and draw a gradient. Does that way of writing it out make more sense to you?

I don't disagree that the current wording seems a tad confusing, so if you have an alternate suggestion I'd like to hear it. But I believe that your currently suggested change means something different, as it implies that angle, start color, and end color are the only arguments that can be passed.

themetar:
My only objection was the grammatical wording of the sentence "..calculates the angle, starting color, and ending color of a gradient." It's like saying: The baker bakes the flower and yeast of the bread. What?! No, the baker _bakes bread_ (out of flower and yeast).

The paragraph is related to the css snippet, above it in the lesson, where linear-gradient is called with three parameters. It just restates the code, emphasizing that css functions are used. It might imply that that's the only way to use linear-gradient, but linear-gradient itself isn't the focus of the lesson. We can: add a clarification to that effect; add a link to docs (to nudge the student to see other uses of linear-gradient); or do nothing (as it was before). What should I do?

wise-king-sullyman:
@themetar I think just a small clarification would be enough. We don't need to go in any great depth about it since, as you said, that function isn't the focus of the lesson. I just don't want us to replace potentially misleading information with other information that is potentially misleading in another way 🙂

themetar:
@wise-king-sullyman I made some changes. Please review.

wise-king-sullyman (approved):
This looks great, thanks!

PR #24016, opened 2022-04-13
CSS Functions lesson: Fix max() function explanation

themetar:
modified:   html_css/intermediate_css/css_math_functions.md

<hr>

**1. Because:**

Previous content was at parts wrong

> The max function is most useful when the viewing window is either exceptionally large, or the user increases the content size by using the browser’s zoom feature.

Zooming is actually similar to reducing the window size

> Specifying the maximum possible size of a given element, for example, will prevent that element from infinitely filling the page as it expands.

Implies that the max() function sets the maximum allowed value for a property. max just selects the biggest value. A fixed value can be used to guard against _smaller_ values.

> From this list of given sizes, max will select the largest that will possibly fit within the viewing window.

max() doesn't choose by testing whether things would fit in the window   

**2. This PR:**

* Corrects the zoom/window size relation
* Deletes"setting the maximum value" part.
* Adds setting a minimum value as a common use-case
* Deletes "select the largest that will possibly fit within the viewing window" part.

### DIFF
--- a/html_css/intermediate_css/css_math_functions.md
+++ b/html_css/intermediate_css/css_math_functions.md
@@ -87,9 +87,7 @@ If there are not `150px` available, the image will switch to `100%` of the paren
 
 #### max()
 
-Max works the same way as min, only in reverse. It will select the largest possible value from within the parentheses.
-
-The max function is most useful when the viewing window is either exceptionally large, or the user increases the content size by using the browser’s zoom feature. Specifying the maximum possible size of a given element, for example, will prevent that element from infinitely filling the page as it expands.
+Max works the same way as min, only in reverse. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a _minimum_ allowed value for a property.
 
 Consider the following property of a given element:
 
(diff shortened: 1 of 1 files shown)


html_css/intermediate_css/css_math_functions.md
```diff
@@ -87,18 +87,17 @@ If there are not `150px` available, the image will switch to `100%` of the paren
 
 #### max()
 
-Max works the same way as min, only in reverse. It will select the largest possible value from within the parentheses.
-
-The max function is most useful when the viewing window is either exceptionally large, or the user increases the content size by using the browser’s zoom feature. Specifying the maximum possible size of a given element, for example, will prevent that element from infinitely filling the page as it expands.
+Max works the same way as min, only in reverse. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a _minimum_ allowed value for a property.
```
wise-king-sullyman, line 90: I feel like the idea of `max()` ensuring a *minimum* value feels a bit counterintuitive and might cause confusion, can you add just a touch of explanation here for why that's the case?

wise-king-sullyman (changes requested):
I like these changes as a whole, I just have one small change I'd like to see.

themetar:
@wise-king-sullyman I added to the code example an explanation for max's behavior. How does it look?

themetar:
Yup. Sorry. Fixed it.

wise-king-sullyman (approved):
Looks perfect, thanks for doing this!

PR #24088, opened 2022-04-30
Action Cable Assignment: Update 'Simple Messaging App' code-along 

thejwuscript:

<hr>

**1. Because:**

Turbolinks is no longer in active development, and Rails 7 comes with Hotwire which includes Turbo. It has essentially replaced turbolinks. Therefore "turbolinks:load" event listener no longer works in this lesson.

Another minor fix/typo involves clarifying the file path to connection.rb, so learners do not mistakenly create/duplicate the file at the wrong location.

**2. This PR:**

- Replace "turbolinks:load" with "turbo:load"
- Fix connection.rb file path to "app/channels/application_cable/connection.rb"

### DIFF
--- a/ruby_on_rails/mailers_advanced_topics/actioncable_lesson.md
+++ b/ruby_on_rails/mailers_advanced_topics/actioncable_lesson.md
@@ -226,7 +226,7 @@ Yours will have some comments that aren't important, but you will need to verify
 Right at the bottom of our `message_channel.js` file add the following code
 
 ~~~js
-document.addEventListener("turbolinks:load", () => {
+document.addEventListener("turbo:load", () => {
   let form = document.querySelector('#message-form')
   if(form) {
     form.addEventListener('submit', (e) => {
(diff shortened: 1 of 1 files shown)


ChargrilledChook:
Related issue: #24186

zfz:
@thejwuscript  Thank you so much for your PR, I could make the event listener work after hours of troubleshooting the original js code.  Another thing I found is that we should assign `consumer.subscriptions.create` to a const object `messageChannel`, which will be called later to send the message.  The line 4 of `app/javascript/channels/message_channel.js` generate by Rails: 
```js
consumer.subscriptions.create("MessageChannel", {
```

So in Line 207 of the current tutorial, we should remind user to add this `const messageChannel` declaration, like:

```
Now when the form is submitted, we need to intercept and handle it. Open up our `app/javascript/channels/message_channel.js` and take a look at the code as it stands.
```
to
```
Now when the form is submitted, we need to intercept and handle it. Open up our `app/javascript/channels/message_channel.js` and take a look at the code as it stands. Assign the consumer.subscriptions to a const messageChannel, which will send the message later.
```

CouchofTomato:
@ChargrilledChook Happy for this to be merged?

ChargrilledChook:
@CouchofTomato I think these changes are good, but it's possible we may have to update it a bit more - bit of discussion from a few learners on the issue here https://github.com/TheOdinProject/curriculum/issues/24186

ChargrilledChook:
Thanks for this contribution @thejwuscript 🚀

PR #24232, opened 2022-06-13
HTML Boilerplate lesson: Warn unexpected behavior with live server

rzliew:
I believe Live Server made some modification before running the code, which might lead to unexpected result being shown. If you attempt to include some Japanese characters inside a paragraph element, then change the charset in use to ISO-8859-1, the browser will still perfectly handle the displaying of those Japanese characters when using Live Server to run it, but not when running by directly opening the file. To reproduce the problem, follow the link https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML#specifying_your_documents_character_encoding. I believe this pull request can warn readers about the unexpected behavior so they are not surprise when they saw one. 

<hr>

**1. Because:**

Closes #XXXXX

**2. This PR:**


ChargrilledChook:
@rzliew Nice idea, but I think we can be a bit more specific. It seems the issue is that [live server uses UTF-8](https://github.com/ritwickdey/vscode-live-server/issues/587) no matter what and ignores our `meta charset` tag. 

What do you think about making your change a bit more specific to this behaviour?

xandora:
Does the Microsoft [Live Preview](https://marketplace.visualstudio.com/items?itemName=ms-vscode.live-server) extension cause the same problem? It's not *much* different from Live Server. Could be a potential alternative should the need arise.

rzliew:
@xandora I have tested the Microsoft Live Preview extension, the problem still persisted. I don't recommend it as an alternative as both extensions behave the same in this circumstance.

rzliew:
@ChargrilledChook Should I edit the original message and make a new commit already?

ChargrilledChook:
@rzliew if you want to have a go at that, that would be great 😃 happy to help get it right after you've made an edit, if needed

rzliew:
@ChargrilledChook I have updated the text, please have a look. If there's no other problem, I think it is ready for merging.

foundations/html_css/html-foundations/html-boilerplate.md
```diff
@@ -168,5 +168,5 @@ This section contains questions for you to check your understanding of this less
...
-*   If you wish, you can add the `lang` attribute to individual elements throughout the webpage. Read through [this doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) for a better understanding of the `lang` attribute.
\ No newline at end of file
+*   Another option for opening your HTML pages in the browser is using the [live server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) with VSCode. This will open your HTML document and automatically refresh it every time you save the document. However, we recommend not using this extension and instead doing it the old fashioned way, by opening the page and refreshing the page manually in the browser for your first few HTML projects. In this way you can get used to that process and won't become reliant on extensions right away. Using the extension may also lead to unexpected result in certain circumstances. One known issue is that live server extension uses UTF-8 as character encoding instead of the one defined in meta charset tag, which prevented you to identify actual characters supported by specific character encoding set.
```
ChargrilledChook, line 171:
```suggestion
*   Another option for opening your HTML pages in the browser is using the [live server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) with VSCode. This will open your HTML document and automatically refresh it every time you save the document. However, we recommend not using this extension and instead doing it the old fashioned way, by opening the page and refreshing the page manually in the browser for your first few HTML projects. In this way you can get used to that process and won't become reliant on extensions right away. Using the extension may also lead to unexpected result in certain circumstances. One known issue is that live server extension uses UTF-8 as character encoding instead of the one defined in meta charset tag, which prevented you to identify actual characters supported by specific character encoding set.
```
```suggestion
*   Another option for opening your HTML pages in the browser is using the [live server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) with VSCode. This will open your HTML document and automatically refresh it every time you save the document. However, we recommend not using this extension and instead doing it the old fashioned way, by opening the page and refreshing the page manually in the browser for your first few HTML projects. In this way you can get used to that process and won't become reliant on extensions right away. One reason is that there may be subtle differences when using extensions. For example, live server will always use UTF-8 character encoding and not the value defined in your `meta-charset` attribute. This could potentially hide some characters on your site not being encoded in the way you expect.
```
Good job! I think it reads a little better with some small changes like this, what do you think? Are you happy with this?

ChargrilledChook:
@rzliew Thanks very much for your first contribution! I like the general flow of your change, but I think we change the wording a little bit. I've added a comment on your change for you to have a look at when you get a chance :smiley:

xandora:
Random thought... Do NPM packages like `lite-server` or `live-server` do the same thing? They're obviously much more complex than needed for that stage of the curriculum, but it would highlight the usefulness of node even during early development stages.

rzliew:
@ChargrilledChook Yeah I think it looks pretty good! Also thank you for guiding me throughout my first ever open source commit, appreciate your hard work man

ChargrilledChook:
@rzliew My pleasure! Thank you, and look forward to seeing your next contribution :tada: :rocket:

rzliew:
@xandora I tested it and still the same result, does not resolve the issue

PR #24244, opened 2022-06-14
React-Testing-Part-One:Fix code in react-testing README.md which causes linter error

azazel-oss:
Error description says that destructuring from render should be avoided
hence using `screen.getByRole`

<hr>

**1. Because:**

Closes #XXXXX

**2. This PR:**

- This PR changes the code markdown inside the README.md file which used to destructure `getByRole` query from `render` object, but now after the fix uses `screen.getByRole`.


01zulfi (commented):
We actually cover this later on the lesson. This first example just shows one way of getting a query method for us to use. Nevertheless, I'm interested in what you think. Is the first example fine the way it is because we cover `screen` later on in the lesson, or should we update the first example to utilize `screen` instead of destructuring?


> React Testing Library provides the screen object which has all the methods for querying. With screen, we don’t have to worry about keeping render’s destructuring up-to-date. Hence, it’s better to use screen to access queries rather than to destructure render.

image

azazel-oss:
I did read through that part where it said that it's better to use `screen` than `render`, but since the first example was not passing through my linter I thought maybe it shouldn't be there in the first place. But now that I think about it, I guess, the lesson is fine the way it is, maybe just an emphasized mention that we should avoid `render` would suffice.

01zulfi:
@azazel-oss Great. You're welcome to modify this pull request and make the changes you proposed.

PR #24270, opened 2022-06-19
Feature: Add Stale Auto-labeler

ChargrilledChook:
**Because:**
* We should use automation to perform these useful but time consuming tasks to improve organisation

**This PR:**
* Adds a github action to auto label issues with stale after 30 days of inactivity.
* Will automatically remove the label if the issue is updated

For more information see the repo [here](https://github.com/actions/stale) or the docs [here](https://github.com/actions/stale). While this can be configured with a lot of different options, including automatically closing PRs / issues, I've gone for a simple and conservative approach at first to see if this has value.

Once a day this will run, labelling issues with 30 days of inactivity with `stale`. The label will be removed if there is activity on the issue again. At this stage there is no auto-closing.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,27 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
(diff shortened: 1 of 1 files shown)


CouchofTomato:
Thanks @ChargrilledChook 

I think this is a great addition.

PR #24293, opened 2022-06-24
Ruby Installation Lesson: Update with instructions for updating ruby-build

### LINKED ISSUE #24264: Ruby Installation Lesson: Update with instructions for updating ruby-build
**Overview**

We have recently updated our recommended Ruby version to 3.1. As part of this upgrade, we have encouraged learners to go back and update their Ruby version using the [Installing Ruby Lesson](https://www.theodinproject.com/lessons/ruby-installing-ruby).

One potential issue when attempting to update Ruby via rbenv is missing versions for Ruby build. While the fix for this is relatively simple (follow the command in the error message) we should add a small line in the lesson to make this possible step explicit. See screenshot for an example error message and fix.

:spiral_notepad:  **Acceptance Criteria**
* [ ] Instructions are updated for Linux 
* [ ] Instructions are updated for Mac (nb - I am unsure of the exact process on Mac, but possibly different)

:eyes: **Contributions Wanted**

Please respond here to be assigned this issue. It may require a bit of independent research for the Mac side. Feel free to ping me here if you need any extra guidance or clarification.

:lady_beetle: **Example Error With Fix**

error screenshot

matristle:

<hr>

**Because:**

Closes #24264

**This PR:**

* Adds an instruction for updating the `ruby-build` plugin on Linux
* Adds instructions for:
  * Installing standalone `ruby-build` on MacOS
  * Updating standalone `ruby-build` on MacOS 
  
 **References:**
  
https://github.com/rbenv/ruby-build/wiki#updating-ruby-build
https://github.com/rbenv/rbenv#homebrew


ruby/introduction/installing_ruby.md
```diff
@@ -188,7 +194,7 @@ Now you're ready to install Ruby. We're going to use a tool called `rbenv`, whic
 To install `rbenv`, run the following in your terminal:
 
 ~~~bash
-brew install rbenv
+brew install rbenv ruby-build
```
ChargrilledChook, line 197: Good idea to add this, as this is what `rbenv` recommend in their official docs. Have never installed it on a Mac so can't really speak to why it's not in there, but parts of this lesson are quite old.

ruby/introduction/installing_ruby.md
```diff
@@ -125,6 +125,12 @@ ruby 3.1.2pxx (20xx-xx-xx revision xxxxx) [x86_64-linux]
 ~~~
 where x represents the version available at the time you installed Ruby.
 
+If you've already installed `rbenv`, we'll need to update the `ruby-build` plugin. As such, let's pull in the latest Ruby version definition:
```
ChargrilledChook, line 128: I would consider softening the language a bit from `need` to `may need` or similar, because it will depend on when they installed rbenv and using which method. Not that it really hurts to do it anyway.

Looking now, I think it may be better to put this either under a `troubleshooting` or `updating a previous install` (not married to these names, just examples) and make it explicit what problem it's trying to fix. Otherwise it's an extra step without context, and could possibly be missed. Perhaps even add an example of the potential error. What do you think?

ruby/introduction/installing_ruby.md
```diff
@@ -258,6 +264,18 @@ ruby 3.1.2pxx (20xx-xx-xx revision xxxxx) [x86_64-darwin18]
 
 If you don't see the output above, log off and log back on, then try again.
 
+If you've already installed `rbenv`, we'll need to install `ruby-build`:
```
ChargrilledChook, line 267: Similar to my first comment, I would consider putting this in its own section and perhaps compressing the length of it a little bit. I understand why you've added the install step twice (given they may have missed it the first time around) but it makes this a little long. Given Mac packages will tend to be more up to date I think people running into errors here is an edge case, but I may be wrong.

ChargrilledChook (commented):
Thanks very much for making these changes @matristle, nice work on the research too. 

I think it's a pretty solid addition but have a few thoughts, let me know what you think and we can start getting these changes in :tada:

ruby/introduction/installing_ruby.md
matristle: I agree with the suggested changes. Here's the corresponding commit:

https://github.com/TheOdinProject/curriculum/pull/24293/commits/df2503be0b60e867838f094e7c85a8bd2be473d9

ruby/introduction/installing_ruby.md
matristle: I'm not sure if I've made the right decision here, but I've chosen your more verbose heading suggestion `updating a previous install` for MacOS. I think it fits better for it while sacrificing consistency with the Linux corresponding heading `Troubleshooting` which I think is a better theme for this subsection.

Here's the corresponding commit: https://github.com/TheOdinProject/curriculum/pull/24293/commits/9da8e065abc4af21242b5ddecb634133ab368d0b

zachmmeyer:
@ChargrilledChook

ruby/introduction/installing_ruby.md
```diff
@@ -106,7 +106,27 @@ rbenv install 3.1.2 --verbose
 
 This command will take 10-15 minutes to complete. The `--verbose` flag will show you what's going on so you can be sure it hasn't gotten stuck. While it installs, take this time to watch [this video](https://youtu.be/X2CYWg9-2N0) or to get a glass of water.
 
-When the last command is finished, set the Ruby version and verify that it's working:
+#### Troubleshooting
```
ChargrilledChook, line 109: One issue with this subheading is now the rest of this section is nested under it, and it seems like all the steps are about troubleshooting. I think this either needs to go to the end of the section, or another subheading needs to be added. I think my preference would be the first, but either could probably work.

An alternative would be to simply make updating `ruby-build` a part of the steps, as it's a lot less verbose overall. What do you think about these options?

ruby/introduction/installing_ruby.md
```diff
@@ -106,7 +106,27 @@ rbenv install 3.1.2 --verbose
...
+ruby-build: definition not found: x.x.x
+
+See all available versions with `rbenv install --list'.
+
+If the version you need is missing, try upgrading ruby-build:
+
+  git -C /home/itorja/.rbenv/plugins/ruby-build pull
+~~~
+
+We could try running the suggested command accordingly:
```
ChargrilledChook, line 123: This sentence is a little bit clunky. Can we reword this to be a bit more direct? (Pending decisions on the first comment)

ruby/introduction/installing_ruby.md
```diff
@@ -243,6 +262,14 @@ rbenv install 3.1.2 --verbose
 
 This command will take 10-15 minutes to complete. The `--verbose` flag will show you what's going on so you can be sure it hasn't gotten stuck. While it installs, take this time to watch [this video](https://www.youtube.com/watch?v=X2CYWg9-2N0) or to get a glass of water.
 
+#### Updating a Previous Install
```
ChargrilledChook, line 265: Same issue with the subheadings as the first. Again, perhaps the answer is to just upgrade `ruby-build` by default?

ChargrilledChook (changes requested):
Apologies it too me so long to get to this @matristle, it slipped through the cracks. 

I have a few concerns about the structure and flow of the lesson with these changes and have left a few ideas; I would love to hear your thoughts.

Thanks again for taking the time to work on this :pray:

ruby/introduction/installing_ruby.md
matristle: > I think this either needs to go to the end of the section

I like the idea of placing `#### Troubleshooting` at the end of the section, but:

> An alternative would be to simply make updating `ruby-build` a part of the steps, as it's a lot less verbose overall.

I like this a bit more because the hooks/step subheadings are giving me a sense of comfort and consistency. Maybe that extends to others as well.

I could also discard chunking it as a step and integrate it into `#### Step 2.2: Install Ruby` (without a `#### Troubleshooting` subheading or step). That's likely what you were implying as well. So I'll go with this one!

ruby/introduction/installing_ruby.md
matristle: > Again, perhaps the answer is to just upgrade ruby-build by default?

I totally agree. It seemed like the flow would be better if the process was explicit, but the way I had arranged the portions actually interrupts the flow in an unanticipated way.

matristle:
> Apologies it too me so long to get to this @matristle, it slipped through the cracks.

No problem! I got a bit concerned that I had done anything terribly wrong, but I understand. Coming back to it after a while made me aware of the mistakes I'd made, especially since they made me confused myself :smile:

ruby/introduction/installing_ruby.md
```diff
@@ -237,12 +254,21 @@ You'll notice nothing happened in the terminal. That's okay and is typical respo
 
 We can now (finally) install Ruby! Our curriculum currently uses version 3.1.2, which will allow you to complete this path's materials and content without error. We upgrade the material to accommodate newer versions as necessary. Without further ado, let's get going!
 
+First, let's upgrade `ruby-build`:
+
+~~~bash
+brew upgrade ruby-build
```
ChargrilledChook, line 260: Do you know what will happen here if the user doesn't have Ruby Build installed yet?

ruby/introduction/installing_ruby.md
```diff
@@ -237,12 +254,21 @@ You'll notice nothing happened in the terminal. That's okay and is typical respo
...
+
+Now we're ready to install our desired version of Ruby:
+
 ~~~bash
 rbenv install 3.1.2 --verbose
 ~~~
 
 This command will take 10-15 minutes to complete. The `--verbose` flag will show you what's going on so you can be sure it hasn't gotten stuck. While it installs, take this time to watch [this video](https://www.youtube.com/watch?v=X2CYWg9-2N0) or to get a glass of water.
 
+
```
ChargrilledChook, line 271: [Nit] Looks like you have an extra line of white space here

ChargrilledChook (commented):
@matristle Looking good, I have one tiny change request and a question, if you could get back to me about those then we can get this merged :muscle:

ruby/introduction/installing_ruby.md
matristle: I had brushed that possibility off as the user simply starting from Step 3.1 which includes the installation of `ruby-build`. I also notice that there is a sudden inconsistent jump from Step 3.1 to 3.3 without a Step 3.2, so we could add the `ruby-build` installation step there such that it's less miss-able as its own step?

Following that train of thought, I think it would be better to make that Step 3.1 and make the `rbenv` installation Step 3.2 and so that it's also not easily dismissable.

ChargrilledChook: Sounds good, give it a go and let's see how it looks :+1:

zachmmeyer:
@matristle If you believe you have resolved the conversations with @ChargrilledChook could you please hit the "Resolve conversation" buttons and re-request a review?

matristle:
@zachmmeyer Thanks for the tip!

ChargrilledChook (approved):
Nice work @matristle, appreciate your efforts  🚀

PR #24312, opened 2022-06-29
"CSS Foundations" first example of specificity

androidmaziou:
In the first example for specificity, in the "CSS Foundations" page, changed .subsection to .list, so the point on precedence makes more sense. As .subsection, there is no conflict, so precedence simply doesn't apply, even if Rule 2 is more specific. Changing .subsection to .list, will actually show how the more-specific rule takes precedence in what would otherwise be a conflict.

<hr>

**1. Because:**

Closes #XXXXX

**2. This PR:**

### DIFF
--- a/foundations/html_css/css-foundations.md
+++ b/foundations/html_css/css-foundations.md
@@ -286,7 +286,7 @@ Let's take a look at a few quick examples to visualize how specificity works.
 
 ~~~css
 /* rule 1 */
-.subsection {
+.list {
   color: blue;
 }
 


fortypercenttitanium (changes requested):
@androidmaziou I think you've uncovered a shortcoming of this lesson, but I think we can do even better than just switching the class names. Could we perhaps include a brief HTML layout to show that we are talking about the same element in both CSS declarations? Similar to how we use HTML examples on line 191.

fortypercenttitanium:
Closed because #24315 makes a better fix on this verbiage. Thank you for your contribution!

PR #24323, opened 2022-07-01
Installations: Directly mention Desktop version in the instructions for dual boot

scheals:

<hr>

**1. Because:**
Sometimes users download the wrong image and proceed with the installation till the very end even though it doesn't fit our instructions nor the linked Canonical's instruction. 

**2. This PR:**
I've added the word "Desktop" in bold in the instructions to make sure this doesn't occur.

### DIFF
--- a/foundations/installations/installations.md
+++ b/foundations/installations/installations.md
@@ -201,7 +201,7 @@ Dual-booting provides two operating systems on your computer that you can switch
 
 ### Step 1: Download Ubuntu
 
-First, you need to download the version of Ubuntu you want to install on your computer. Ubuntu comes in different versions ("flavors"), but we suggest the standard desktop [Ubuntu](https://releases.ubuntu.com/20.04/). If you're using an older computer, we recommend [Xubuntu](https://xubuntu.org/release/20-04/). Be sure to download the 64-bit version of [Ubuntu](https://releases.ubuntu.com/20.04/) or [Xubuntu](https://xubuntu.org/release/20-04/).
+First, you need to download the version of Ubuntu you want to install on your computer. Ubuntu comes in different versions ("flavors"), but we suggest the standard desktop [Ubuntu](https://releases.ubuntu.com/20.04/). If you're using an older computer, we recommend [Xubuntu](https://xubuntu.org/release/20-04/). Be sure to download the 64-bit **Desktop** version of [Ubuntu](https://releases.ubuntu.com/20.04/) or [Xubuntu](https://xubuntu.org/release/20-04/).
 
 ### Step 2: Create a Bootable Flash Drive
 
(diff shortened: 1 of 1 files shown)


foundations/installations/installations.md
```diff
@@ -201,7 +201,7 @@ Dual-booting provides two operating systems on your computer that you can switch
 
 ### Step 1: Download Ubuntu
 
-First, you need to download the version of Ubuntu you want to install on your computer. Ubuntu comes in different versions ("flavors"), but we suggest the standard desktop [Ubuntu](https://releases.ubuntu.com/20.04/). If you're using an older computer, we recommend [Xubuntu](https://xubuntu.org/release/20-04/). Be sure to download the 64-bit version of [Ubuntu](https://releases.ubuntu.com/20.04/) or [Xubuntu](https://xubuntu.org/release/20-04/).
+First, you need to download the version of Ubuntu you want to install on your computer. Ubuntu comes in different versions ("flavors"), but we suggest the standard desktop [Ubuntu](https://releases.ubuntu.com/20.04/). If you're using an older computer, we recommend [Xubuntu](https://xubuntu.org/release/20-04/). Be sure to download the 64-bit **Desktop** version of [Ubuntu](https://releases.ubuntu.com/20.04/) or [Xubuntu](https://xubuntu.org/release/20-04/).
```
thatblindgeye, line 204: I think my worry here is that users will still download the wrong thing. We've seen several attempts in other lessons where maintainers have tried to emphasize some text to make something clearer for users, but users can still be prone to "do" before they finish reading.

A couple of alternative options I can think of:

1. Before we even link to any of the external sites to download anything, we clarify which version they should install first. This can be done one of several ways, like `download the 64 bit desktop version of [Ubuntu](...)`
2. Link directly to the download files themselves to avoid the issue completely. This would require letting users know that what they are about to click on is a download link and include the file type and file size. We would also need to mention the prereqs mentioned on those sites, like RAM and whatnot.

scheals: So for 1. we'd remove the previous links to Ubuntu and Xubuntu and only leave the ones after "Be sure" like:

> First, you need to download the version of Ubuntu you want to install on your computer. Ubuntu comes in different versions ("flavors"), but we suggest the standard desktop Ubuntu. If you're using an older computer, we recommend Xubuntu. Be sure to download the 64-bit **Desktop** version of [Ubuntu](https://releases.ubuntu.com/20.04/) or [Xubuntu](https://xubuntu.org/release/20-04/).

Correct?

For 2. in the VM part of the instructions curriculum links directly to the US mirror of Xubuntu .iso, I see that Ubuntu's page has single download link so maybe server's localization is less of a concern here. Minimum requirements are 2GB RAM for Ubuntu and 1GB for Xubuntu respectively, so this could be done instead indeed.

thatblindgeye: Yep! With rewording like that I think option 1 would be better than linking directly to the download for now.

thatblindgeye (approved):
🎉 Looks good!

PR #24367, opened 2022-07-05
Ruby Project: Linked Lists: Fix incorrect instruction wording

romangdev:

-   [x ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)

<hr>

**1. Because:**

This PR addresses a faulty suggestion in the instructions for the Linked List project, leading to confusion through its wording. In step 2 of the linked list project, the instructions outline that the student should create a Node class "containing a #value method and a link to the #next_node, set both as 'nil' by default."

The instructions should instead prompt the user to create a Node class with "@value" and "@next_node" as instance variables set to nil upon initialization, rather than methods. Because for this project, when you are instantiating a new node  class, you are using those instance variables to store its value and point to another node, not creating methods for those specific actions. 

**2. This PR:**

- Fixed wording by clarifying "value" and "next_node" should be set as instance variables. 
- Removed mention of setting the aforementioned as methods.

### DIFF
--- a/ruby/computer_science/project_linked_lists.md
+++ b/ruby/computer_science/project_linked_lists.md
@@ -40,7 +40,7 @@ For a more thorough explanation, use these resources:
   You will need two classes:
 
 1. `LinkedList` class, which will represent the full list.
-2. `Node` class, containing a `#value` method and a link to the `#next_node`, set both as `nil` by default.
+2. `Node` class, containing a `@value` instance variable and a `@next_node` instance variable linking to the next node, set both as `nil` by default.
 
 Build the following methods in your linked list class:
 


ChargrilledChook (changes requested):
Hi @romangdev, thanks heaps for taking the time to make a suggestion. 

In this case, I think there's a bit of nuance here that means we can't add this change in as is. In order for the linked list to work correctly, the Node class _does_ need these methods. Specifically it probably needs an [attribute reader](http://ruby-for-beginners.rubymonstas.org/writing_classes/attribute_readers.html) or accessor to wrap the instance variable:
```rb
attr_reader :value

# is equivalent to 

def value
  @value
end
```
That seems to me what the lesson is hinting at when talking about using methods here.
Without these, `Node` instances don't really have a way to communicate with each other.

Let me know your thoughts or if I've missed anything. If you think this could be clarified, I'm happy for you to make some changes, but I would probably try to keep it fairly light.

ChargrilledChook:
> Thanks a lot for taking the time to respond @[ChargrilledChook](https://github.com/ChargrilledChook). Apologies if there's any confusion on my end... only been at this for a couple of months.

Hey @romangdev, no worries at all. 

> I did not have nor need these methods for my Linked List to work properly... likewise for my peers. Is there something I'm missing here?

When you use `attr_reader` or `attr_writer` etc you are essentially using a shortcut to define some simple methods. Consider this example:

```rb
class NodeOne
  attr_accessor :value

  def initialize(value = nil)
    @value = value
  end
end

class NodeTwo
  def initialize(value = nil)
    @value = value
  end

  def value
    @value
  end

  def value=(new_value)
    @value = new_value
  end
end

puts "Node one's methods are #{NodeOne.new(1).public_methods(false)}"
puts "Node two's methods are #{NodeTwo.new(2).public_methods(false)}"

#=> Node one's methods are [:value, :value=]
#=> Node two's methods are [:value, :value=]
```

Even though `NodeOne` has not  explicitly declared a `value` or `value=` method, that is the end result of using `attr_accessor :value`. 

Does that make sense?

romangdev:
Ahh... totally slipped my mind that `attr_accessor` literally sets the getter and setter methods, and I missed that that was what you were getting at in your initial reply. Been using it for a while but forgot what it did at its core. Thanks so much for your patience and thorough refresher.... I really appreciate it.

ChargrilledChook:
@romangdev No worries, my pleasure. Thanks for taking the time to make a suggestion!

Issue #24177, opened 2022-05-30
Bug - Flexbox Growing and Shrinking lesson needs a rewrite 


aziham:
Complete the following REQUIRED checkboxes:
-   [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [x] The title of this issue follows the `Bug - location of bug: brief description of bug` format, e.g. `Bug - React section: Knowledge Checks don't link to resource`

The following checkbox is OPTIONAL:

-   [x] I would like to be assigned this issue to work on it

<hr>

**EDIT - Maintainer Team - Daniel** 

**Acceptance Criteria:**

- [X] Replace current screenshot at start of lesson with an example of `flex: 1` and a description of what values that equates to
- [X] Add a new section that explains `flex: auto` and the values that it equates to
- [X] Add a new Assignment section (following the Layout Style Guide) that includes resources that could cover flex shorthand, flex-grow, flex-shrink and flex-basis. See some possible options here:
    - https://www.w3.org/TR/css-flexbox-1/#flex-common
    - https://css-tricks.com/almanac/properties/f/flex/
    - https://developer.mozilla.org/en-US/docs/Web/CSS/flex
- [ ] Add at least 1 more Knowledge Check to help cover the new lesson content
------------

**1. Description of the Bug:**

According to https://www.w3.org/TR/css-flexbox-1/#flex-common

>[flex: initial](https://www.w3.org/TR/css-flexbox-1/#propdef-flex)
[...]

PR #24455, opened 2022-07-25
Foundations flexbox growing and shrinking lesson: Flex auto section creation 

aaronlrv:

<hr>

**1. Because:**

**2. This PR:**

Linked to issue: #24177 Specifically told not to close
Link : https://github.com/TheOdinProject/curriculum/issues/24177

### DIFF
--- a/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
+++ b/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
@@ -58,7 +58,7 @@ An important implication to notice here is that when you specify `flex-grow` or 
 
 > #### Important note about flex-basis:
 >
-> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`.
+> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand. Or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`, which is also equivalent to using `flex: auto`.
 
 > #### What is flex: auto?
 >
--- a/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
+++ b/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
@@ -56,13 +56,13 @@ An important implication to notice here is that when you specify `flex-grow` or 
 
 `flex-basis` simply sets the initial size of a flex item, so any sort of `flex-grow`ing or `flex-shrink`ing starts from that baseline size. The shorthand value defaults to `flex-basis: 0%`. The reason we had to change it to `auto` in the `flex-shrink` example is that with the basis set to `0`, those items would ignore the item's width, and everything would shrink evenly. Using `auto` as a flex-basis tells the item to check for a width declaration (`width: 250px`).
 
-### flex-auto
-
-`flex: auto` is one of the short hands of flex. When `flex: auto` is called it inherits the values of `flex-grow: 1 flex-shrink: 1 flex:basis: auto` or `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first.
-
 > #### Important note about flex-basis:
 >
 > There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`. You could also use the predefined value `flex: auto`, which is equivalent to `flex: 1 1 auto`
+
+#### What is flex: auto?
+
+`flex: auto` is one of the short hands of flex. When `flex: auto` is called it inherits the values of `flex-grow: 1 flex-shrink: 1 flex:basis: auto` or `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first.
 
 ### In practice...
 
--- a/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
+++ b/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
@@ -58,7 +58,7 @@ An important implication to notice here is that when you specify `flex-grow` or 
 
 > #### Important note about flex-basis:
 >
-> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`. You could also use the predefined value `flex: auto`, which is equivalent to `flex: 1 1 auto`
+> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: auto`.
 
 #### What is flex: auto?
 
--- a/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
+++ b/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
@@ -22,6 +22,8 @@ In this case, `flex` is actually a shorthand for `flex-grow`, `flex-shrink` and 
 
 The default value of the `flex` property is shown in the above screenshot: `flex-grow: 0`, `flex-shrink: 1`, `flex-basis: 0%`. Very often you see the flex shorthand defined with only _one_ value. In that case, that value is applied to `flex-grow`, so when we put `flex: 1` on our divs, we were actually specifying `flex: 1 1 0%`.
 
+### flex-auto
+
 ### flex-grow
 
 `flex-grow` expects a single number as its value, and that number is used as the flex-item's "growth factor". When we applied `flex: 1` to every div inside our container, we were telling every div to grow the same amount. The result of this is that every div ends up the exact same size. If we instead add `flex: 2` to just one of the divs, then that div would grow to 2x the size of the others.
--- a/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
+++ b/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
@@ -62,7 +62,7 @@ An important implication to notice here is that when you specify `flex-grow` or 
 
 #### What is flex: auto?
 
-`flex: auto` is one of the short hands of flex. When `flex: auto` is called it inherits the values of `flex-grow: 1 flex-shrink: 1 flex:basis: auto` or `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first.
+If you noticed, we mentioned a new flex shorthand `flex: auto` in the previous note however we didn't fully introduce it. `flex: auto` is one of the short hands of flex. When `flex: auto` is called it inherits the values of `flex-grow: 1 flex-shrink: 1 flex:basis: auto` or `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first.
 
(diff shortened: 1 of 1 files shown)


aaronlrv:
Hey! @dm-murphy I'm honestly a bit unsure about the placement of where the flex-auto content is at the moment, I wanted to introduce it before the note on flex basis where flex-auto is being mentioned however I'm still 50/50

I'd love to know your thoughts and opinions on where we should place the flex-auto section

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
```diff
@@ -56,6 +56,10 @@ An important implication to notice here is that when you specify `flex-grow` or
 
 `flex-basis` simply sets the initial size of a flex item, so any sort of `flex-grow`ing or `flex-shrink`ing starts from that baseline size. The shorthand value defaults to `flex-basis: 0%`. The reason we had to change it to `auto` in the `flex-shrink` example is that with the basis set to `0`, those items would ignore the item's width, and everything would shrink evenly. Using `auto` as a flex-basis tells the item to check for a width declaration (`width: 250px`).
 
+### flex-auto
```
dm-murphy, line 59: **issue:** Just noting that it's actually `flex: auto`. The property is `flex` and the value for the property is `auto`. 

**suggestion:** Perhaps a good title for this section could be something like: What is flex: auto?

That might help distinguish it from the actual properties that are previously discussed as sections (`flex-grow`, `flex-shrink`, `flex-basis`).

To your comment about where to place this section, I believe we want it after the flex-basis section and I also think after the "Important note about flex-basis". 

To that end, I think we could either make this new section another note instead of a sub-header. So we could have the flex-basis section, then the note "Important note about flex-basis" and then another note like "What is flex: auto?" Or we could make it into its own sub-header section. We just want to make sure it is distinguished as a separate thing from the previous section properties. 

We could also take a look at the "Important note about flex-basis" note and probably remove the explanation on the equivalent value of `flex: auto` being `flex: 1 1 auto` since that would be covered in this new section. We could even signify that the learner will learn more about this in the next part with a rewrite of that final line like this: "You could also use the predefined value `flex: auto`, which you will see next."

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
```diff
@@ -56,6 +56,10 @@ An important implication to notice here is that when you specify `flex-grow` or
...
+
+`flex: auto` is one of the short hands of flex. When `flex: auto` is called it inherits the values of `flex-grow: 1 flex-shrink: 1 flex:basis: auto` or `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first. The default value for flex is `flex: 0 1 auto` or `flex: initial` which we won't get too in depth about here.
```
dm-murphy, line 61: **praise:** I like the note on `flex: auto` not being the default value for flex short hand despite its name as "auto". Good point!

**suggestion:** I think we should drop the final line on "The default value for flex..." There really is no default value of the shorthand property `flex` itself because you have to specify a keyword value for the property. I think it's better to just drop the line and not try to even get into `flex: initial` since the focus on this section is `flex: auto`.

dm-murphy, line 61: **issue (typo):** `flex:basis: auto` -> `flex-basis: auto`

**nitpick:** Some of the language here could be tweaked just a bit. "Called" sort of implies a method/function so perhaps we should use something like "specified". I'm also not sure that "inherits" is the right word choice here either. 

Maybe something like: "When `flex: auto` is specified it is equivalent to using the `flex-grow: 1` `flex-shrink: 1` and `flex-basis: auto` properties or using the flex shorthand `flex: 1 1 auto`."

dm-murphy, line 61: **thought:** What do you think about including one more line in the section that could summarize the standard use case of `flex: auto`?

If you look at the docs from w3 or MDN they both give a brief summary of how the `flex: auto` property will size the item by its `width` and `height` properties but still flex to the free space of the container. I think we could either add our own line here that doesn't directly copy w3/MDN, or use an exact quote and give source attribution to one of those docs. 

Let me know your thoughts on adding something like this to the section. Or if you think that will be too much here, we could also tell the learner something like "you will learn more about when to use `flex: auto` when reading through the docs in the Assignment section."

dm-murphy (changes requested):
Thanks for getting started on this @aaronlrv! 🙂 

I've added some comments/suggestions in the review and addressed your initial comment as well. Let me know your thoughts whenever you get the chance!

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Got it! Will do

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Hey! My bad for seeing this so late, I don't recieve the github notifications for odd reasons. Let me know your thoughts on the current phrasing in the recent push and hopefully I cleared the confusion on that one. If it's a bit wordy I still have the remaining drafts that we could consider 


- When `flex: auto` is defined it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto (your option)
- When `flex: auto` is specified it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto` or `flex: 1 1 auto`
- When `flex: auto` is defined as a flex parameter it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Hey! My bad for seeing this so late, I don't get the github notifications for some odd reason. Let me know your thoughts on the wording within the recent push and hopefully I cleared the confusion there, I  have these drafts that I made that we can come back too if it isn't as clear as we want it to be yet


- When `flex: auto` is defined it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto (your option)
- When `flex: auto` is specified it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto` or `flex: 1 1 auto
- When `flex: auto` is defined as a flex parameter it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Hey do let me know if the placement is good around here, or if we'd rather it be a subheading on its own. I think it's great where its at personally

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Going to be working on this tommorow! I will re-request a review upon finishing

aaronlrv: Hey, Upon looking at this i'm not sure if this aligns with our purpose of informing about flex-auto. If we leave the note header as  "What is flex-auto" as a continuation explanation for the flex-basis article, Our goal would be to clarify about the definition of flex-auto and what it is in itself rather than how to use it and it's potential use cases.

I've added a line equivalent to the  

> you will learn more about when to use flex: auto when reading through the docs in the Assignment section."

Described earlier, Let me know your viewpoint on this.

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
```diff
@@ -58,7 +58,11 @@ An important implication to notice here is that when you specify `flex-grow` or
 
 > #### Important note about flex-basis:
 >
-> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`. You could also use the predefined value `flex: auto`, which is equivalent to `flex: 1 1 auto`
+> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: auto`.
```
dm-murphy, line 61:
```suggestion
> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`.
```
**issue:** The original `flex: 1 1 auto` is the full 3 value shorthand here. Did you mean to add a new line describing `flex: auto`? Or did you accidentally delete the "You could also use the predefined value `flex: auto`" line? Let me know when you get the chance! :-)

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
```diff
@@ -58,7 +58,11 @@ An important implication to notice here is that when you specify `flex-grow` or
...
+
+> #### What is flex: auto?
+>
+> If you noticed, we mentioned a new flex shorthand `flex: auto` in the previous note however we didn't fully introduce it. `flex: auto` is one of the short hands of flex. When `auto` is defined as a flex parameter it is equivalent to the values of `flex-grow: 1 flex-shrink: 1 flex-basis: auto` or `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first. You will encounter and learn more about flex: auto and it's potential use-cases when reading through the assignment section.
```
dm-murphy, line 65:
```suggestion
> If you noticed, we mentioned a new flex shorthand `flex: auto` in the previous note. However we didn't fully introduce it. `flex: auto` is one of the short hands of flex. When `auto` is defined as a flex keyword it is equivalent to the values of `flex-grow: 1`, `flex-shrink: 1` and `flex-basis: auto` or to `flex: 1 1 auto` using the flex shorthand. Note that `flex: auto` is not the default value when using the flex short hand despite the name being "auto" which may be slightly confusing at first. You will encounter and learn more about `flex: auto` and its potential use-cases when reading through the assignment section.
```
**suggestion:** Looks great overall! I just added some grammatical fixes here and separated the properties a bit more to try and imply these would be separate lines. I also changed "parameter" to "keyword" here which is the proper term for a CSS property value. 

I like the idea of telling learners they will find out more in the assignments without overwhelming them here. 👍

dm-murphy (changes requested):
Thanks @aaronlrv for going through all these iterations! 🙂 

I added in a couple suggestions you can commit from the comment box if you agree. Also a question left for you about the flex-basis note's final line. 

No rush and thanks for all your work here! I know you mentioned not receiving GitHub notifications so if that's something you would like to set up maybe check out some of the docs here: https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Oh my bad I actually did end up deleting it by accident, I've committed this change but in any event you want it back let me know!

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: Commited!

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
```diff
@@ -58,7 +58,11 @@ An important implication to notice here is that when you specify `flex-grow` or
 
 > #### Important note about flex-basis:
 >
-> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`. You could also use the predefined value `flex: auto`, which is equivalent to `flex: 1 1 auto`
+> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand, or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`.
```
dm-murphy, line 61:
```suggestion
> There is a difference between the default value of `flex-basis` and the way the `flex` shorthand defines it if no `flex-basis` is given. The actual default value for `flex-basis` is `auto`, but when you specify `flex: 1` on an element, it interprets that as `flex: 1 1 0`. If you want to _only_ adjust an item's `flex-grow` you can simply do so directly, without the shorthand. Or you can be more verbose and use the full 3 value shorthand `flex: 1 1 auto`, which is also equivalent to using `flex: auto`.
```
**suggestion:** I think we do want to mention `flex: auto` here somewhere since the next section's note says "If you noticed, we mentioned a new flex shorthand `flex: auto` in the previous note". I also tried to break up the sentences here a little more. Let me know if you think this works or if it sounds too confusing.

dm-murphy (commented):
@aaronlrv One more note/question for you! 🙂

foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
aaronlrv: No this works absolutely perfectly, Committed!

dm-murphy (approved):
Thanks @aaronlrv! Great work all around on this whole flex issue 🙂

PR #24458, opened 2022-07-26
Workflows: Add Labeler GitHub Action for automatically adding certain labels

01zulfi:

-   [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)

<hr>

**1. Because:**

Automation

**2. This PR:**

Adds [Labeler GitHub Action](https://github.com/actions/labeler)

n/a

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,32 @@
+'Status: Needs Review':
+  - '**/*'
+
+'Course: Advanced HTML/CSS':
+  - 'advanced_html_css/**/*'
+
+'Course: Databases':
+  - 'databases/**/*'
+
+'Course: Getting Hired':
+  - 'getting_hired/**/*'
+
+'Course: Intermediate HTML/CSS':
+  - 'html_css/**/*'
+
+'Course: JavaScript':
+  - 'javascript/**/*'
+
+'Course: NodeJS':
+  - 'nodeJS/**/*'
+
+'Course: Ruby':
+  - 'ruby/**/*'
+
+'Course: Ruby on Rails':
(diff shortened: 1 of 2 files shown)


ChargrilledChook (approved):
Great idea!

I would like to consider splitting up the `needs review` label because currently it means both code review and initial triage. I think splitting these into two different labels would help increase clarity, but this is somewhat adjacent to your changes here.

We might also want to think about changing `Type: Foundations` to use the course system since it sticks out a little bit now. Perhaps with this granularity the path types are redundant.

Love this change :heart:

01zulfi:
@ChargrilledChook 
How do you propose we split the needs review label? Maybe having a triage label could help?

Sure, we can go for a `Course: Foundations` label instead of the type one. Although we have to first specify the structure of Foundations. Is it a path or a course? It's technically a course but sure feels like a path.

ChargrilledChook:
@01zulfi 
> How do you propose we split the needs review label? Maybe having a triage label could help?
Yep exactly, my bad for not being more clear on this. I think that's the ideal solution.

> Sure, we can go for a Course: Foundations label instead of the type one. Although we have to first specify the structure of Foundations. Is it a path or a course? It's technically a course but sure feels like a path.
I guess it's a bit unique / weird since it is a path that has one course, and they're both called Foundations. 

Not a blocker by any means, more a thought. Revisiting / changing the name in the future is unlikely to be hard if we decide to go that route

ChargrilledChook (approved):

PR #24461, opened 2022-07-26
Cs ruby removal

CouchofTomato:

<hr>

**1. Because:**
The CS course has some ruby specific content which makes it unsuitable for sharing with the JS course.

**2. This PR:**

* remove ruby specific content from CS intro lesson
* remove ruby specific content from binary search tree project
* remove ruby specific content from linked list project
* remove ruby specific content from recursion project
* remove ruby specific content from recursion lesson

### DIFF
--- a/ruby/computer_science/a_very_brief_intro_to_cs.md
+++ b/ruby/computer_science/a_very_brief_intro_to_cs.md
@@ -2,7 +2,7 @@
 
 You've learned how to build some cool stuff already and, frankly, you could probably make some decent websites without a crazy amount of additional formal education.  However, just because you can write English at a grade school level doesn't mean you will be editing the New York Times anytime soon.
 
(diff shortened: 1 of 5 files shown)


LeoKirasic (changes requested):
Seems pretty good to me. 

I think we should do something about function/method names though. 

For example `Write a #build_tree method` doesn't follow the JS naming convention. 

I'm guessing that the `#` character is used for implying that something is a method in ruby. However, that's not the case for JS. We should find a workaround for that.

dm-murphy:
> Seems pretty good to me.
> 
> I think we should do something about function/method names though.
> 
> For example `Write a #build_tree method` doesn't follow the JS naming convention.
> 
> I'm guessing that the `#` character is used for implying that something is a method in ruby. However, that's not the case for JS. We should find a workaround for that.

That is the way of describing a Ruby method. There might be a way of working around this method name issue in this particular example that doesn't confuse learners of either path, but I also worry about future instances where we add to these lessons and need to distinguish between Ruby and JavaScript.

Might end up needing to have separate courses here :-(

PR #24500, opened 2022-08-05
project_linked_lists.md: Fix typo

Roxanoel:
Assignment point 8 reads "returns....nil" instead of (presumably) "returns....null".

<hr>

**1. Because:**

In point 8 of the assignment, it is written that find(value) must return nil if the value is not found; presumably, this should read "null". This change also ensures consistency with the rest of the instructions. 

**2. This PR:**

- In assignment, step 8, "nil" was changed to "null". 

Nothing else to add :)


KevinMulhern (approved):
Thanks @Roxanoel

PR #24540, opened 2022-08-12
JavaScript: Binary Search Tree Project: add `prettyPrint()` function translated from Ruby's BST project

JoshDevHub:

<hr>

**1. Because:**

 This section was recently translated from the computer science section in the ruby path. For the binary search tree project, an important component to the assignment was lost in the translation: the `#pretty_print` method. You can see it at the bottom of the [assignment in the ruby section here](https://www.theodinproject.com/lessons/ruby-binary-search-trees). This method allows learners to print their tree in a human reader friendly way, which is especially valuable given how this section comes prior to testing -- when learners will only be able to test by looking at console output. Here are two terminal screenshot comparisons to show how much more readable this output is with `prettyPrint()`:
 
 **Using `console.dir` to print a nested object. The entire tree given at the end of the assignment doesn't even fit in my full terminal window ->**
 Screenshot from 2022-08-12 16-38-23
 
 **Using this `prettyPrint()` function on a tree of equal size ->**
Screenshot from 2022-08-12 16-39-52

**2. This PR:**

 * Translates the `prettyPrint()` function present in the ruby course over to JavaScript. I've verified it works in my [own binary search tree project in JS](https://github.com/JoshDevHub/JavaScript-Binary-Search-Tree/blob/main/binary-tree.js#L129-L137).

I've positioned this at the very end of the exercise, just like it is in the Ruby course. Glad to make a further commit if it's considered better elsewhere on the page.

### DIFF
--- a/javascript/computer_science/project_binary_search_trees.md
+++ b/javascript/computer_science/project_binary_search_trees.md
@@ -45,7 +45,7 @@ Write a simple driver script that does the following:
 
 </div>
 
-**Tip:** If you would like to visualize your binary search tree, here is a `prettyPrint()` function that will `console.log` your tree in a structured format. This function will expect to receive the root of your tree as an argument.
+**Tip:** If you would like to visualize your binary search tree, here is a `prettyPrint()` function that will `console.log` your tree in a structured format. This function will expect to receive the root of your tree as the value for the `node` parameter.
 
 ~~~javascript
 const prettyPrint = (node, prefix = '', isLeft = true) => {
--- a/javascript/computer_science/project_binary_search_trees.md
+++ b/javascript/computer_science/project_binary_search_trees.md
@@ -9,27 +9,41 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
 <div class="lesson-content__panel" markdown="1">
 
-1. Build a `Node` class / factory.  It should have an attribute for the data it stores as well as its left and right children.
+1.  Build a `Node` class / factory.  It should have an attribute for the data it stores as well as its left and right children.
 
-1. Build a `Tree` class / factory which accepts an array when initialized. The `Tree` class should have a `root` attribute which uses the return value of `buildTree` which you'll write next.
+1.  Build a `Tree` class / factory which accepts an array when initialized. The `Tree` class should have a `root` attribute which uses the return value of `buildTree` which you'll write next.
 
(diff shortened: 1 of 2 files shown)


javascript/computer_science/project_binary_search_trees.md
```diff
@@ -44,3 +44,17 @@ Write a simple driver script that does the following:
 1. Print out all elements in level, pre, post, and in order
 
 </div>
+
+**Tip:** If you would like to visualize your binary search tree, here is a `prettyPrint()` function that will `console.log` your tree in a structured format. This function will expect to receive the root of your tree as an argument.
```
01zulfi, line 48: Wording suggestion: Just being extra specific on how to pass in the root.

```suggestion
**Tip:** If you would like to visualize your binary search tree, here is a `prettyPrint()` function that will `console.log` your tree in a structured format. This function will expect to receive the root of your tree as the value for the `node` parameter.
```

01zulfi (changes requested):
This is an amazing and a valuable addition :fire: 

As for the placement, I personally feel the pretty print function should be placed further up in the lesson for more visibility (even in the Ruby version of the lesson). What do you think?

JoshDevHub:
@01zulfi I've committed that change of phrase for the parameter. I also think it's probably a good idea to move it higher on the page. I seem to remember at least one time where a learner in the ruby path made it through most of the project before even noticing it existed haha.

Could it maybe fit in after the `buildTree()` function? Learners are guaranteed to see it, and they'll be able to use it straight away. Let me know what you think or if you have some other idea for where it could go. I'll gladly commit those changes for both courses :+1:

01zulfi:
> I seem to remember at least one time where a learner in the ruby path made it through most of the project before even noticing it existed haha.

Oh wow, I can definitely see that happening :laughing: 


Yes, right after `buildTree` sounds good.

JoshDevHub:
@01zulfi Alright I moved the function up after `buildTree` for both paths. Sorry about the messy diff, I had to change the spacing after the numbers in the list to make sure the codeblock doesn't interrupt the ordering of the list. The [style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#nested-list-spacing) seems to recommend two spaces after numbers in an ordered list anyway.

01zulfi (approved):
This looks great :fire: Thanks for your work on this! :rocket:

PR #24544, opened 2022-08-14
Update motivation_and_mindset.md

ADL1337:
I believe that comparing yourself to your past self is not the best way to achieve a growth oriented mindset or to manifest motivation, in fact it can lead to depressive thoughts. I believe that an optimistic outlook on future learning without any comparison is more helpful.


rlmoser99:
Thank you for taking time to make a PR, however we do not believe this change is a solution to the problem people run into when comparing themselves to others. I am sorry that these words are not motivating to you, but we have talked with many people over the years, who have found this exact advice to be very helpful and encouraging.

In addition, people do not need to figure out the next steps when they are following our curriculum. The industry professionals who wrote and maintain this curriculum have taken out this guess work.

PR #24546, opened 2022-08-15
Common Data Structures And Algorithms lesson: Added an Item in Additional Resource Section

szmazhr:

<hr>

**1. Because:**

This seems great and latest addition in Additional Resources section of the lesson. 

**2. This PR:**

* Added a playlist of JavaScript Algorithms and Data Structures from Codevolution on YouTube, this seems great and latest addition in Additional Resources section of the Common Data Structures And Algorithms lesson.

### DIFF
--- a/javascript/computer_science/common_data_structures_algorithms.md
+++ b/javascript/computer_science/common_data_structures_algorithms.md
@@ -44,7 +44,7 @@ Look through these now and then use them to test yourself after doing the assign
 
 ### Additional Resources
 This section contains helpful links to other content. It isn't required, so consider it supplemental.
-
+* [This video series from Codevolution](https://youtube.com/playlist?list=PLC3y8-rFHvwjPxNAKvZpdnsr41E0fCMMP) really provides it all. Watch it for a greater understanding.
 * [Khan Academy's great Algorithms Course](https://www.khanacademy.org/computing/computer-science/algorithms)
 * [Stanford's Coursera 4-Part Algorithm Course](https://www.coursera.org/specializations/algorithms)
 * [Visualizing Algorithms from Mike Bostock](http://bost.ocks.org/mike/algorithms/)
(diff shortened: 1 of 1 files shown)


wise-king-sullyman:
This resource looks interesting, thank you for PRing it and fixing the order of knowledge checks and additional resources!

We'll need some more time to be able to review this as there's quite a lot of material here, do you have any particular recommendations from the series? If so it might be better to call those out or link directly to them rather than linking to the whole playlist.

szmazhr:
I found the whole playlist useful, and I personally watched all videos there in, It is little longer but great resource.

zachmmeyer:
OK. I don't think this playlist is bad. What's stopping me from immediately approving this PR is two things:

1. There are a total of seven additional resources for this project
2. Three of them are courses

Just under half of the addition resource content for this lesson are full length courses. What you are suggesting is adding a fourth course on top of the three, in addition to the seven total resources.

The ideal amount of additional resources has historically been a number between four and six, and typically not so many full length courses.

If I were to ask you to not only put your course at the bottom of the list, I would have to ask you to substitute out another resource and tell us why.

I brought up a brief idea on how to help this in future, and it's a rather large undertaking, but until then it's going to have to be something that you, the contributor, has to figure out.

It's up to you if you want to make those changes.

wise-king-sullyman (changes requested):
@szmazhr after reviewing the playlist I actually would suggest that we place it at the end of the knights travails project since several of the data structures it explains are projects in the CS section.

It should be done in a manner similar to what we see at the end of [the todo list project](https://www.theodinproject.com/lessons/node-path-javascript-todo-list#want-more-practice).

szmazhr:
> @szmazhr after reviewing the playlist I actually would suggest that we place it at the end of the knights travails project since several of the data structures it explains are projects in the CS section.
> 
> It should be done in a manner similar to what we see at the end of [the todo list project](https://www.theodinproject.com/lessons/node-path-javascript-todo-list#want-more-practice).

oh yes, that will be better. I will try to make required changes soon.

zachmmeyer:
Closing due to staleness. If you'd like to work more on this PR, we'd be more than happy to reopen it.

PR #24560, opened 2022-08-19
Objects and Object Constructors: Add explanation to distinguish b/w Constructor.prototype vs obj.__proto__

anupamvamsi:

<hr>

**1. Because:**

The primary aim of this PR was to update the lesson's **Recommended Method for Prototypal Inheritance** content from using `Object.create()` to the `Object.setPrototypeOf()` method.

However I realized that the lesson also would need to better explain the `prototype` function property vs. the `__proto__` /  `[[Prototype]]` property and prototypal inheritance in this lesson, as personally to me as a learner, it was confusing and I had to refer to quite a few resources to understand it.

This is the first PR in an upcoming set of PRs to make the review process easier as suggested in the Discord forum post.

**2. This PR:**

* Explains the slight differences between the `Object.prototype` and the `obj.__proto__` concepts
* It also identifies that the `__proto__` way of referring to the `.prototype` property is now done using `Object.getPrototypeOf()` method.
* Provides examples to illustrate that `obj.__proto__` (or) `Object.getPrototypeOf(obj)` refers to `Object.prototype`.
* Recognizes that `__proto__` is also referred to as `[[Prototype]]` in the specification for clarity.
* Puts forth the idea that `__proto__` can have a `null` value and will be explained further in the lesson, and the explanation will be brought in through further edits (and PRs).

* Suggestions-Bugs Forum discussion on Discord: https://discord.com/channels/505093832157691914/1010249197544538112
* Discussion on mdn/content: https://github.com/mdn/content/pull/18880 
* Next PR: #24563.

### DIFF
--- a/javascript/organizing_your_javascript_code/objects_and_object_constructors.md
+++ b/javascript/organizing_your_javascript_code/objects_and_object_constructors.md
@@ -182,8 +182,10 @@ Object.getPrototypeOf(player1) === Player.prototype
 Object.getPrototypeOf(player2) === Player.prototype
 ~~~
 
-Note: 
+Notes: 
+
 1. You might also find that `__proto__` is also referred to as `[[Prototype]]` which is from the JavaScript language specification.
+
 2. If you tried something like `Player.prototype.__proto__`, or you have a `__proto__` value which is equal to `null` and are confused, read on to find the explanation in the **Prototypal Inheritance** section!
 
 Now that you know that every object created by calling `new Player()` has its `__proto__` / `[[Prototype]]` special property set to refer to the `Player.prototype` value, let's move on!
--- a/javascript/organizing_your_javascript_code/objects_and_object_constructors.md
+++ b/javascript/organizing_your_javascript_code/objects_and_object_constructors.md
@@ -151,6 +151,43 @@ console.log(theHobbit.info());
 
 Before we go much further, there's something important you need to understand about JavaScript objects. All objects in JavaScript have a `prototype`. Stated simply, the prototype is another object that the original object _inherits_ from, which is to say, the original object has access to all of its prototype's methods and properties.
 
+To elaborate, there are two kinds of `prototype`, which are intertwined with each other. The _first_ kind is the one that exists on a **function**, which is referred to as `Player.prototype` in the case of the `Player` function in the previous example. 
+
+The _second_ kind is the one which exists on _all_ objects in JavaScript as a special property. This second `prototype` is referred to in a _non-standard_ way as, reusing our example, `player1.__proto__`, or `player2.__proto__`. This `__proto__` property comes into existence when an object is created, and is set to refer to the first kind of `prototype`. That is, in our example, `player1.__proto__` and `player2.__proto__` will have the value `Player.prototype`. You can test this in your browser console by defining `Player`, and creating the `player1` and `player2` objects.
+
+~~~javascript
+function Player(name, marker) {
+  this.name = name
+  this.marker = marker
+  this.sayName = function() {
+    console.log(name)
+  }
+}
+
+const player1 = new Player('steve', 'X')
+const player2 = new Player('also steve', 'O')
+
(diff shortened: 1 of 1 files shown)


wise-king-sullyman:
Hey @anupamvamsi! I wanted to pop in real quick, say thanks for putting up these PR's, and let you know that this _is_ under review.

As I'm sure you're super aware after doing the research to put this together, the real "in the weeds" specifics of `prototype` in JS are (in my opinion) fairly convoluted and hard to clearly discuss even between people that are experienced in the language, not to mention trying to explain it to people that are still in the early stages of learning and may not have any OOP experience to use as a reference when discussing the behavior.

For that reason, the review process on this has been pretty slow, but I hope to have some comments back to you within the next few days. Broadly I like the ideas put forward, but I might have some suggestions to reword some things for the sake of clarity 🙂

anupamvamsi:
Hi @wise-king-sullyman, thanks for letting me know! I did have another part to this PR, which I have not yet pushed. Would it be fine if I pushed it for review as well?

wise-king-sullyman:
@anupamvamsi yes that should be fine

javascript/organizing_your_javascript_code/objects_and_object_constructors.md
```diff
@@ -151,6 +151,45 @@ console.log(theHobbit.info());
 
 Before we go much further, there's something important you need to understand about JavaScript objects. All objects in JavaScript have a `prototype`. Stated simply, the prototype is another object that the original object _inherits_ from, which is to say, the original object has access to all of its prototype's methods and properties.
 
+To elaborate, there are two kinds of `prototype`, which are intertwined with each other. The _first_ kind is the one that exists on a **function**, which is referred to as `Player.prototype` in the case of the `Player` function in the previous example.
```
wise-king-sullyman, line 154: I think I get what you're going for in saying there are two "kinds" of prototype, but I feel like the way things are phrased isn't *quite* accurate. There's only one "kind" of prototype, the actual prototype of something which in the spec is referred to as `[[prototype]]`, but there's also the `.prototype` _property_ which behaves in a similar kind of way but has distinctly different rules as to what that property is.

I suppose I'm just not sure about referring to this as two "kinds" of prototype, since really one is the prototype, and the other is the value of the `.prototype` property, if that makes sense.

Do you think there might be a way you can reword this to make that differentiation more clear?

wise-king-sullyman (changes requested):
Apologies for the delay! I have one question/suggestion that I'd like to know your thoughts on regarding how `prototype` is being discussed here.

anupamvamsi:
It's been a while since I looked at this, honestly, so I'm looking at all this with a fresher, newer perspective? When I read it all again, overall, I do feel there's some things that can be improved - in addition to the points you have mentioned.

I'll try making it more clear. Hopefully by keeping it simple.

One thing I could use your advice on though -- #24563 (the next PR) adds on to the commits I've made in this PR. And #24782 (the latest one) adds on to the commits present in #24563. 

1. So, if I make changes, they would affect all the three PRs. 

2. Would it be okay if I make a new PR with all the changes that I am thinking of currently? This new PR would include the commits present in #24782 (the latest one). There are three PRs currently as I was told that separate PRs with small changes would be good and easy to review, but to me, the way I've submitted them seems to have defeated that purpose. My apologies if these PRs have been submitted in a messy way, if there's something I can do about that, please do let me know!

wise-king-sullyman:
@anupamvamsi yeah if you want to just make a new PR with all the changes you're thinking of and close your open PRs that would be fine!

And yeah generally when you want to make multiple PRs for something like this it's often best to make each one after the previous one has been merged so that you don't run into these kinds of issues. Another option would be to just rebase those other branches after your updates here are merged, but that adds an extra level of Git complication that many people aren't super comfortable with.

Absolutely no worries btw, I can recognize that you're just trying to do things in the best way 🙂

anupamvamsi:
#24941 closes this.

PR #24651, opened 2022-09-18
React Hooks: changed confusing phrasing about 'useEffect' and ESLint

marcodallagatta:

<hr>

**1. Because:**
Wording was confused and probably a result of a copy paste error.

**2. This PR:**
- The dependency array at the end is optional, however, you will include it more often than not. A dependency is any state, prop, context that is used within the `useEffect` callback. You can also include state or props that are not.
=== becomes ===
The dependency array at the end is optional, however, you will include it more often than not. A dependency is any state, prop or context that is used within the `useEffect` callback, but you can also include state or props that are not.
- `useEffect` will trigger based on changes in the dependencies listed. ESLint will warn you if it expects a dependency, but one is not added, however, this is a warning and they are not **required**. 
=== becomes ===
`useEffect` will trigger based on changes in the dependencies listed. ESLint will warn you if it expects a dependency, however, this is a warning and they are not **required**.

### DIFF
--- a/javascript/react_js/hooks.md
+++ b/javascript/react_js/hooks.md
@@ -107,7 +107,7 @@ The syntax is as follows:
 
 `useEffect(() => {}, [])`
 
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (approved):
Looks good, thank you for your contribution!

PR #24690, opened 2022-09-26
Fix Grammar: Problem Solving

Fiattarone:
Saw that we were using double quotes on double quotes for the beginning quote. Fixed by replacing inner double quotes with single quotes.

<hr>

**1. Because:**

Closes #XXXXX
Because ""Fiattarone's quote" -- Fiattarone" is incorrect, it should be:  "'Fiattarone's quote' -- Fiattarone"

It may not be obvious at first, but the " > " induces a quote line which includes the " "'s automatically. 

**2. This PR:**

- Replaced inner double quotes with a set of single quotes

### DIFF
--- a/foundations/javascript_basics/problem_solving.md
+++ b/foundations/javascript_basics/problem_solving.md
@@ -6,7 +6,7 @@ Problem solving is the core thing software developers do. The programming langua
 
 V. Anton Spraul defines problem solving in programming as:
 
-> "Problem solving is writing an original program that performs a particular set of tasks and meets all stated constraints."
+> 'Problem solving is writing an original program that performs a particular set of tasks and meets all stated constraints.'
 - Think Like a Programmer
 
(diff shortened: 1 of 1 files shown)


CouchofTomato (changes requested):
The `>` should add the quote marks for us, which is why we see the double quotes so I think we can just get rid of the quote marks altogether otherwise we'd probably see `"'` instead.

CouchofTomato:
> Actually on the fence about this one, shouldn't this be single quotes within double quotes because Anton Spraul is quoting Think Like a Programmer? I know he wrote the book, but I believe the single & double quote rules still apply because the book itself is still a separate entity. The rule also applies when quoting yourself.

Your proposed solution looks like this

image

This doesn't look right to me.

Fiattarone (commented):
Took changes from @CouchofTomato into account and fixed outstanding errors.

zachmmeyer (approved):
I like the way this looks, except for the quotes, but that's an us problem and not a you problem.

Thank you!

PR #24693, opened 2022-09-27
Typo error in for loop explanation. location of change: Problem solving: explanation of for loop.

shonebinu:
The loop is (i <= answer), so shouldn't it be "We want to loop until `i` is equal to `answer`"?

<hr>

**1. Because:**

Closes #XXXXX

**2. This PR:**

### DIFF
--- a/foundations/javascript_basics/problem_solving.md
+++ b/foundations/javascript_basics/problem_solving.md
@@ -119,7 +119,7 @@ for (let i = 1; i <= answer; i++) {
 }
 ~~~
 
-If you haven't seen this before and it looks strange, it's actually straightforward. We declare a variable `i` and assign it 1: the initial value of the variable `i` in our loop (**NOTE**: Most of the time, programmers find themselves looping from 0. Due to the needs of our program, we're starting from 1). The second clause, `i <= answer` is our condition. We want to loop until `i` is greater than `answer`. The third clause, `i++`, tells our loop to increment `i` by 1 every iteration. As a result, if the user inputs 10, this loop would print numbers 1 - 10 to the console.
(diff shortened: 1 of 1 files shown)


CouchofTomato:
Hey @shonebinu 

Thanks for creating a PR. In this case I think the existing terminology is correct. We want the loop to run up to and including until `i` is equal to the answer. For that to happen we need to increase the loop until `i` is greater than `answer` at which point to loop condition fails.

It's a difficult one because I know the point you're trying to make. We want to loop *while* `i` is less than or equal to `answer`, but since we state `until` rather than `while`, and since for loops check the condition first, we want the loop condition check to continue until `i` is higher than `answer` and then break.

It could be that we change the wording to better reflect the point trying to be made. Or that we add some further explanation that because the condition in the loop is evaluated first we need it to run until `i` is higher than `answer` and then break the loop so if you want to consider and take that on let me know. I'll leave this open for if you want to make that change.

shonebinu:
Thank you for your detailed explanation on this topic. Really!

I'm sorry, i wanted to work on this as my first pr.

But I'm only a beginner and English isn't my primary language.

It would be best if it's done by someone else.

CouchofTomato:
Thanks @shonebinu 

Appreciate you letting us know. I'll close this for now. If you want to create an issue for someone to pick up about this then let me know.

shonebinu:
Hi @CouchofTomato 
I just noticed that there isn't any semicolon at the end of 'console.log()' in this chapter (for every console.log(), there isn't any semicolon) https://github.com/TheOdinProject/curriculum/blob/main/foundations/javascript_basics/fundamentals-1.md

Should I make a pr or an issue mentioning this? Should I work on it?
I got the same results without semicolons, but I saw in other resources that it's a good practice to reduce bugs and other insecurities.

I don't want to waste maintainers' time by creating a new pull req / issue. That's why asked. Sorry if it's a stupid thing to ask.

PR #24835, opened 2022-10-22
Ruby event manager lesson: Fix csv.open and file.read filepaths

Asmatzaile:
## Because

If event_attendees.csv, form_letter.html and form_letter.erb are not in event_manager/lib, where the main script (event_manager.rb) resides, the script can't find them with the path specified in the lesson.

## This PR

- replaces all 'event_attendees.csv' with '../event_attendees.csv'
- replaces all 'form_letter.html' with '../form_letter.html'
- replaces all 'form_letter.erb' with '../form_letter.erb'

## Pull Request Requirements

### DIFF
--- a/ruby/files_and_serialization/project_event_manager.md
+++ b/ruby/files_and_serialization/project_event_manager.md
@@ -132,7 +132,7 @@ most straightforward is `File.read`.
 ~~~ruby
 puts 'EventManager initialized.'
 
-contents = File.read('event_attendees.csv')
+contents = File.read('../event_attendees.csv')
 puts contents
 ~~~
 
@@ -156,7 +156,7 @@ save each line as a separate item in an array.
 ~~~ruby
 puts 'EventManager initialized.'
 
-lines = File.readlines('event_attendees.csv')
+lines = File.readlines('../event_attendees.csv')
 lines.each do |line|
   puts line
 end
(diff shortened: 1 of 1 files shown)


ChargrilledChook (changes requested):
Hi @Asmatzaile, thanks for making a contribution! 

When you were running the commands in the lesson, how were you doing it? Were you doing it from inside the `lib` directory?

The filepaths should be based on the directory that you run your script from. If you are inside the `event_manager` directory and running your script with `ruby lib/event_manager.rb` like the examples in the lesson, the existing paths should work. 

I would be hesitant to change it to using these relative paths, because people would likely have the same problem but flipped around.

Maybe a line to clarify how that behaviour works could be useful, though? I'm open to suggestion, let me know your thoughts

Asmatzaile:
Hi @ChargrilledChook !
I was in fact running the commands from inside the `/lib` directory. The thing is that even if the `event_manager.rb` file is inside `event_manager/lib`, the CSV is not: in the lesson it states "download the csv file and save it in the root of the project, `event_manager`".
So I put that in the main directory, and from that it made sense for me to put the other non-script files there as well. The file structure I understood was the following:
 - `event_manager`
     - `lib`
            - `event_manager.rb `
     - `event_attendees.csv`
     - `form_letter.html`
     - `form_letter.erb`

ChargrilledChook:
@Asmatzaile That structure seems absolutely correct, but we want to run the command from the project root (`event_manager`) via `ruby lib/event_manager.rb`. 

If you do it like that, do the original path names work correctly? The script should still have access to the CSV even if it's in the root dir, because it resolves from where you run the script, not where the script itself is placed.

ChargrilledChook:
@Asmatzaile I'm going to close this since I belive the current behaviour is correct, but I'm happy to revisit it if I've missed something, or you'd like any further clarification. Just message me here if you'd like :slightly_smiling_face:

PR #24875, opened 2022-11-06
Include explanation for CSS values and their units

SumitChaudhary:
As I went through the various CSS examples, blog posts and videos, I observed a mix of absolute and relative values being used. Since these tutorials are for beginners, I strongly believe that adding a section on the various values and their types will help them understand all the downstream content much better.

## Because

As a beginner, it would be extremely useful to understand the various values applicable to the properties.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/foundations/html_css/css-foundations.md
+++ b/foundations/html_css/css-foundations.md
@@ -13,7 +13,7 @@ This section contains a general overview of topics that you will learn in this l
 
 ### Basic Syntax
 
-At the most basic level, CSS is made up of various rules. These rules are made up of a selector (more on this in a bit) and a semi-colon separated list of declarations, with each of those declarations being made up of a property:value pair. To understand more about the various values and their types possible, click the values and units link in the additional resources section towards the end of this page.
+At the most basic level, CSS is made up of various rules. These rules are made up of a selector (more on this in a bit) and a semi-colon separated list of declarations, with each of those declarations being made up of a property:value pair.
 
 ![Basic CSS syntax](https://cdn.statically.io/gh/TheOdinProject/curriculum/05ce472eabf8e04eeb2cc9139e66db884074fd7d/foundations/html_css/css-foundations/imgs/00.jpg)
 
--- a/foundations/html_css/css-foundations.md
+++ b/foundations/html_css/css-foundations.md
@@ -13,7 +13,7 @@ This section contains a general overview of topics that you will learn in this l
 
 ### Basic Syntax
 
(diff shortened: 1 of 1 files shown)


foundations/html_css/css-foundations.md
```diff
@@ -23,6 +23,10 @@ At the most basic level, CSS is made up of various rules. These rules are made u
 A `<div>` is one of the basic HTML elements. It is simply an empty container. In general, it is best to use other tags such as `<h1>` or `<p>` for content in your projects, but as we learn more about CSS you'll find that there are many cases where the thing you need is just a container for other elements. Many of our exercises use plain `<div>`s for simplicity. Later lessons will go into much more depth about when it is appropriate to use the various HTML elements.
 </div>
 
+### Values and Units
+
+When you are specifying any property in your CSS file, there could be applicable values depending on the property. To understand what the various values and their units could be, please go through this page [CSS values and units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units).
+
```
thatblindgeye, line 29: Because this is such a short amount of content, it feels like this would serve better as an Additional Resource, especially because it is only linking to another resource to read up on the subject.

thatblindgeye (changes requested):

foundations/html_css/css-foundations.md
SumitChaudhary: Agree with your perspective. I have moved the link to additional resources section and added one line to direct readers if they wish to learn about values and units.

foundations/html_css/css-foundations.md
```diff
@@ -13,7 +13,7 @@ This section contains a general overview of topics that you will learn in this l
 
 ### Basic Syntax
 
-At the most basic level, CSS is made up of various rules. These rules are made up of a selector (more on this in a bit) and a semi-colon separated list of declarations, with each of those declarations being made up of a property:value pair.
+At the most basic level, CSS is made up of various rules. These rules are made up of a selector (more on this in a bit) and a semi-colon separated list of declarations, with each of those declarations being made up of a property:value pair. To understand more about the various values and their types possible, click the values and units link in the additional resources section towards the end of this page.
```
thatblindgeye, line 16: We may not need this verbiage here. Typically we don't mention an additional resource exists in the middle of a lesson. I think one issue with doing so could be that it might cause users to shift to that resource and go down a rabbit hole of info they don't necessarily *need* to understand at that moment.

thatblindgeye (changes requested):

thatblindgeye:
Thanks for the addition @SumitChaudhary 🎉

PR #24885, opened 2022-11-09
Project: Micro-Reddit: Add tip for using rails c instead of rails console

Overlandcoder:
## Because

Provides readers with a helpful tip that the command `rails c` may be used instead of `rails console`

## This PR

Added a short phrase in which the tip is explained

## Issue

Closes #XXXXX

## Pull Request Requirements


CouchofTomato:
I wonder if it would be better to add a little knowledge section somewhere instead letting users know that most rails commands have shorter versions. Will also be a handy reference area for them to refer back to. What do you think?

Overlandcoder:
> I wonder if it would be better to add a little knowledge section somewhere instead letting users know that most rails commands have shorter versions. Will also be a handy reference area for them to refer back to. What do you think?

Yes that's a good idea, I was initially thinking that too but wasn't sure where a section like that should be placed.

CouchofTomato:
I'll discuss with the team and get back to you.

ChargrilledChook:
@Overlandcoder we agree that fleshing it out more would be a good addition, but we're also happy to merge this as a quick improvement. Your call how to approach it - would you like to do it as part of this PR or come back to it with a different issue?

Overlandcoder:
> @Overlandcoder we agree that fleshing it out more would be a good addition, but we're also happy to merge this as a quick improvement. Your call how to approach it - would you like to do it as part of this PR or come back to it with a different issue?

It's probably easier for me to make a new PR.

Also, I did some searching, but couldn't find any other short-cut commands other than "g", "s" and "c", at least not ones that are useful for that point in the course. I'm thinking we could add a little paragraph right above this section (https://www.theodinproject.com/lessons/ruby-on-rails-micro-reddit#your-task). Or would it be better somewhere else?

ChargrilledChook:
@Overlandcoder There's a few different ways we could go with it - could you make an issue for it and we can discuss more there? If you reference this PR in the issue it will give it some more context too

PR #24899, opened 2022-11-15
Add template for generic issue form

ChargrilledChook:
## Because
* It can help create guidelines for contributors to make better issues. and help maintainers by requiring a better baseline of info

## This PR
* Adds a YAML based GH issue form

## Additional Information

See the[ docs for more info](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)

### DIFF
--- a/.github/ISSUE_TEMPLATE/suggestion.yaml
+++ b/.github/ISSUE_TEMPLATE/suggestion.yaml
@@ -54,7 +54,7 @@ body:
     id: contact
     attributes:
       label: (Optional) Discord Name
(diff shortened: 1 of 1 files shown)


ChargrilledChook:
> This looks awesome. Just a small github actions spelling error suggested change. I say we get it out and then adjust if necessary.

@CouchofTomato Good catch on the typo, that was a very subtle failure :laughing: fixed. Agree on the iterating, I think that's a good approach

PR #24931, opened 2022-11-27
Factory Functions And The Module Pattern lesson: Add mention of the prototype

### LINKED ISSUE #23105: Factory functions lesson doesn't mention performance penalty compared to constructors
[The lesson in question](https://www.theodinproject.com/paths/full-stack-javascript/courses/javascript/lessons/factory-functions-and-the-module-pattern)

The resources in the previous lessons talked about how important it was to put methods on the prototype instead of the constructor itself, since that would mean the method isn't duplicated on every instance. However, that isn't done on factory functions at all. In my opinion this feels confusing, having the prototype introduced and talked about in terms of efficiency and in this lesson having it ignored. 

I suggest that an explanation for this be added so that it doesn't leave you confused as to why the prototype was introduced previously (e.g. Using factory functions doesn't utilize the prototype, which makes them less efficient, but they are a good choice when [insert when]).

EmilioAK:
## Because

The resources in the previous lessons talked about how important it was to put methods on the prototype instead of the constructor itself, since that would mean the method isn't duplicated on every instance. However, that isn't done on factory functions at all. In my opinion this feels confusing, having the prototype introduced and talked about in terms of efficiency and in this lesson having it ignored.

## This PR

- Adds a simple explanation as to why factory functions are useful even though they do not utilise the prototype.

## Issue

Closes #23105

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
+++ b/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
@@ -53,6 +53,8 @@ const Person = function(name, age) {
 const jeff = new Person('jeff', 27);
 ~~~
 
+Note that factory functions do not utilize [the prototype](https://www.theodinproject.com/lessons/node-path-javascript-objects-and-object-constructors#the-prototype), which does come with a performance penalty. However, as a general rule, this peanalty isn't significant unless you're creating thousands of objects.
+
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (approved):
Looks good, thanks!

PR #25041, opened 2023-01-22
Foundations: Added iframe sandbox options

lnxd:
## Because
The Chromium console (assuming also all other browsers that have introduced iframe sandboxes) is full of error messages on all pages that contain iframes. 

## This PR
This whitelists the [iframe sandbox options](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox):
- allow-scripts
Lets the resource run scripts (but not create popup windows).

- allow-same-origin
If this token is not used, the resource is treated as being from a special origin that always fails the [same-origin policy](https://developer.mozilla.org/en-US/docs/Glossary/Same-origin_policy) (potentially preventing access to [data storage/cookies](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_data_storage_access) and some JavaScript APIs).

These options remove the error messages and allow the iframes to continue working as expected. 

Unfortunately there are still some error messages generated by Scrimba's use of iframes within their embed URL (nested iframes for TOP's purpose) without adding code to handle the new sandbox rules. But the console is a lot cleaner from this change regardless. 

## Issue
As this is a small change I haven't created an issue for this. I came across the problem while trying to troubleshoot an unrelated problem with a cookie setting on my browser. I found it hard to troubleshoot as the console was flooded with irrelevant errors. 

## Additional Information
I understand the potential security implications of doing this, for example if [Scrimba](https://scrimba.com/) or [wheeldecide.com](https://wheeldecide.com) are compromised, but as this course needs to use external resources regardless I believe it's a risk worth taking. 

In support of this perspective, there's an iframe in `/react/class_based_components.md` where iframe sandboxing is disabled completely for [codesandbox.io](https://codesandbox.io).

## Pull Request Requirements

### DIFF
--- a/foundations/html_css/flexbox/flexbox-alignment.md
+++ b/foundations/html_css/flexbox/flexbox-alignment.md
@@ -37,11 +37,11 @@ Because `justify-content` and `align-items` are based on the main and cross axis
 
 Check out this Scrim for an interactive demo of how `justify-content` and its different properties behave:
 
-<iframe src="https://scrimba.com/learn/flexbox/justify-content-flexbox-tutorial-cVWPacR?embed=odin,mini-header,no-big-play,no-next-up" width="100%" height="400"></iframe>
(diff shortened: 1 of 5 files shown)


KevinMulhern (approved):
Looks good to me @lnxd, Thanks for the detailed write up 💪

PR #25101, opened 2023-02-14
grammar fix : express 101

omthakare16:
Grammar fixed in line number 17

## Because
The grammar was was wrong in line number 17 instead of  " Most of the Odin staff prefer ejs to Pug simply because we like working with HTML " it should be " Most of the Odin staff prefer ejs over Pug simply because we like working with HTML "


KevinMulhern (approved):
Good change, thanks @omthakare16

omthakare16:
> Good change, thanks @omthakare16

Hey, I didn't expect it to be merged because it was a very small contribution though it did. Thanks for that. Hope I might help in more contribution

PR #25111, opened 2023-02-17
Change functions to methods

aziham:
- The word **methods** is the right word in the context of an object.

### DIFF
--- a/foundations/javascript_basics/DOM_manipulation_and_events.md
+++ b/foundations/javascript_basics/DOM_manipulation_and_events.md
@@ -323,7 +323,7 @@ btn.addEventListener('click', function (e) {
 ~~~
 \*_Note that `function (e)` is a callback from addEventListener. Further explanation of callbacks can be found [HERE](https://dev.to/i3uckwheat/understanding-callbacks-2o9e)._
 
-The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and methods such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
+The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and methods (functions that live inside an object) such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
 
 Try this:
 
--- a/foundations/javascript_basics/DOM_manipulation_and_events.md
+++ b/foundations/javascript_basics/DOM_manipulation_and_events.md
@@ -323,7 +323,7 @@ btn.addEventListener('click', function (e) {
 ~~~
 \*_Note that `function (e)` is a callback from addEventListener. Further explanation of callbacks can be found [HERE](https://dev.to/i3uckwheat/understanding-callbacks-2o9e)._
 
-The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and functions such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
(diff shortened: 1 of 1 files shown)


xandora:
This is true, but it would be the first time the word "method" is introduced to the learner, so we might need also some expounding of what `method` means.

xandora (changes requested):
Needs an explanation of what a `method` is/means.

aziham:
> This is true, but it would be the first time the word "method" is introduced to the learner, so we might need also some expounding of what `method` means.

We might just make the word "method" as a hyperlink that links to an external resource explaining what a method is, or an internal one if it exists.

xandora:
@aziham It could be something as simple as:

"[and] useful properties and methods (functions stored within the object itself) such as which mouse button or key was [pressed]"

This does present an opportunity for a potential feature improvement to the platform through either definition boxes when we introduce new terms, or definition hoverlinks (words that when you hover over them show alt text like definitions.)

> This is true, but it would be the first time the word "method" is introduced to the learner, so we might need also some expounding of what `method` means.
> 
> We'll end up explaining what properties are also. The lesson will be too dense if we kept adding those definitions.
> 
> I suggest to just leave it as it is, this will make the learner more curious about what is a method or a property then he can go do little research about it (good opportunity to try googling stuff).
> 
> Or we can keep the learner on training wheels mode and make the word "method" as a hyperlink that links to an external resource explaining what a method is, or an internal one if it exists.

foundations/javascript_basics/DOM_manipulation_and_events.md
```diff
@@ -323,7 +323,7 @@ btn.addEventListener('click', function (e) {
 ~~~
 \*_Note that `function (e)` is a callback from addEventListener. Further explanation of callbacks can be found [HERE](https://dev.to/i3uckwheat/understanding-callbacks-2o9e)._
 
-The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and functions such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
+The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and methods such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
```
xandora, line 326: Something like this would be passable IMO.

```suggestion
The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and methods (functions that live inside an object) such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
```

xandora (commented):
Something like the suggested change perhaps?

foundations/javascript_basics/DOM_manipulation_and_events.md
```diff
@@ -323,7 +323,7 @@ btn.addEventListener('click', function (e) {
 ~~~
 \*_Note that `function (e)` is a callback from addEventListener. Further explanation of callbacks can be found [HERE](https://dev.to/i3uckwheat/understanding-callbacks-2o9e)._
 
-The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and methods such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
+The `e` in that function is an object that references the __event__ itself.  Within that object you have access to many useful properties and methods (functions that live inside an object) such as which mouse button or key was pressed, or information about the event's __target__ - the DOM node that was clicked.
```
xandora, line 326: Looks good to me.

xandora (approved):

PR #25167, opened 2023-03-04
Rails: Installing Postgres: Add note for handling capitalized usernames

JoshDevHub:
## Because

One of the setup goals with this lesson is to have learners create a database in their postgres service with a name that matches their OS username (allowing for peer authentication). If the user has any capital letters in their username, they need to wrap their username in quotation marks when running the command to create the database. This is because postgres automatically folds all identifiers to lowercase *unless* they're wrapped in quotes. So if a user with the username `Josh` tries to create a database and just says `CREATE DATABASE Josh;`, postgres will create a database called `josh`, and this won't work correctly for the following steps in the lesson. The user would need to do `CREATE DATABASE 'Josh';`

## This PR

- Add a note telling users who have capital letters in their username to be sure to surround their name in quotation marks when running the command to create the database.

## Issue

N/A

## Additional Information

Link to a conversation where this was a problem: https://discordapp.com/channels/505093832157691914/514204607770001411/1042881200513089707

## Pull Request Requirements

### DIFF
--- a/ruby_on_rails/advanced_forms_and_activerecord/installing_psql.md
+++ b/ruby_on_rails/advanced_forms_and_activerecord/installing_psql.md
@@ -108,11 +108,13 @@ One other important step in setting up PostgreSQL is that each role must have it
 
 You can try to run `psql` now, but you will get an error that the database does not exist. Not to worry, let's create one to resolve fix this:
 
+<div class="lesson-note" markdown="1">
+  If your username has any capital letters, you must surround it in quotes when running the below command.
+</div>
+
 ~~~bash
 sudo -i -u postgres createdb <linux_username>
 ~~~
-
-*note: If your username has any capital letters, you must surround it in quotes when running the above command*
 
 Now our role is fully set up: we've got `<role_name>` and that role has a database.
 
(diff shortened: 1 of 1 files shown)


ruby_on_rails/advanced_forms_and_activerecord/installing_psql.md
```diff
@@ -112,6 +112,8 @@ You can try to run `psql` now, but you will get an error that the database does
 sudo -i -u postgres createdb <linux_username>
 ~~~
 
+*note: If your username has any capital letters, you must surround it in quotes when running the above command*
```
KevinMulhern, line 115: Just to bring extra attention to these notes, we could use our note box styling:

```suggestion
<div class="lesson-note" markdown="1">
 If your username has any capital letters, you must surround it in quotes when running the above command*
</div>
```

I wonder if we should include these notes above the commands incase they are only seen after users run the command? But I'm really not sure if it will matter too much if the notes are already highlighted enough. Not a blocker, whatever you think is best works for me.

Screenshot 2023-03-04 at 12 00 19

JoshDevHub: Both of those ideas are great I think! I'll make those changes

KevinMulhern (approved):
Great stuff, thanks Josh!

ruby_on_rails/advanced_forms_and_activerecord/installing_psql.md
JoshDevHub: I added the changes in. One thing I was struck by is that the two different sections preview with different gap styles between the note block and the code block below it.

From the linux section:
linux_section_note

From the MacOS section:
mac_section_note

From using the inspector and experimenting, it seems to be caused by the linux section have the `bash` tag on the code block, which is giving it less of a top margin compared to the MacOS section, which uses a generic code block (that code is run in the context of pgsql service, so it wouldn't be semantically correct to call it bash). Just something I thought you might want to be aware of @KevinMulhern -- I personally think the MacOS section's styling looks slightly better but not a big deal or anything.

KevinMulhern: Thanks for pointing this out @JoshDevHub. I agree the Mac OS spacing looks much better and should be the version we make everything consistent with. 

It should hopefully just require a little tweaking in our [code syntax custom styling file](https://github.com/TheOdinProject/theodinproject/blob/main/app/javascript/stylesheets/prism-theme.css). Would you like to make an issue in the site repo about it?

JoshDevHub: Sure thing. I'll make the issue on the site repo :+1:

PR #25289, opened 2023-03-26
CS > Time Complexity: Fix step calculation

c-auri:
## Because

The step calculation for `oddNumbersLessThanTen` is inaccurate: When executing the `while` loop, the condition is checked in the beginning of every iteration **and** once more after the last iteration (when `currentNumber` equals 10). This last check is not mentioned in the lesson.

## This PR

- Mentions the last check of the loop condition in the step listing
- Updates the step calculation accordingly
- Refers to the new total number of steps in the following text

## Additional Information

I understand that calculating the steps of a program is not a well defined process. For example, one could argue that the  `currentNumber += 1` in the last line of the loop is actually two steps (increment and assignment). Presumably, this is still counted as 1 step here to keep the example simple. I agree with that.

However, I feel that the checking of the exit condition is different and should be mentioned for two reasons:
1. It is counted for all the other iterations and is therefore established as something that takes exactly 1 step.
2. When running through the program with the debugger, that last check will show up. The total number of steps you need to manually step over (F10) to reach the end of the program is 34.

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/time_complexity.md
+++ b/javascript/computer_science/time_complexity.md
@@ -54,9 +54,12 @@ Let's go back to our `oddNumbersLessThanTen function`. How many steps does our a
     3. If it is then we output it to the terminal. That's 1 step every 2 iterations.
     4. We increase `currentNumber` by 1. That is 1 step.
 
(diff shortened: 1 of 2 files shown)


CouchofTomato (approved):
Thanks @c-auri 

Agree this is a good addition.

PR #25390, opened 2023-04-24
Fundamentals Part 2 lesson: Update linked article for Switch statements

### LINKED ISSUE #24723: Bug - : Foundations - Fundamentals Part 2 - Linked article (#6) under Conditionals has incorrect information
-   [x] The title of this issue follows the `Bug - location of bug: brief description of bug` format, e.g. `Bug - React section: Knowledge Checks don't link to resource`

The following checkbox is OPTIONAL:

-   [x] I would like to be assigned this issue to work on it

<hr>

**1. Description of the Bug:**

The linked article from digital ocean contains incorrect information about switch statements, and needs to be replaced with a better, more accurate tutorial. An alternative could be this great article from [javascript.info](https://javascript.info/switch)

**2. How To Reproduce:**
1. Visit the [Fundamentals Part 2](https://www.theodinproject.com/lessons/foundations-fundamentals-part-2) lesson.
2. Scroll down to the [Conditonals](https://www.theodinproject.com/lessons/foundations-fundamentals-part-2#conditionals) section.
3. Go the tutorial from digital ocean linked at [number 6](https://www.digitalocean.com/community/tutorials/how-to-use-the-switch-statement-in-javascript)
4. Scroll down to the paragraph just above the 'Switch Ranges' heading.
5. The text reads: _'If we had omitted the break keyword in each statement, none of the other case statements would have evaluated to true, but the program would have continued to check until it reached the end. In order to make our programs faster and more efficient, we include the break.'_
6. This is wrong as it implies that the reason of using a break keyword in every case of a switch statement is efficiency, and the code would run correctly even if it was omitted. That is not the case as omitting a break leads to the code in next case running, without the condition being checked.

**3. Expected Behavior:**
1. The linked tutorial on Switch statements should have the correct information, especially since this is supposed to be the reader's first introduction to the topic.
2. I also do not like how the tutorial goes on to teach the use of evaluating a different condition in each case, by keeping the original switch expression to be just 'true', and thus evaluate on a 'range of values'. This is the typical case of not using the right tool for a particular job, as an if-else construct is always the better solution in such cases.

**4. Desktop/Device:**
-   Device: MacBook Pro
-   OS: macOS Monterey 12.3
-   Browser: Google Chrome

**5. Additional Information:**
A screenshot with the highlighted erroneous text:
image

ziayanj:
## Because
The linked article from DigitalOcean contains incorrect information about switch statements, and needs to be replaced with a better, more accurate tutorial.

## This PR
- Replaces the article from DigitalOcean with this great article from [javascript.info](https://javascript.info/switch).

## Issue
Closes #24723

## Additional Information
Confirmed to be an issue by @CouchofTomato 

## Pull Request Requirements

### DIFF
--- a/foundations/javascript_basics/fundamentals-2.md
+++ b/foundations/javascript_basics/fundamentals-2.md
@@ -33,7 +33,7 @@ Now it's time for the fun stuff...  So far we haven't done much with our program
 3. [This tutorial](http://javascript.info/logical-operators) will teach you about logical operators. A little heads up regarding this reading's tasks; there will be questions where you see `alert()` (with a number or string inside of the parenthesis), what's happening here will be discussed later in the curriculum. Some of the answers may not make sense now but they are accurate and will come to be understood as you progress in the curriculum. Don't worry too much about it now!
 4. [This article](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/conditionals) reinforces the concept and provides several interesting examples of how you could use it building websites.
 5. [This article](http://javascript.info/ifelse) covers the same basic concept \(read through it as a review!\) and - more importantly - offers the usual 'tasks' at the bottom of the page!
-6. [This tutorial](https://www.digitalocean.com/community/tutorials/how-to-use-the-switch-statement-in-javascript) teaches you about the `switch` statement, which comes in handy when you have multiple conditions.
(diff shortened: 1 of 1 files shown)


KevinMulhern (approved):
Looks great to me, thank you @ziayanj

PR #25593, opened 2023-06-05
JS CS Time Complexity: edit sentence and grammar

ManonLef:
## Because

I think these changes make these sentences more correct and/or easier to follow.
I also changed a sentence that mentioned we covered different ways of 'measuring' efficiency. I think we covered different ways of "quantifying" or different "notations" of efficiency with Big O, Theta and Omega, and those 3 are what that sentence seems to refer to. Some terms are used interchangeably online. If you think 'measuring' would not be the best term, it may be best to review other parts of the lesson as well. 

## This PR

- change `instead of set` to `instead of setting` 
- change a sentence as seen below and in the commit. I personally found this sentence a bit hard to read and think the updated version might be easier to follow and more correct. 

before commit:
> Now we've touched on the different ways we can measure an algorithm's efficiency, it's hopefully clear on why it is that we choose to use the worst-case scenario when measuring the efficiency of that algorithm.

after commit:
> Now that we've touched on the different ways of quantifying an algorithm's efficiency, hopefully it's clear why we choose to use the worst-case scenario when measuring the efficiency of that algorithm.

## Issue

Closes #XXXXX

## Pull Request Requirements


CouchofTomato:
Thanks @ManonLef 

As things stand we unfortuantely have to maintain two different lessons in the CS section, one for Ruby and one for JS. Any changes made in one need to be mirrored in the other. Are you happy to make those changes too?

There are also a couple of failures on the spell checking which you can fix if you want to at the same time (not related to your own PR)

ManonLef:
@CouchofTomato I considered the separate lessons across the paths not long before you responded. I'd be happy to edit those as well. 

I'm a bit puzzled why the spelling errors came up on a file I didn't commit changes for. Maybe you can shed some light on that. I don't mind picking those up however.

Feel free to assign the above :)

For possible future edits; Do you have any opinions on the usage of the word 'measuring' throughout these lessons? My current idea is that we **quantify** the complexity with a Big O (or other) notation and we don't technically **measure** it unless we would plug in a value of `n`, and take into account the constants as well. The results of the quantifications give us a way to compare different algorithm complexities taking out variables that are insignificant in the grand scheme of scaling. Since I'm rather new to the topic I'm not 100% sure of my understanding (nor on my english) so would love your view.

CouchofTomato:
@ManonLef 

Yeah I do get what you mean. Most articles online also use measurement that I've seen so I'm not sure we need to stress to avoid it. I guess Big O is more a way express the efficiency rather than any kind of measurement since within Big O you can also have a wide variety of performance outcomes.

So if you do want to go ahead and invest the time to change it I wouldn't be opposed.

Are you happy for me to merge this in the meantime?

ManonLef:
Yes I noticed the same thing online. 

You can absolutely go ahead and merge these changes. 

I believe the wording doesn't cause any confusion to most learners and is therefore not imminent. I'll personally note it for future consideration. 

Thanks for taking the time!

CouchofTomato (approved):
Thanks @ManonLef

PR #25780, opened 2023-07-10
JavaScript: Remove Additional Resources

Raymond-Adams:
## Because

References to resources with a wide scope of content (i.e. books and long playlists) are unproductive for learners. Instead, we should select specific chapters or videos for relevant lessons.

## This PR

1. Removes the ambiguous additional resources section, including YDKJS and a ~7.5-hour 'Namaste JavaScript' playlist.

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/organizing_your_javascript_code_introduction.md
+++ b/javascript/organizing_your_javascript_code/organizing_your_javascript_code_introduction.md
@@ -19,5 +19,6 @@ Going through these will give us a chance to learn about a few other important c
 
 This section contains helpful links to other content. It isn't required, so consider it supplemental.
 
-- The book [You Don't Know JS (YDKJS)](https://github.com/getify/You-Dont-Know-JS/tree/1st-ed#titles) is free on GitHub, and explains how Javascript works "under the hood". If you ever wondered why JavaScript works the way it does, this book is for you!
+- The book [You Don't Know JS (YDKJS)](https://github.com/getify/You-Dont-Know-JS/) is free on GitHub, and explains how Javascript works "under the hood". If you ever wondered why JavaScript works the way it does, this book is for you!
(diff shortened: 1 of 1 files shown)


bycdiaz:
@Raymond-Adams I'm curious about why Eloquent JavaScript should be mentioned any time YDKJS is mentioned. Could you share your motivation for that?

I've found that people will often read everything in additional resources when additional resources are present. I think we can improve the presence of these two resources by telling learners exactly what content in both of these resources they should read. I haven't read all of Eloquent JavaScript but I know that anyone going through our curriculum would not benefit from reading all of YDKJS in its entirety here. Do you have a sense of what sections a learner should focus on for both resources?

bycdiaz (changes requested):
Left a comment. With how much content comes from those two resources, I think we need to clarify what content learners should read when visiting both of those resources. I think the language as it is would be an invitation to do everything. And at this point in the curriculum, doing everything from both of those resources is not productive.

Raymond-Adams:
@bycdiaz Both are excellent standards for learning JavaScript and provide robust overviews while remaining suitable for beginners.

I agree that their scope is vast. Considering that 'Organizing Your Javascript Code Introduction' is nonspecific, it may be appropriate to include bulkier resources to complement the entire chapter, 'Organizing Your Javascript Code', or the whole course, 'JavaScript'--perhaps with more explicit notice. Alternatively, we can remove the extensive resources altogether and include only specific chapters (e.g. [foundations/DOM Manipulation and Events](https://www.theodinproject.com/lessons/foundations-dom-manipulation-and-events) references ch. 13 & 14 of Eloquent JavaScript) in relevant lessons.

bycdiaz:
@Raymond-Adams I think your idea of pointing learners to specific chapters based on lesson content makes a ton of sense.

I think including two very big resources with no guidance is going to set up learners for a lot of frustration and we should avoid doing that when possible.

bycdiaz (approved):
Thanks for this change. While those resources are useful, I think the change here will help keep folks focused.

PR #25781, opened 2023-07-10
 JavaScript: Project: Linked Lists lesson: Add minor reason why linked lists aren't necessary in JS

### LINKED ISSUE #25771: JavaScript: Project: Linked Lists
### Describe your suggestion

In the Linked List lesson it is stated that the benefit of linked lists over arrays are:
1. linked list elements can be easily inserted or removed without reallocation of other elements
2. in some programming languages the array size can be problematic, e.g. if you want to add a 6th element to an array of size 5.

It is then later stated that Linked Lists aren't really necessary in JavaScript, because the array size is not a limitation in JavaScript
> So if array size is not a limitation in Javascript, are linked lists really necessary? The short answer to that is _no_;

The reason above, although valid, only addresses the second point mentioned above. It doesn't address the first point (how insertion/removal in JavaScript arrays is possible without manual reallocation of other elements)

- Solution
	- also mention that native array methods, like `array.splice`, are also the reason why linked lists aren't really necessary in JavaScript. 

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-linked-lists

### Checks

### (Optional) Discord Name

mogoatlhe

_No response_

Mogoatlhe:
## Because

In the linked list lesson (https://www.theodinproject.com/lessons/javascript-linked-lists) `two` reasons are mentioned as to why linked lists are beneficial over arrays (`easy insertion/deletion (no need to move other elements around)` and `size not being a problem`). But only `one` of the reasons (`array size not being a limitation in JavaScript`) is given as to why linked lists aren't really necessary in JavaScript. That has left me (the reader) thinking "wait... what about the other benefit you mentioned? does that benefit also not make a difference in JavaScript, and if so how?"

## This PR

- This PR explains that despite the linked list benefit (elements can easily be inserted or removed without reallocation of any other elements) mentioned in the lesson linked lists aren't really necessary.

## Issue

Closes #25771

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_linked_lists.md
+++ b/javascript/computer_science/project_linked_lists.md
@@ -8,7 +8,7 @@ removed without reallocation of any other elements.
 In some programming languages, the size of an array is a concern and one of the ways
 to overcome that problem and allow dynamically allocated data is using linked lists.
 
-Luckily in **Javascript**, arrays aren't limited to a certain size, and both insertion and deletion can be done trivially at any index using the `slice` array method, so you don't have to think about overcoming those limitations.
+Luckily in **Javascript**, arrays aren't limited to a certain size, and both insertion and deletion can be done trivially at any index using the `splice` array method, so you don't have to think about overcoming those limitations.
 
 So if array size, array insertion and array deletion are not limitations in Javascript, are linked lists really necessary?
 The short answer to that is _no_; however, it's the simplest of the dynamic data
--- a/javascript/computer_science/project_linked_lists.md
+++ b/javascript/computer_science/project_linked_lists.md
@@ -8,7 +8,7 @@ removed without reallocation of any other elements.
 In some programming languages, the size of an array is a concern and one of the ways
 to overcome that problem and allow dynamically allocated data is using linked lists.
 
-Luckily in **Javascript**, arrays aren't limited to a certain size, and both insertion and deletion can be done trivially at any index using the `splice` array method, so you don't have to think about overcoming those limitations.
+Luckily in **Javascript**, arrays aren't limited to a certain size, and both insertion and deletion can be done trivially at any index using the appropriate built in array method, so you don't have to think about overcoming those limitations.
 
 So if array size, array insertion and array deletion are not limitations in Javascript, are linked lists really necessary?
 The short answer to that is _no_; however, it's the simplest of the dynamic data
--- a/javascript/computer_science/project_linked_lists.md
+++ b/javascript/computer_science/project_linked_lists.md
@@ -8,10 +8,9 @@ removed without reallocation of any other elements.
 In some programming languages, the size of an array is a concern and one of the ways
 to overcome that problem and allow dynamically allocated data is using linked lists.
 
-Luckily in **Javascript**, arrays aren't limited to a certain size, so you don't have to think
-about overcoming that limitation.
+Luckily in **Javascript**, arrays aren't limited to a certain size, and both insertion and deletion can be done trivially at any index using the `slice` array method, so you don't have to think about overcoming those limitations.
 
-So if array size is not a limitation in Javascript, are linked lists really necessary?
+So if array size, array insertion and array deletion are not limitations in Javascript, are linked lists really necessary?
 The short answer to that is _no_; however, it's the simplest of the dynamic data
 structures and it will give you a solid foundation, so you can understand more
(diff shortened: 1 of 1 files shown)


Mogoatlhe:
wait... i meant `splice`

Mogoatlhe:
Should this get approved, for the Ruby section I'd mention both `insert` and `delete` arr methods instead of `splice` since I can't seem to find the Ruby equivalent of JS `splice`.

CouchofTomato:
I think I'm more in favour of keeping it more generic and not mentioning specific methods. Learners should already know what they are by this point.

CouchofTomato:
> Agreed. I'll update the PR accordingly.

Thanks. Can you make the same change to the Ruby course lesson and then we can close this out.

Mclilzee:
In my opinion it would be best if we drop the "we don't need linkedlist, because we have array" and give a brief summary of why JavaScript (or Ruby) haven't Implemented a built in LinkedList instead.

JavaScript was initially designed as a simple scripting language for web browsers, with the primary focus on providing a way to manipulate the Document Object Model (DOM) of web pages. The language has evolved over time and now includes features such as built-in support for object-oriented programming and functional programming, but it does not include many of the advanced data structures that are commonly found in other programming languages. JavaScript focused on providing easy to use general purpose language with less complexity than other languages. 

Not all languages Implement all features, and LinkedLists is one of those features, that if you need to benifit from it you need a very big data set while you perform insertion and deletion of that set very frequently, which is something JavaScript is not used for in the real world.

As for Ruby its about the same Idea i would say, we just need to replace JavaScript with Ruby, and remove the Web page language details.

I would also go further and remove the part where it says, "In other languages where Arrays is limited sized people would implement the LinkedList", cos that is not entirely true. All arrays are limited in size, and languages would implement ArrayLists which we know as (Array) in JavaScript, or in other languages like Rust and C++ they would call it a Vector. You can think about them as "Array managers" Its their job to allocate new memory for the new array with bigger size and copy data, insertions and deletion and all that.

Mogoatlhe:
@CouchofTomato I have made created a PR for the Ruby course lesson here: #25787

Mogoatlhe:
> In my opinion it would be best if we drop the "we don't need linkedlist, because we have array" and give a brief summary of why JavaScript (or Ruby) haven't Implemented a built in LinkedList instead.


Up until this point in the curriculum the only data structure that we've used is an array. So in order to make it easier to understand what a linked list is and its benefits, it makes sense to compare it to an array. Now the mentioned linked lists benefits are relevant in  most languages, but not in JavaScript and Ruby. And all that we're doing here is mentioning why they're not necessary in JavaScript and Ruby. So I think my initial point still stands.

> LinkedLists is one of those features, that if you need to benifit from it you need a very big data set while you perform insertion and deletion of that set very frequently


I disagree with this. I gave you an example (in the issue #25771  comments) detailing how managing an array of size 10 would be tedious without the JS built array methods. In that case a linked list would be beneficial, no big data set needed.

> ...which is something JavaScript is not used for in the real world


JavaScript is used as a back-end language, I'm pretty sure that there are plenty of nodeJS implementations that handle lost of data... but this is speculative (from the both of us)

> I would also go further and remove the part where it says, "In other languages where Arrays is limited sized people would implement the LinkedList", cos that is not entirely true


There is no such statement in the lesson. The lesson states: "In some programming languages, the size of an array is a concern and `one of the ways` to overcome that problem and allow dynamically allocated data is using linked lists." Which is true, linked lists help with not having to deal with manually managing an array because it has reached its capacity.
> All arrays are limited in size, and languages would implement ArrayLists which we know as (Array) in JavaScript, or in other languages like Rust and C++ they would call it a Vector.


JavaScript arrays are not limited in size (from an implementation level, i.e. the coder's POV). Each and every language has their own implementation of an array. Regardless of how it's implemented under the hood, it's an array as far the reader is concerned. So what we're doing is we're drawing parallels between these arrays and linked lists to make it easier for the reader to see the benefits of a linked list. (which don't really matter in JavaScript because of how its array is implemented under the hood).

I agree that a JavaScript array is more a C++ vector than it is a C++ array. But from the reader's POV it's an array, and all that we're doing is drawing parallels between these languages' array equivalent and linked lists to show the benefits of a linked list from an angle that they're already comfortable with.

CouchofTomato (approved):

PR #25817, opened 2023-07-17
few typos and grammar fixes

ShivanshBhatnagar8:
I found out there was some grammatical errors which I tried to fix. This mistakes are less in number and unnoticeable. There is no harm if these are not fixed but still I tried to fix some errors to contribute to this project .

### DIFF
--- a/react/introduction/introduction_to_react.md
+++ b/react/introduction/introduction_to_react.md
@@ -10,11 +10,11 @@ This section contains a general overview of topics that you will learn in this l
 
 ### What Is React?
 
-React according to their [official website](https://react.dev/) is "The library for web and native interfaces" however lets break this down a bit further.
+React according to their [official website](https://react.dev/) is "The library for web and native interfaces" however let's break this down a bit further.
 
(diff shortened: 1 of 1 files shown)


react/introduction/introduction_to_react.md
```diff
@@ -28,7 +28,7 @@ Just to name a few reasons on why to learn React:
 - Reusability of components
 - Well supported due to its popularity and large community
 - React is not opinionated, which means that it won't force you to follow any specific design patterns, project organizational structure, or logic. It's all up to you.
-- Smaller learning curve, especially when you already have a good grasp of JavaScript and HTML/CSS from our previous lessons.
+- The smaller learning curve, especially when you already have a good grasp of JavaScript and HTML/CSS from our previous lessons.
```
rlmoser99, line 31: I am not sure "the" is the right word here. I would either go with nothing, or "A". 

```suggestion
- A smaller learning curve, especially when you already have a good grasp of JavaScript and HTML/CSS from our previous lessons.
```

rlmoser99 (changes requested):
These changes look good to be, with the exception of one. You can either accept my suggestion, or make a different change that you think would work better.

react/introduction/introduction_to_react.md
ShivanshBhatnagar8: Yes ,I can go with your suggestion.

PR #25830, opened 2023-07-18
Fix missing {} around animal

Oscar-P-E:
## Issue:
Missing {} around the word _animal_ in the .map function.

This will cause it to return the string "animal" for each item in the array.

----

The existing code will produce this:

### Animals:
- animal
- animal
- animal
- animal


linkonsat:
Thanks for doing this.

PR #25900, opened 2023-07-21
Wai_aria lesson: Rewrite the text to be clearer and fix grammar errors

Eveieve:
## Because

* Some sentences could use more active voice
* It could use a link to a Wikipedia page definition of 'braille' for those who never encountered the word 
* The use of the word 'though' could be reduced
* The sentence could be made more concise
* The lesson contains some grammar errors 

## This PR

* Rewrite some setences to be more concise and to the point 
* Adds a hyperlink to the Wikipedia page for the word 'braille'
* Fix grammar errors 

## Issue

Closes #XXXXX

## Pull Request Requirements


advanced_html_css/accessibility/wai_aria.md
```diff
@@ -58,29 +58,29 @@ When using such ARIA attributes, you would provide an `id` to one element, and y
 
 #### `aria-label`
 
-The `aria-label` attribute overrides any native label and modifies the name property in the accessibility tree, though it's best used when an element doesn't already have a native label. When you add `aria-label` to an element, you pass in a string as the value, which will become that elements accessible name. `aria-label` doesn't work on every HTML element, though. Adding the attribute to a plain `<div>` or a `<span>` will have no effect, for example.
+The `aria-label` attribute overrides any native label of an element and modifies its name property in the accessibility tree. It is best used when an element doesn't already have a native label. The string value of `aria-label` attribute  becomes the element's accessible name. `aria-label` does not have any effect on some HTML elements, such as `<div>` or a `<span>`.
```
Lofty-Brambles, line 61:
```suggestion
The `aria-label` attribute overrides any native label of an element and modifies its name property in the accessibility tree. It is best used when an element doesn't already have a native label. When added, the string value of `aria-label` attribute  becomes the element's accessible name. **However**, [`aria-label` does not have any effect on some HTML elements](https://github.com/w3c/aria/issues/756), such as `<div>` or a `<span>`. 
```

advanced_html_css/accessibility/wai_aria.md
```diff
@@ -58,29 +58,29 @@ When using such ARIA attributes, you would provide an `id` to one element, and y
...
 ~~~
 
 Once a screen reader reaches the above HTML, it would announce "Main navigation, navigation landmark". If you had multiple navigation elements on a page, you could give each a different `aria-label` value in order to separate them from one another, making them more understandable for screen reader users. Pretty neat, huh?
 
-One thing you should avoid using `aria-label` for is trying to change how a word is phonetically announced. Certain words may not get announced correctly by a screen reader, and you may have a temptation to try and fix this. To put it simply: **don't**. You may end up fixing how a word is announced by a screen reader, but that "fix" could end up making no sense when announced by other assistive technologies like a braille reader.
+Do not use `aria-label` to change how a word is phonetically announced. Certain words may not get announced correctly by a screen reader, and you may have a temptation to try and fix this. You may end up fixing how a word is announced by a screen reader, but that "fix" could end up making no sense when announced by other assistive technologies like a [braille](https://en.wikipedia.org/wiki/Braille) reader.
```
Lofty-Brambles, line 77:
```suggestion
**Do not** use `aria-label` to change how a word is phonetically announced. Certain words may not get announced correctly by a screen reader, and you may have a temptation to try and fix this. You may end up fixing how a word is announced by a screen reader, but that "fix" could end up making no sense when announced by other assistive technologies like a [braille](https://en.wikipedia.org/wiki/Braille) reader.
```
Putting more emphasis on this would be better since it is a common mistake.

Eveieve: Agreed, thanks!

advanced_html_css/accessibility/wai_aria.md
```diff
@@ -58,29 +58,29 @@ When using such ARIA attributes, you would provide an `id` to one element, and y
 
 #### `aria-label`
 
-The `aria-label` attribute overrides any native label and modifies the name property in the accessibility tree, though it's best used when an element doesn't already have a native label. When you add `aria-label` to an element, you pass in a string as the value, which will become that elements accessible name. `aria-label` doesn't work on every HTML element, though. Adding the attribute to a plain `<div>` or a `<span>` will have no effect, for example.
+The `aria-label` attribute overrides any native label of an element and modifies its name property in the accessibility tree. It is best used when an element doesn't already have a native label. When added, the string value of `aria-label` attribute  becomes the element's accessible name. **However**, [`aria-label` does not have any effect on some HTML elements](https://github.com/w3c/aria/issues/756), such as `<div>` or a `<span>`.
```
thatblindgeye, line 61:
```suggestion
The `aria-label` attribute overrides any native label of an element and modifies its name property in the accessibility tree. It is best used when an element doesn't already have a native label. When added, the string value of the `aria-label` attribute  becomes the element's accessible name. **However**, [`aria-label` does not have any effect on some HTML elements](https://github.com/w3c/aria/issues/756), such as `<div>` or a `<span>`. 
```

For the last sentence, I think this could be a good use-case for our note boxes, but not really a blocker. @Lofty-Brambles I'm not entirely sure about linking to the W3C issue here... Also not really a blocker, but could we instead just mention the same info ("'aria-label' may not have any - or the intended - effect on certain HTML elements, such as a `div` or `span` without a `role` attribute.")?

advanced_html_css/accessibility/wai_aria.md
```diff
@@ -58,29 +58,29 @@ When using such ARIA attributes, you would provide an `id` to one element, and y
...
 
-A common use for this attribute can be for the "close" button often seen in menus or modals:
+A common use for `aria-label` can be found in the "close" buttons of menus or modals:
 
 ~~~html
 <button type='button' aria-label='Close menu'>X</button>
 ~~~
 
-Instead of a screen reader announcing, "X, button", which wouldn't make any sense to the user, it would announce, "Close menu, button". Another way you could use `aria-label` is on landmark elements (our Semantic HTML lesson gets another shoutout... again!):
+Instead of a screen reader announcing, "X, button", which makes little sense to the user, it would announce, "Close menu, button". Another way you could use `aria-label` is on landmark elements (our Semantic HTML lesson gets another shoutout... again!):
```
thatblindgeye, line 69: Let's add a link:

```suggestion
Instead of a screen reader announcing, "X, button", which makes little sense to the user, it would announce, "Close menu, button". Another way you could use `aria-label` is on landmark elements (our [Semantic HTML](https://www.theodinproject.com/lessons/node-path-advanced-html-and-css-semantic-html) lesson gets another shoutout... again!):
```

advanced_html_css/accessibility/wai_aria.md
```diff
@@ -159,8 +159,8 @@ This section contains questions for you to check your understanding of this less
 
 This section contains helpful links to other content. It isn’t required, so consider it supplemental.
 
-* [An in-depth guide to ARIA roles](https://www.a11yproject.com/posts/2020-08-20-an-indepth-guide-to-aria-roles/#landmark-roles) goes over an ARIA attribute that alters the semantics of elements, which can allow you to create custom inputs, buttons, and more. Just keep in mind rule number one of the ARIA rules!
+* [An in-depth guide to ARIA roles](https://www.a11yproject.com/posts/2020-08-20-an-indepth-guide-to-aria-roles/#landmark-roles) goes over an ARIA attribute that alters the semantics of elements, which can allow you to create custom inputs, buttons, and more. Just keep in mind that rule number one of the ARIA rules!
```
thatblindgeye, line 162:
```suggestion
* [An in-depth guide to ARIA roles](https://www.a11yproject.com/posts/2020-08-20-an-indepth-guide-to-aria-roles/#landmark-roles) goes over an ARIA attribute that alters the semantics of elements, which can allow you to create custom inputs, buttons, and more. Just keep in mind rule number one of the ARIA rules!
```

Not sure "that" is really necessary here.

thatblindgeye (changes requested):

advanced_html_css/accessibility/wai_aria.md
Eveieve: That's my mistake.

Eveieve:
Hi @thatblindgeye, I wanted to ask, when you make a code suggestion, should I wait for other maintainers/contributors to review it or can I just `commit suggestion` if I agree with your suggestion?

advanced_html_css/accessibility/wai_aria.md
Lofty-Brambles: Sure thing! I found the conversation interesting, since it came from the very same confusion, which's why I linked it here. Not a blocker if you think it is not important. @thatblindgeye

thatblindgeye: I think my main worry is whether it will bog users down on a conversation that they may not understand the context of (could just be an invalid worry 😆 ). I don't really feel super strong either way; worse case is we keep the link and if any users start questioning it we can replace it in the future.

thatblindgeye:
@Eveieve Good question! Usually I try to say something like "would like feedback from so and so" to directly ask for more input, though sometimes it can be kind of in the air. I'd say if you agree with it and it doesn't seem readily obvious that more input is needed, go ahead and make suggested changes.

Worst case scenario is you end up needing to undo that change/make another update

advanced_html_css/accessibility/wai_aria.md
Eveieve: I do not feel so strongly about whether to keep the link or not either. I think the learners will know not to spend too much time understanding the reasoning behind this. If they *want* to dive deeper into the issue, they've got their source.

Eveieve:
@thatblindgeye Thanks for answering my question! I'm fairly new to contributing so wanted to make sure I'm doing it in the way it's supposed to be done.

thatblindgeye (approved):
Looks good! Thanks for the updates on these lessons 💪🏼

PR #25976, opened 2023-07-27
 Express 101: delete repeated text

hebiscus:
## Because

Paragraph is a repetition of paragraph above.

## This PR

- delete line 25, that repeats information on line 21

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/nodeJS/express_and_mongoose/express_101.md
+++ b/nodeJS/express_and_mongoose/express_101.md
@@ -22,8 +22,6 @@ This step of the MDN tutorial mentions middleware, but does not clearly define i
 
 For example, you might have a logger (that prints details of the request to the console), an authenticator (that checks to see if the user is logged in, or otherwise has permission to access whatever they're requesting) and a static-file server (if the user is requesting a static file then it will send it to them). All of these functions will be called in the order you specify every time there's a request on the way to your `app.get("/")` function.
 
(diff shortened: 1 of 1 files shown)


wise-king-sullyman:
I might not be understanding what you mean, but it doesn't seem that this paragraph is a repeat of the one on line 21 to me.

They share the "middleware are plain functions" bit, which I agree could be removed, but the second paragraph seems to be focused on that:
1. it's common to write your own middleware functions
2. they have a specific signature

Which are things not mentioned in the line 21 paragraph.

hebiscus:
@wise-king-sullyman Agreed, deletion was too hasty. What issue I had with that is it seemed to me like 2 ways of describing a middleware function, because of "let’s take a minute to demystify what they’re actually doing" bit and the common part between those too definitions. Would it be better to somehow combine the two instead? 
Let me know what you think and then I'll create an issue.

codyloyd:
The reason we've repeated the phrase "middleware are plain functions" is because it's important, and easily overlooked.

Students hear the word "middleware" and imagine something more advanced than "just a plain javascript function"

We want to emphasize that they're nothing special, and that writing your own is not difficult, or complicated.  _maybe_ we could smooth out the wording, but I don't even think that's necessary.

> it seemed to me like 2 ways of describing a middleware function

that is correct, and intentional.

PR #25981, opened 2023-07-27
Added explanation of blocks/yields to enumerable project.

### LINKED ISSUE #23481: Add a custom enumerable example to the custom enumerable project for the ruby track.
One of the things I noticed while helping in the chat is that it seems a few people seem to have difficulty understanding how to use blocks and yield. I believe it may be useful to provide a detailed example in the custom enumerable projects as an example showing both the ```do...end``` and ``` {|a| does something with a }```. An example of that could be the sort enumerable in either the custom enumerable project page itself OR as the first example in the enumerable repo for learners to go over and refresh themselves on what a block is, where you could put yield within that block, and a walkthrough of how that is handled. 

I am also going to link this other issue since I believe that splitting this into two lessons/add exercises would help with making blocks more clear. #23482

linkonsat:
## This PR

Adds an example to the custom enumerable project to explain and refresh learners' memories on this topic before the custom enumerable projects. As folks in the project seemed to have a bit of hard time understanding how to use yields and blocks.

## Issue

Closes #23481
## Pull Request Requirements


ruby/advanced_ruby/project_custom_enumerables.md
```diff
@@ -1,6 +1,36 @@
 ### Introduction
 You should be very familiar with the Enumerable module that gets mixed into the Array and Hash classes (among others) and provides you with lots of handy iterator methods. To prove that there's no magic to it, you're going to rebuild those methods.
 
+We will explain an example of how an enumerable works to give you an idea of how to start tackling these methods. Specifically were going to breakdown the `sort` enumerable method. The sort method sorts the elements by comparing their values based on whether one elment is bigger, equal or smaller than the next element as shown below.
```
zachmmeyer, line 4: GitHub found a typo here.
```suggestion
We will explain an example of how an enumerable works to give you an idea of how to start tackling these methods. Specifically were going to breakdown the `sort` enumerable method. The sort method sorts the elements by comparing their values based on whether one element is bigger, equal or smaller than the next element as shown below. 
```

zachmmeyer:
Reviewing this now, I missed this notification and it got cleared out.

ruby/advanced_ruby/project_custom_enumerables.md
```diff
@@ -1,6 +1,36 @@
...
+    ary
+  end
+end
+
+~~~
+
+ary in this case would be the array that we use to put in our results from building our map method. Self in this case is _calling the method on itself_ which is how were able to do `.my_map` without having to pass the array directly through the method parameters. Now that we that self refers to the array we can use the `each` method to pass each element. Now this is where yield becomes extremely useful. In our very first example of using `a.map` we used a special tool called a block. The block in this case is the `a.my_map {}` and this is where the values our changed. `yield` in this case specifies _what_ is passed, in this case the elem. Now that elem has been yielded it's accessible through the `{}` brackets and we can perform an operation on it. In this case adding two and therefore we use the shovel method to add that element to the array that has been yield to the block/`{}`. To recap briefly, the element is given to the `yield` through the elem or can be done by just putting something in the `{}` such as `puts 5` and the `{}` brackets is where the operation is performed. In essence the block/`{}` is where the `yield` operation is performed.
```
zachmmeyer, line 33: It's difficult to see what I modified here.
- our to have
- modified punctuation
- yield to yielded

I'd throw it in a real diff checker, it should show the changes to see if you agree with them.
```suggestion
ary in this case would be the array that we use to put in our results from building our map method. Self in this case is _calling the method on itself_ which is how were able to do `.my_map` without having to pass the array directly through the method parameters. Now that we that self refers to the array we can use the `each` method to pass each element. Now this is where yield becomes extremely useful. In our very first example of using `a.map` we used a special tool called a block. The block in this case is the `a.my_map {}` and this is where the values have changed. `yield` in this case specifies _what_ is passed, in this case the elem. Now that elem has been yielded it's accessible through the `{}` brackets and we can perform an operation on it. In this case: adding two. Therefore, we use the shovel method to add that element to the array that has been yielded to the block/`{}`. To recap briefly, the element is given to the `yield` through the elem or can be done by just putting something in the `{}` such as `puts 5` and the `{}` brackets is where the operation is performed. In essence the block/`{}` is where the `yield` operation is performed.
```

zachmmeyer (changes requested):

JoshDevHub:
I think this is an incredible idea, and I've been giving an example for a while now when talking to learners in the discord who are struggling with this exercise.

I will say that I think `sort` isn't ideal for demonstrating because it has a complicated implementation. The logic underpinning the method is the [quicksort algorithm](https://en.wikipedia.org/wiki/Quicksort). `map` also isn't great in my opinion because it's a part of the project itself. It'd be giving learners the answer to one of the problems.

I like using `#find`. It's pretty simple, and it's not part of the custom enumerables project
```rb
def my_find
  self.each do |elem|
    # if the element meets the block's conditions, return it
    return elem if yield(elem)
  end
  # if we're here, we looped the items without finding an element that meets the condition
  # return `nil`
  nil
end
```

zachmmeyer:
@JoshDevHub Is there anything you could add to this PR to improve upon it? Or would you be open to submitting something of your own?

JoshDevHub:
Hey @zachmmeyer 

Thank you for reminding me about this. I had forgotten about it completely.

My personal preference is to use something that isn't `#map` for demonstration. This is because `#map` is a part of the project, so it's essentially showing the learner how to solve one of the problems. But maybe that isn't such a huge deal.

In discord when helping people with this, I've often used `#find`, which is pretty straightforward and not part of the project. If you want, I could show an example of what that would look like.

zachmmeyer:
@JoshDevHub I think your reasoning is sound. Would you be up for opening a PR that uses `#find` instead? Then we could close this one down.

JoshDevHub:
@zachmmeyer Yeah, I'll whip something up :thumbsup:

JoshDevHub:
@zachmmeyer @linkonsat Okay I put in a PR for what I was thinking: #26915 

Let me know if y'all have any thoughts.

PR #26093, opened 2023-08-10
Update time_complexity.md (#1)

damon314159:
Time Complexity Lesson: Add detail to O(N log N) explanation

## Because
The section regarding O(N log N) mentions merge sort in a way that could imply all N log N complexities are of the format of an algorithm of O(log N) steps, with a nested algorithm of O(N) steps. While this is true for the merge sort, is it not the only way in which O(N log N) can arise.
The inverse could also occur, where you would have an algorithm of O(N) steps, with a nested algorithm of O(log N) steps. Or even an algorithm that doesn't use such a "divide into two and conquer" approach, which may have an N log N complexity without nesting algorithms in such a neatly explainable way.
The example I have chosen to highlight this is the algorithm for constructing a Cartesian tree from a given array. In short it works as follows
- Initialize an empty stack and an empty tree.
- Traverse the input array from left to right.
- For each element in the array, pop elements from the stack until the top element is larger than the current element. Each popped element becomes the right child of the previous element.
- Push the current element onto the stack.
It leverages the properties of a stack and a binary tree to achieve this complexity compared to the clear-cut nature of merge sorts two distinct component algorithms.

I think it is important to note, even very briefly, that there are other ways this sort of complexity can arise, as the current lesson does not paint a complete picture.

## This PR
- Add extra detail to the subsection regarding another example of an algorithm with such a complexity, but achieved in a different way.

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/time_complexity.md
+++ b/javascript/computer_science/time_complexity.md
@@ -187,6 +187,8 @@ You can't say this one isn't appropriately named. This notation means we have an
 
 One such algorithm is merge sort, and it just so happens you tackle this project in our course :)
 
+However, not all `O(N log N)` situations are built this way. There are special cases, like constructing a Cartesian tree, where the algorithm naturally behaves like `O(N log N)` without using smaller parts with `O(N)` or `O(log N)` complexities inside. This shows that while nested complexities can be common, they're not the only way an algorithm can achieve a particular time complexity.
+
 #### O(n&#178;) - Quadratic complexity
 
 You've probably written code with a Quadratic Complexity on your programming journey. It's commonly seen when you loop over a data set and within each loop you loop over it again.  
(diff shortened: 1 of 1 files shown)


codyloyd:
This seems legit to me, but it is outside of my area of expertise.  tagging @CouchofTomato who might be able to confirm that this is a disclaimer we want to add here.

CouchofTomato:
@damon314159 

Thanks for the edit. Agree we need to clarify here.

We do try to avoid introducing concepts that are never explained so in order to accept this how about we do the following.

1) We can make a very small change to the first paragraph. In the second sentence we can state that This notation *usually* implies. That will lead better into your edit.
2) Can you change where you write Cartesian tree to be a link to a source which describes what it is. Just so learners can take a peek to get a basic understanding if they wish.

damon314159:
I have made the requested changes and clarified that the article is only if they're curious. I hope I did that right, but let me know if not. Its my first PR! :)

CouchofTomato (approved):
Thanks @damon314159

PR #26140, opened 2023-08-16
JS ES6 modules/webpack - improve intro to bundlers

### LINKED ISSUE #26104: JS: Improve explanation on the point of Webpack/bundlers
### Describe your suggestion

I remember when going through the Webpack lessons/tutorials that I was *very* stuck on understand the point of using webpack (especially with all the configuration needed), especially when the official webpack tutorials linked as assignments, in my honest opinion, do not do the greatest job of explaining the point/benefits in using a bundler in general vs no bundler at all. They feel like they focus more on the "what" with simply telling you what code/config to achieve specific things that still leave you with more questions on "why would this ever be useful".

Evidently from multiple discord conversations, people have felt the same (both going through it and looking back at it). A big example of this is multiple entry points - something that especially nowadays with smallers SPAs (which is what most if not all of the projects for the rest of the JS course are) you may not need - I've seen plenty get so confused as to what this is for/why you'd use multiple entry points and start trying to do so for their restaurant page/todo list which understandably confuses them further since they shouldn't need it. The official tutorial on that only briefly mentions the use of it including a short link to code splitting at the very top of the page.

I feel the section on webpack in the ES6 modules lesson in particular could greatly benefit from a little more explicit explanation into why bundlers are used and some simple reasons why someone would benefit from using one over not.
Something even like a short explanation about the server needing to download multiple JS module files as well as any dependencies individually, and how a bundler can bundle everything into a single file which is more performant to download *and* includes all dependencies for that bundle.

One such resource that was of great help to me was: https://snipcart.com/blog/javascript-module-bundler
Simply to explain and demonstrate the difference between using a bundler and not.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-javascript-es6-modules and possibly also https://www.theodinproject.com/lessons/node-path-javascript-webpack

### Checks

### (Optional) Discord Name

Mao

_No response_

MaoShizhong:
## Because
Introduce and explain the point behind bundlers in general before then going into webpack specifically.
A common source of confusion (including my own experience at the time) around the webpack lessons is that people often don't understand why they're using it in the first place, especially with all of the configuration options thrown in their face. The official website's tutorials, in my opinion, also leave a lot to be desired as they very well explain the *what* but in my opinion, not so much the *why*.
I felt a lot of "you can use webpack to do bundle your assets like fonts and images! You can have multiple entry points!" but any "why" explanations often led to more rabbit holes like a simple link to their tutorial on code-splitting, which still isn't the most helpful if you don't really understand why you're using it in the first place.

I feel the issue can at least start being addressed by explaining what bundlers are in general and why they benefit us, and *then* going into introducing webpack as one such bundler. I believe this will help "set the scene" and help make the next lesson's tutorials go more along the lines of examples of ways the bundler can help us achieve the benefits introduced already.

## This PR
- Add intro paragraphs to the introduction to webpack in the ES6 modules lesson, introducing and explaining bundlers and why we would benefit from using them.
- Edit the last paragraph of the section to refer to the new content above.

## Issue
Closes #26104

## Additional Information
N/A

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/es6_modules.md
+++ b/javascript/organizing_your_javascript_code/es6_modules.md
@@ -43,9 +43,17 @@ Read through the npm links below but don't worry about running any of the comman
 
 At some point, you will probably run into [Yarn](https://yarnpkg.com/en/) - a replacement for the default `npm`. For the most part, it does the same things, though it _does_ have a few more features. Recent versions of `npm` have incorporated some of the best features of Yarn, so using it won't offer you any real advantages at this point in your career. It _is_ a fine package manager, however, and may be worth your consideration in the future.
 
-### Webpack
+### Webpack and bundlers
 
-Webpack is simply a tool for bundling modules. There is a lot of talk across the net about how difficult and complex it is to set up and use, but at the moment our needs are few and the setup is simple enough. In fact, you can see an example of getting it up and running on the front page of [their website](https://webpack.js.org/).
+So far, your projects have had relatively simple file structures. As project complexity grows, so too will the benefits of well-organized code. A project consisting of a single, long file with lots of code can be made easier to navigate and maintain by being broken down into multiple smaller files (modules). Further benefits of writing code in modules will come below when we introduce ES6 modules.
+
+But there's a problem! The browser would need to make a separate HTTP request for each file. The more files you have, the more costly this becomes, particularly on slower networks and would only increase if you also imported third-party libraries into your app.
+
+What if we had a way to write multiple files and/or import multiple third-party libraries but eventually combine them all into fewer files at the end so the browser did not have to make so many requests?
+
+Enter bundlers. Give a bundler a starting file (an entry point) and it will build a dependency graph of the modules and dependencies starting from that file, then combine them into a single output file. Provide multiple entry points and it will do the same for each separately. You can read more about why we use bundlers and how they work in [this short article](https://snipcart.com/blog/javascript-module-bundler).
+
+Webpack is one such tool for bundling modules. There is a lot of talk across the net about how difficult and complex it is to set up and use, but at the moment our needs are few and the setup is simple enough. In fact, you can see an example of getting it up and running on the front page of [their website](https://webpack.js.org/).
 
 Webpack _is_ a very powerful tool, and with that power comes a decent amount of complexity. Don't let it scare you off! The basic configuration is not difficult and proficiency with webpack looks _amazing_ on resumes.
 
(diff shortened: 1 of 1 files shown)


MaoShizhong:
In a recent [Discord discussion (inc. messages before the linked message)](https://discord.com/channels/505093832157691914/505093832157691916/1143731198296866917), further specific issues with the webpack lessons/tutorials were brought up.

In the [actual Webpack lesson](https://www.theodinproject.com/lessons/node-path-javascript-webpack), assignment step 2 indicates the lesson is simply to learn about `HtmlWebpackPlugin` and how it can create/insert HTML files for you and that's it. Yet, the lesson starts off talking about multiple entry points (and poorly explained too, regarding the practical uses/"why" part) - something which I cannot see how it would be relevant to any of the course's projects using Webpack. I've seen people unnecessarily use them, not knowing why they're using them/why they should only need a single entry point which ended up leading to a bunch of bad.

The tutorial introduces `HtmlWebpackPlugin` but it and subsequent tutorials teach you to work with an index.html file in `dist/` and not `src/`. I've seen people then have projects where they work with html and assets inside `dist/` but work with css/js inside `src/` whereas ideally, they do *everything* within `src/` and simply build for production which goes into `dist/`.
HtmlWebpackPlugin can easily faciliate this with the simple one-line template option in `webpack.config.js`, with a filepath to an html file in `src/` which it will then use to build the dist html file for production. This is a very simple step that will help people more easily understand the workflow and organisation, and udnerstand how to keep their static html without HtmlWebpackPlugin overwriting it in dist when building.

I would propose making some sort of change to step 2 of the webpack lesson's assignment. I'm not entirely sure how to go about this yet but I feel the current tutorial is not appropriate. Even if you skip the multiple entry points part, the HtmlWebpackPlugin section continues off the previous part, and it doesn't mention anything about templates which I honestly feel is the most useful part of HtmlWebpackPlugin.

Thoughts?

MaoShizhong:
[The webpack page on HtmlWebpackPlugin](https://webpack.js.org/plugins/html-webpack-plugin/) is a bit short and isn't really a tutorial of any kind. [The plugin's repo README](https://webpack.js.org/plugins/html-webpack-plugin/) does explain better (especially template) but the useful stuff is a small drop in an ocean of vastly out-of-scope stuff, and I can't really find any good brief tutorials that would fall into "output management" like the original but are presented more appropriately without going into out-of-scope topics that only serve to confuse people at this point in the course (especially with stuff that will not be remotely relevant for the course material).

bycdiaz (approved):
This context all feels helpful. And I think any other additions can be raised in a different PR/issue.

If you wanna send this in as is, let me know so I can merge.

MaoShizhong:
@bycdiaz I'm happy to send this in as is and I will raise a separate issue for the other stuff. Many thanks!

OmarNaguib:
Excellent addition @MaoShizhong !

PR #26230, opened 2023-08-27
Update predicate_enumerable_methods.md: #none is the opposit of #any, not of #all

dimbyr:
I think #none? is the opposit of #any?, not of #all?. Indeed, "none" means "there is nothing", "any" means there are something, and "all" means everything is there.

## Issue

Closes #XXXXX

## Pull Request Requirements


ruby/basic_ruby/predicate_enumerable_methods.md
```diff
@@ -164,7 +164,7 @@ fruits.all? { |fruit| fruit.length > 6 }
 Special note to keep in mind while debugging: `#all?` will return `true` by default unless the block returns `false` or `nil`. So if you call `#all?` on an empty array or hash (i.e., there are no elements  for the block to evaluate), it will return `true`.
 
 ### The none? method
-As you might expect, `#none?` performs the opposite function of `#all?`. It returns `true` only if the condition in the block matches *none* of the elements in your array or hash; otherwise, it returns `false`.
+As you might expect, `#none?` performs the opposite function of `#any?`. It returns `true` only if the condition in the block matches *none* of the elements in your array or hash; otherwise, it returns `false`.
```
KevinMulhern, line 167: Thanks for taking the time to contribute @dimbyr. But I don't think this is correct, functionally `none` is closer to being the opposite of `all` in ruby.

To put them another way:
`#all?` - returns true if all of the elements match the condition in the block
`#none?` - returns true if **all** the elements do not match the condition in the block

dimbyr: @KevinMulhern 

The negation  of `#all`, which I refer to as `not #all`,  returns true if **some** of the elements **do not** satisfy the condition in the block. However  it is possible that some of them **do satisfy** the condition. 

The following points show that `#any` is the opposit of `#none`, i.e, these two methods can never return the same value for the same array.  

1. If `#none` returns `false`, this means that there is at least one element that satisfy the property. In this case, `#any` returns `true`.  
2. One can also see that if `#any` returns `true`, then `#none` will return `false`.
3. If `#none` returns `true`, this means that there is no element satisfying the property. In this case, `#any` returns `false`. 
4. Similarly, if `#any` returns `true`, then `#none` will return `false`.

rlmoser99: I see both sides and wonder if we could clear up the confusion by just getting rid of the "opposite" explanation. 
```suggestion
As you might expect, `#none?` returns `true` only if the condition in the block matches *none* of the elements in your array or hash; otherwise, it returns `false`.
```

KevinMulhern: Thanks for the detailed response @dimbyr, I can see where you're coming from better now. 

Semantically both any and all are the opposite of none so I think what @rlmoser99 is suggesting is the best way forward.

dimbyr: @KevinMulhern ,

Sorry, I wanted to clarify that `#none?` is NOT the opposite of `#all?`. 

Of course `#none?` returns `false` when `#all?` returns `true`, BUT one does not need `#all?` to return `true` for `#none?` to return `false`.

So, the two sentences on `line 167` is correct IF AND ONLY IF one replace `#all?` with `#any?`

Thanks.

KevinMulhern: Screenshot 2023-08-30 at 06 58 43

KevinMulhern: I think w might be splitting hairs at this point, both all and any are the opposite of none **semantically**. To get this PR merged can we remove the opposite comparison please?

dimbyr: Yes, I agree, they are semantically opposite, but in our context, the two methods do not always perform actions that are opposite to each other. As I have mentioned, they can return `false` at the same time.

dimbyr: I think the word "some"  in the list of opposites in the dictionary is the one that translates better the opposite of `#all?` in our context.

dimbyr: @KevinMulhern,
Sorry, maybe I was not clear enough. 
My point is that the statement " `#none?` performs the opposite function of `#all?`. " is not correct. I do not mean the words "all" and "none" are not semantically opposite.

KevinMulhern: Thats a fair point and I appreciate where you're coming from. I think we've maybe got a different view on what the context of this is. You're coming at it from a mathematics standpoint and I'm coming at it from what intention you'll want to communicate through the code.

If we both disagree, then theres a good chance students will be confused with this too. I think doing as Rlmoser suggested by removing this opposite comparison entirely is the best compromise and course of action.

CouchofTomato:
Hey @dimbyr 

I think you're both correct here and that's because technically `none?` could be the opposite of `any?` or `all?` as it really depends what you're trying to express in your code. This is why all 3 methods exist.

So, yes you are correct in that `none?` isn't strictly the opposite of `all?` and is therefore misleading. This is why we've compromised with a solution to just remove the `opposite` line. If you can make that change we're happy to merge. If not then we need to close this out please.

CouchofTomato:
Thanks @dimbyr 

We appreciate the contribution and the discussion was pretty interesting. Glad we could close this one out favourably.

CouchofTomato (approved):

PR #26267, opened 2023-09-01
Javascript & Ruby / Knights Travails instructions

Mclilzee:
## Because

A proper BFS or DFS algorithms done correctly shouldn't get stuck in an infinite loop traversing the same nodes that were already visited. Unless I'm missing the point of this hint, I thought I would be best if we remove it.

## This PR

- Removes the hint information about getting stuck in an infinite series.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_knights_travails.md
+++ b/javascript/computer_science/project_knights_travails.md
@@ -22,7 +22,7 @@ You can think of the board as having 2-dimensional coordinates.  Your function w
 <div class="lesson-content__panel" markdown="1">
 1. Put together a script that creates a game board and a knight.
 2. Treat all possible moves the knight could make as children in a tree.  Don't allow any moves to go off the board.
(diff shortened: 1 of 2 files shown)


CouchofTomato:
The hint is mostly to point people away from DFS. Since the point is to find the shortest path, DFS is going to be tricky to avoid an infinite loop. You can't know if you've traversed the path before until you reach the node you're looking for so the code for it would be very tricky to implement.

Therefore I'm not inclined to change this as is, but would be willing to make the change if you edit to say something like "One of them could lead to an infinite loop if not coded correctly" or something.

Mclilzee:
Understandable, but I wouldn't choose to change it since the word "correctly" is not 100% true, some implementations of DFS does not save visited nodes to save memory, especially on nodes that are just one way connected. Another problem would be that we potentially informing students that shortest path can be achieved with DFS.

I will leave it as it originally was and close this PR.

PR #26291, opened 2023-09-06
Update fundamentals-1.md

AnityaSharma:
The value I got was 0.82191, I know its not that big issue but the value is varing by 0.1.


rlmoser99:
Thanks for taking the time to make this PR. However, I do not believe your calculations are correct. Here is the screenshot of my calculations:



In addition, you are required to fill out the PR template that you deleted when you created this PR. We often close PRs when they do not have this vital information. This is covered in our [https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md](contributing guide), which is a link provided in the PR template.

PR #26459, opened 2023-10-09
Update flexbox-growing-and-shrinking.md

sapienfrom2000s:
## Because
flex-shrink covers the point what happens if parent container is smaller than the sum of flex items. flex grow doesn't cover the complementary case of what happens if parent container is larger than the sum of flex items.

## This PR
Adds content to the lesson.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
+++ b/foundations/html_css/flexbox/flexbox-growing-and-shrinking.md
@@ -28,6 +28,8 @@ Very often you see the flex shorthand defined with only _one_ value. In that cas
 
 `flex-grow` expects a single number as its value, and that number is used as the flex-item's "growth factor". When we applied `flex: 1` to every div inside our container, we were telling every div to grow the same amount. The result of this is that every div ends up the exact same size. If we instead add `flex: 2` to just one of the divs, then that div would grow to 2x the size of the others.
 
+"grow-factor" is also applied if size of all flex items is smaller than their parent container. If after resizing parent container grows, the growth of flex items will depend on `flex-grow` number.
+
 In the following example the `flex` shorthand has values for `flex-shrink` and `flex-basis` specified with their default values.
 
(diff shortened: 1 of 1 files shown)


codyloyd:
I like this inclusion.. but the wording is awkward.  Specifically, this sentence:

> If after resizing parent container grows

can you clean it up?

sapienfrom2000s:
@codyloyd I have made a new commit. Are you okay with it?

codyloyd:
Unfortunately, I think what you've written now has ended up being more confusing.  Consider something like this:



> When the flex-container's main size is larger than the combined main sizes of the flex items, the extra space is distributed among the flex items, with each item growth being their growth factor value as a proportion of the sum total of all the container's items' flex grow factors.

> For example, if there are two items in a flex container, and they both have `flex-grow: 1`, then they will be the same size.  Likewise, if they both have `flex-grow: 2` or `flex-grow: 1000`, then they will be the same size because the proportions are the same.

If, however, one of the items has `flex-grow: 2` and the other has `flex-grow: 1`, then the sizes will be different.  The size of each element will be calculated as a fraction of the total of the various growth-factors among all children.  In this case, we have a `2` and a `1`, so the item with `flex-grow: 2` will be `2/3` the width of the total, and the item with `flex-grow: 1` will be `1/3` of the total.

> This can be difficult to calculate in more complex situations, and it is usually not something you will want to rely on to precisely size specific flex-children.  In real-world usage, the most common situation **by far** is simply using `flex-grow: 1` to make a specific child grow to fill a parent or using `flex-grow: 1` on all children to size them equally.

> For example, if you are using flexbox to position a sidebar on a webpage, you will likely have two main flex-children, the sidebar and the main-content area.  Do not try to use extreme flex-grow proportions to size your sidebar.  Rather, just make the sidebar a static width (200px-300px, or whatever seems logical in your use case) and then put `flex-grow: 1` on the main-content div to make it grow to fill the remaining area.

sapienfrom2000s:
_The "grow factor" also applies when the combined size of all flex items is smaller than their parent container. In that case, the distribution of remaining space among flex items is proportional to the ratio of their flex-grow numbers._
@codyloyd , How about this?

PR #26518, opened 2023-10-18
Computer Science: HashMap Project

Mclilzee:
## Because

A Hash Map or Hash Table is one of the most used data structures in programming to this day. I have created a draft lesson that talks about it in the computer science section.

I personally believe that it is one of the most important Computer Science data structures that people need to learn about.

In Ruby, we talk about how to use them with a lesson called [Hashes](https://www.theodinproject.com/lessons/ruby-hashes) From glancing over this lesson, I don't think it talks about how it works or how is it implemented (correct me if I'm wrong) so this lesson can also go to Ruby CS if it was found a good fit for TOP.

## This PR

Adds a Hash Map lesson

## Issue

Closes #XXXXX

## Pull Request Requirements


KevinMulhern:
Thanks for doing this @Mclilzee the content and illustrations look amazing 🔥 

Would you like someone to be assigned for review now? or is there more to be done before it's ready and out of draft?

Mclilzee:
@KevinMulhern Thanks for the nice words.

I didn't want to put too much work into it before I get the green light from the team. Work is still needed, like some resources some links to explain concepts for those who are still struggling and at the end section there will be a project to build the class itself and be able to have an official Map that students will build themselves from the ground up will be an amazing feeling.

JoshDevHub:
Yeah I'd be glad to help translate it to Ruby. Also I suspect the main party you want buy-in from is the DSA team. I'll tag them. @TheOdinProject/ds-a

01zulfi (commented):
A hashmap lesson is definitely on the roadmap. @Mclilzee great work and the lesson already lookin 🔥 

Before further work is put in the lesson, I'd like for us to have an outline of the lesson. This will be really helpful when the lesson gets reviewed.

From an initial glance, the lesson covers:

1. hash functions
1. hash collisions
1. hash table/map data structure
1. resizing hashmaps 

Please post your version of the lesson outline. Will it cover more concepts than the above mentioned ones?

Mclilzee:
@01zulfi I think this should be enough coverage to create a hash map class project. The project will cover creating a HashMap, adding methods like `set, get, has, keys, values, entries, clear, remove, length` as a bonus assignment to create a HashSet that only contains keys.

And to find some resources that have different perspective as well maybe a youtube video or two to put in the lesson for those who want to learn further about it.

Other than that, it needs cleaning up and to follow TOP style guideline.

01zulfi:
@Mclilzee Sounds good. Do mark the PR as "Ready for review" when you think it's ready, really looking forward to HashMap's addition to the curriculum

Mclilzee:
@01zulfi any updates on this?

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
+### Introduction
+
+One of the most used data structure across all languages is a Hash Table, aka Hash Map. If you worked before with JavaScript Object Literal `{}`, `Set`, and `Map` then you are already familiar with a hash table. But how do they work exactly, how can we save values using strings and symbols as keys and then retrieving those values using the same keys.
+
+In this project, you will learn how it all works, and you will implement your own Hash Map! To start we need to learn a few concepts, one of which is a hash code, what does it mean, and how to use it. Let us get started and hope you are exited to finally know what is happening with an object when you put keys and values in them.
```
JoshDevHub, line 5: There are a bunch of different ways "hash map" is written in this lesson. It appears as "hash map", "Hash Map", "HashMap", etc.

I think it'd be good to have a consistent approach. My preference would probably be "hash map" and then if we're referring to the class the user is making in the project, maybe `HashMap` can be fine too.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+Let us start by learning what does it mean to hash a value. Hashing means taking an input in and generating an output that represents this input. A Hash function should be a pure function, hashing the same input should always return the same hash code, there should be no random generation that is happening. For example, let's work a hashing function that takes a name and gives us the first letter of that name:
+
+```javascript
+function hash(name) {
+  return name.charAt(0);
+}
+```
+
+We created our first basic hashing function, it takes a name and returns the first letter of that name.
```
JoshDevHub, line 17:
```suggestion
We created our first basic hashing function. It takes a name and returns the first letter of that name.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+You might be familiar with this concept from your cipher exercise earlier in foundation, but there is a key difference between hashing and ciphering (encryption), which is reversibility.
```
JoshDevHub, line 18:
```suggestion
You might be familiar with this concept from your cipher exercise earlier in the foundations course, but there is a key difference between hashing and ciphering (encryption): reversibility.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+Hashing is a one-way process. Here is an example given a name, you can get a hash out of it, but given a hash, you cannot revert it back to a name. If you have a name `"Carlos"` we can hash it to `"C"`. But it will not be possible to reverse it from `"C"` back to its original form. You cannot know if it is `"Carlos"`, maybe it's `"Carla"` or `"Carrot"` we don't know. This is very good for security also, given a password you can save the hash of that password, but if someone steals your hashes they cannot know the original passwords since they are unable to reverse the hash back to the password. We are not going to dive deeper into hashing passwords, this is a subject for another lesson.
```
JoshDevHub, line 20:
```suggestion
Hashing is a one-way process. In the above example, you can get a hash code from a name, but when given a hash, you cannot revert it back to a name. If you have a name `"Carlos"` we can hash it to `"C"`. But it will not be possible to reverse it from `"C"` back to its original form. You cannot know if the original name was `"Carlos"`. Maybe it's `"Carla"` or `"Carrot"`. We can't know. This is very good for security. Given a password you can save the hash of that password, but if someone steals your hashes they cannot know the original passwords because they are unable to reverse the hash back to the password. We are not going to dive deeper into hashing passwords, and this is a subject for another lesson.
```
I also think the information about passwords might fit better as a note.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Use cases
+
+What can we do with those hashes, You have probably seen it in school where a folder is organized into smaller folders, each folder holds information about people with the same first letter, so we get something like this:
```
JoshDevHub, line 24:
```suggestion
What can we do with those hashes? You have probably seen it in school where a folder is organized into smaller folders, and each folder holds information about people with the same first letter, so we get something like this:
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+B:
+  bryan.txt
+  bob.txt
+  beatrice.txt
+  bella.txt
+  benjamin.txt
+  bianca.txt
+```
+
+If we get new student in our school with the name `"Carlos"` we can run our hash function to find out what folder do we assign them in `hash("Carlos") -> "C"` so we put `"Carlos"` in the directory labeled `C`
```
JoshDevHub, line 39:
```suggestion
If we get new student in our school with the name `"Carlos"`, we can run our hash function to find out which folder to place them in. `hash("Carlos") -> "C"` so we put `"Carlos"` in the directory labeled `C`
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+You might have spotted a problem, what if our school is populated with many people that share the same first letter `C`? Then we will have a directory with labeled `C` that have too many names that start with `"C"` and the other directories are empty. We need to rework our hash function, to eliminate more duplication and separate our students, while keeping constant structure in folders.
+
+```javascript
+function hash(name, surname) {
+  return name.charAt(0) + surname.charAt(0);
+}
+```
+
+Instead of just taking the first name letter, we take the first name and last name letters. `"Carlos Smith"` will have a hash code of `"CS"` this will surely spread our students in more directories, it will eliminate many duplicate hash codes from being generated.
```
JoshDevHub, line 49:
```suggestion
Instead of just taking the first name letter, we take the first name and last name letters. `"Carlos Smith"` will have a hash code of `"CS"`. This will spread our students among more directories and will eliminate many duplicate hash codes from being generated.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+  return hashCode;
+}
+
+function hash(name, surname) {
+  return stringToNumber(name) + stringToNumber(surname);
+}
+```
+
+Using this technique we don't only take the first letters into consideration, now we take the whole name and convert it into numbers.
```
JoshDevHub, line 70:
```suggestion
We not only consider the first letters with this technique. Instead, we take the entire name and convert it into numbers.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+You might be thinking, wouldn't it be just better to save the whole name as a hash code? That is true, this would make it unique for each name, but there are other reasons that we have chosen a number:
```
JoshDevHub, line 72:
```suggestion
You might be thinking, wouldn't it be just better to save the whole name as a hash code? That is true. This would make it unique for each name, but there are some good reasons to choose a number:
```

JoshDevHub (changes requested):
I really like this! I still need to review more things about it. There are still places where the English grammar could be cleaned up. I also want to try actually doing the project for myself to see how that aspect feels.

But also given your tag of Xari, I figured you'd be eager to make more progress on this. So I'm gonna go ahead and show the comments I have so far.

Most of them are English adjustments. Hopefully you don't feel about these. I get that it isn't your first language. But there are also a few comments that are about technical details.

I'm also wondering a bit about the structure of this project in the curriculum. This feels a bit like a lesson on hashing and hash tables pulled together with a project on making them. The curriculum doesn't often do this, and I wonder if it'd be a good idea to split the more "lesson"-type stuff up into a dedicated lesson and have the assignment section be the project. That's something I'd definitely be interested in getting @01zulfi thoughts on.

Mclilzee:
> Most of them are English adjustments. Hopefully you don't feel bad about these. I get that it isn't your first language. But there are also a few comments that are about technical details.
> 
> I'm also wondering a bit about the structure of this project in the curriculum. This feels a bit like a lesson on hashing and hash tables pulled together with a project on making them. The curriculum doesn't often do this, and I wonder if it'd be a good idea to split the more "lesson"-type stuff up into a dedicated lesson and have the assignment section be the project. That's something I'd definitely be interested in getting @01zulfi thoughts on.

I do not feel bad at all! This was tremendous help, I really appreciate it. As for splitting up the lesson, I totally agree, I felt this was too lengthy for a project and I even tried to cut corners as to not put too many details to make it more concise. I tried my best, but it still is lengthy.

01zulfi:
@Mclilzee Thanks for the reminder. I'll share my thoughts within the wekk

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,223 @@
...
+
+This is it, making this will result in a hash table with `search`, `set` and `get`.
+
+What if we found the hash code, but also the key value is the same as what we already have in the bucket. We check if it's the same item by comparing the keys, then we overwrite the value with our new value. This is how we can only have unique values inside a `Set`, `Set` is similar to a hash map but the key difference (pun intended) is that a `Set` will have Nodes with only keys and no values.
+
+<div class="lesson-note lesson-note--warning" markdown="1">
+A Hash Map does not guarantee insertion order, when you iterate over it. That is because of the buckets indexing nature of hash coding. That means if you are to retrieve the array of keys and values to iterate over them, then they will not be in order of when you inserted them.
+For example, if we insert the values `Mao`, `Zach`, `Xari` in this order, we may get back `["Zach", "Mao", "Xari"]`
+If iterating over the hash map frequently is your goal, then this data structure is not the right choice for the job, a simple array would be better.
+</div>
```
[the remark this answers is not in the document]
Mclilzee: I thought about adding reserving order as a bonus assignment, but not sure if that will deter people from the main concept of hash maps.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+      if (index < 0 || index >= buckets.length) {
+        throw new Error("Trying to access index out of bound");
+      }
+      ```
+
+  1. Inside the class DO NOT use Objects to store your key value pairs, And DO NOT use the array method `push` to insert your nodes into the buckets, always use indexes.
+
+#### Time to create a real hash map
+
+  1. Start by creating a HashMap class or factory function, up to you. Then proceed to create the following methods:
```
JoshDevHub, line 193:
```suggestion
  1. Start by creating a `HashMap` class or factory function. It's up to you which you want to use. Then proceed to create the following methods:
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `hash` takes a value and produces a hash code with it. There are a lot of thought that goes into creating the most optimal hashing function, it is up to you if you want to do a research to find a good hashing function but do not dive too deep into the rabbit hole. Make sure to test that it's not producing too many similar hash codes that will lead to too many collisions.
```
JoshDevHub, line 195:
```suggestion
  1. `hash` takes a value and produces a hash code with it. There is a lot of thought that goes into creating optimal hashing functions. It is up to you if you want to research to find a good hashing function, but don't dive too deep into the rabbit hole. Make sure to test that it's not producing too many similar hash codes. This will lead to frequent collisions when storing keys.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `set` takes two arguments, the first is a key and the second is a value that is assigned to this key.
+
+     - Remember to grow your buckets size when it needs to, by calculating the `load factor` and `capacity`.
```
JoshDevHub, line 199: I think another good reminder here is to think about what needs to happen when the given key already exists in the hash map.

This can be a bit of a balance. I'm sure some things we want them to "discover" while doing the project, but with this, I think it's entirely plausible that a learner does the project without testing or thinking about what happens when you `set()` with a key that's already there. So I think it'd be good to call attention to this case.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `get` takes one argument as a key and returns the value that is assigned to this key.
```
JoshDevHub, line 201: What should happen when the given key doesn't exist in the hash? When a function has a return value, we should be explicit about the different possible returns. Just like with the [Linked List assignment](https://www.theodinproject.com/lessons/javascript-linked-lists#assignment) for example

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `has` takes a key as an argument and checks if your hash map has that key.
+
+  1. `remove` takes a key as argument and remove it from the hash table.
```
JoshDevHub, line 205:
```suggestion
  1. `remove` takes a key as argument and removes it from the hash table.
```

JoshDevHub, line 205: Should `remove()` return anything? Usually it does return something, if for no other reason than to let the developer know if an item has been deleted or not.

[JavaScript Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete#return_value) returns `true` if the key was deleted and `false` if it wasn't.

In Ruby hashes, [the value for the deleted item is returned](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete#return_value) (or `nil` if the key wasn't found).

Either one of those approaches could make sense to me. Maybe with this JS project, we can design it like the JS `Map`?

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `length` returns how many stored keys in the hash code.
```
JoshDevHub, line 207:
```suggestion
  1. `length` returns the number of stored keys in the hash map.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `clear` removes all entries in the hash map.
+
+  1. `keys` returns an array containing all the keys inside the hash map.
+
+  1. `values` returns an array containing all the values.
+
+  1. `entries` returns an array that contains each a `key, value` pairs. Example: `[[firstKey, firstValue], [secondKey, secondValue]]`
```
JoshDevHub, line 215:
```suggestion
  1. `entries` returns an array that contains each `key, value` pair. Example: `[[firstKey, firstValue], [secondKey, secondValue]]`
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,227 @@
...
+
+  1. `hash` takes a value and produces a hash code with it. There are a lot of thought that goes into creating the most optimal hashing function, it is up to you if you want to do a research to find a good hashing function but do not dive too deep into the rabbit hole. Make sure to test that it's not producing too many similar hash codes that will lead to too many collisions.
+
+  1. `set` takes two arguments, the first is a key and the second is a value that is assigned to this key.
+
+     - Remember to grow your buckets size when it needs to, by calculating the `load factor` and `capacity`.
+
+  1. `get` takes one argument as a key and returns the value that is assigned to this key.
+
+  1. `has` takes a key as an argument and checks if your hash map has that key.
```
JoshDevHub, line 203:
```suggestion
  1. `has` takes a key as an argument and returns `true` or `false` based on whether or not the key is in the hash map.
```

JoshDevHub (changes requested):
Did the project. I like it! I think it'll be a nice opportunity for learners to tinker more with linked lists as well. Especially since the interface design for what this hash map needs is different than what is created in the Linked List project.

[Here's my code](https://github.com/JoshDevHub/JavaScript-HashMap/tree/main) if you want to see it. Hopefully my JavaScript doesn't read too weird :laughing: ... I don't write it very much at work. But yeah, mostly curious if that fits your vision of the project?

Another thing I think could be nice: in the [binary search tree assignment](https://www.theodinproject.com/lessons/javascript-binary-search-trees#assignment), if you scroll to the bottom of the assignment, you'll see the "Tie it all together" section. This provides learners with a sequence of steps they can put their `BinarySearchTree` through to make sure it's all working properly. I think this is cool to do because the Computer Science section occurs before testing in the curriculum. You could potentially do something similar with this project, which I think would be a nice convenience for the learners to make sure they everything functioning well without having to think of the manual test cases for themselves.

Let me know what you think! @Mclilzee

Mclilzee:
> Did the project. I like it! I think it'll be a nice opportunity for learners to tinker more with linked lists as well. Especially since the interface design for what this hash map needs is different than what is created in the Linked List project.
> 
> [Here's my code](https://github.com/JoshDevHub/JavaScript-HashMap/tree/main) if you want to see it. Hopefully my JavaScript doesn't read too weird 😆 ... I don't write it very much at work. But yeah, mostly curious if that fits your vision of the project?
> 
> Another thing I think could be nice: in the [binary search tree assignment](https://www.theodinproject.com/lessons/javascript-binary-search-trees#assignment), if you scroll to the bottom of the assignment, you'll see the "Tie it all together" section. This provides learners with a sequence of steps they can put their `BinarySearchTree` through to make sure it's all working properly. I think this is cool to do because the Computer Science section occurs before testing in the curriculum. You could potentially do something similar with this project, which I think would be a nice convenience for the learners to make sure they everything functioning well without having to think of the manual test cases for themselves.
> 
> Let me know what you think! @Mclilzee

I'm glad you like it! And really happy to hear you implementing it. I wanted to see if someone is willing to test it out and I'm glad of the outcome. Your Implementation of the hash map is really good! except for this part
```js
 #growCapacity() {
    const currentLength = this.length;
    this.length = currentLength;
  }
  ```
  But hey I'm not judging 👀

File: javascript/computer_science/project_hash_map.md [hunk 19]
```diff
@@ -0,0 +1,227 @@
...
+
+  1. When you create a function that takes a hash code value and returns the bucket index, use this snippet before the function returns to make sure it's accessing inbound array indexes only.
+
+      ```javascript
+      if (index < 0 || index >= buckets.length) {
+        throw new Error("Trying to access index out of bound");
+      }
+      ```
+
+  1. Inside the class DO NOT use Objects to store your key value pairs, And DO NOT use the array method `push` to insert your nodes into the buckets, always use indexes.
```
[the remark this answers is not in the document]
Mclilzee: > Also, is it okay to store the key/value pair as an object on the linked list nodes?

Since Node is not part of the hash map and using an object literal is not violating the rule, but I still feel that this takes the magic away. In the end of the day, using this will still make us give our key values to JS built in library to handle for us. That's what we're trying to avoid, and building our own key value mechanics.

javascript/computer_science/project_hash_map.md
Mclilzee: This is true, I don't think I have come across a language that returns nothing with its remove function. Good catch, I think returning true or false will suffice for the JS implementation, maybe it can be reworked in Ruby lesson to be more similar to the hash map in ruby where it returns `nil` if not found? That part is up to you

javascript/computer_science/project_hash_map.md
Mclilzee: I put it as part of the lesson that if a key already present, then its value is overwritten. I can add another small reminder here, like the one I put for growing the table.

javascript/computer_science/project_hash_map.md
Mclilzee: Can you go over this again, please? I don't feel this present the idea that if we have a function that produces similar keys, then it will lead to frequent collisions. It seems you separated the concerns. Or we can re-arrange it to talk about frequent collisions earlier and that we need to avoid them.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Notice the usage of prime number. We could have chosen any number we wanted, but prime numbers are even better to introduce less hash codes that are divisible by the same bucket length, which will make collisions less likely to happen.
+</div>
+
+Even though we reworked our hash function to avoid the `"Sara"/"raSa"` collision, there is always the possibility for collisions. Because we have a finite number of buckets, there is no way to eliminate collisions entirely. But we can try to minimize them as much as possible.
+
+#### Dealing with collisions
+
+Enters `Linked Lists`. If each `Node` inside the bucket is also a Linked List, Then we look for bucket `508` if it's empty we insert the head of Linked List, If a Node head exists in a bucket we follow that Linked List to add to the end of it.
```
01zulfi, line 135: Wording suggestion:
```suggestion
Up until now, our hash map is a one-dimensional data structure. What if each `Node` inside the bucket can store more than one value? Enters `Linked Lists`. Now, each `Node` inside the bucket is also a Linked List. When inserting to a bucket, if it's empty, we insert the head of Linked List. If a `Node` head exists in a bucket, we follow that Linked List to add to the end of it.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+You probably understand by this point why we must write a good hashing function which eliminates as many collisions as possible. Most likely you will not be writing your own hash functions, as most languages have it built in. But understanding how hash functions work is important.
+
+Check out [This Video](https://www.youtube.com/watch?v=btT4bCOvqjs) from CS50 that explains the concept shortly with visualization
```
01zulfi, line 139: This resource should go in the assignment section of the lesson.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Growth of a hash table
+
+Let's talk about the growth of our buckets. We don't have infinite memory, so we can't have an infinite amount of buckets. We need to start somewhere, but starting too big is also a waste of memory if we're only going to have a hash map with a single value in it. So to deal with this issue, we should start with a small array for our buckets. We'll use an array with 10 elements and indexes from 0 to 9.
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Most programming languages start with the default size of `16` because it's a power of 2, which help with some techniques for performance that require bit manipulation for indexes. But for this example, we will be using a starting size of 10.
+</div>
```
01zulfi, line 147: Nice addition. I wonder if we can also just use 16 as the starting size of our array?

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+How are we going to insert into those buckets when our hash function generates big numbers like `20353924`? We make use of the modulo `%` operation `given any number modulo by 10 we will get a number between 0 and 9`.
+For example, if we are to find the bucket where the value `"Manon"` will land, then we do the following:
+![hashing using hash code and modular operation example](./project_hash_map/imgs/01.png)
```
01zulfi, line 151: This paragraph doesn't contribute to the "Growth of hash tables" section. Maybe move it to "Buckets" section?

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+If we keep adding nodes into our buckets then the buckets will start filling up, but what is more important is we know for a fact that if almost all buckets have items in them, then it is guaranteed that some buckets will have collisions, It is Mathematically impossible not to.
```
01zulfi, line 153: Wording suggestion:
```suggestion
If we keep adding nodes into our buckets then the buckets will start filling up, but what is more important is we know for a fact that if almost all buckets have items in them, then collisions are guaranteed, it is mathematically impossible not to.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+Remember we don't want collisions, in a perfect world each bucket will either have 0 or 1 Node only, so we grow our buckets to have more chance that our Nodes will spread and not stack up in the same buckets. To grow our buckets, we create a new buckets list that is double the size of the old buckets list, then we retrieve all nodes from the old buckets and insert them into the new buckets.
```
01zulfi, line 155: Wording suggestion: 
```suggestion
Remember we don't want collisions, in a perfect world each bucket will either have 0 or 1 Node only, so we grow our buckets to have more chance that our Nodes will spread and not stack up in the same buckets. To grow our buckets, we create a new buckets list that is double the size of the old buckets list, then we copy all nodes over to the new buckets.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+#### When do we know that it's time to grow our buckets size
+
+To deal with this, our hash map class need to keep track of two new fields, the `capacity` and the `load factor`.
+
+- The `capacity` is the amount of buckets we currently have. Keeping track of this will let us know if our map has reached the threshold, the threshold will depend on our factor variable as well.
```
01zulfi, line 161: I'm not sure what this means: "the threshold will depend on our factor variable as well." We could eliminate this phrase entirely, thoughts?
```suggestion
- The `capacity` is the total number of buckets we currently have. Keeping track of this will let us know if our map has reached a certain threshold aka `load factor`,
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+- The `load factor` is a number that we can assign our hash map to at the start. It's the factor that will determine when is it a good time to grow our buckets, for example a load factor of `0.75` means our hash map will need to grow its buckets when the capacity reaches 75% full. Setting it too low will consume too much memory by having too many empty buckets, while setting it too high will allow our buckets to have collisions before we grow them. Usually a good balance of `0.75 to 1` is used.
```
01zulfi, line 163: Wording suggestion: 
```suggestion
- The `load factor` is a number that we can assign our hash map to at the start. It's the factor that will determine when is it a good time to grow our buckets, for example a load factor of `0.75` means our hash map will need to grow its buckets when the capacity reaches 75% full. Setting it too low will consume too much memory by having too many empty buckets, while setting it too high will allow our buckets to have many collisions before we grow them. Usually a good balance of `0.75 to 1` is used.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+How are we going to insert into those buckets when our hash function generates big numbers like `20353924`? We make use of the modulo `%` operation `given any number modulo by 10 we will get a number between 0 and 9`.
+For example, if we are to find the bucket where the value `"Manon"` will land, then we do the following:
+![hashing using hash code and modular operation example](./project_hash_map/imgs/01.png)
+
+If we keep adding nodes into our buckets then the buckets will start filling up, but what is more important is we know for a fact that if almost all buckets have items in them, then it is guaranteed that some buckets will have collisions, It is Mathematically impossible not to.
+
+Remember we don't want collisions, in a perfect world each bucket will either have 0 or 1 Node only, so we grow our buckets to have more chance that our Nodes will spread and not stack up in the same buckets. To grow our buckets, we create a new buckets list that is double the size of the old buckets list, then we retrieve all nodes from the old buckets and insert them into the new buckets.
+
+#### When do we know that it's time to grow our buckets size
```
01zulfi, line 157: Love this section and how in-depth the lesson goes on discussing the inner workings of hash map 💪

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+### Computation complexity
+
+A hash map is very efficient in its insertion, retrieving and removing operations. That is because we use array indexes to do those operations. A hash map has an average case complexity of `O(1)` for the following methods:
+- Insertion
+- Retrieving
+- Deletion
+
+Assuming we have a good hash map written. The worst case of those operations would be `O(n)` and that happens when we have all our data hashes to the same exact bucket.
+
+The Growth of our hash map has the complexity of `O(n)` at all times.
```
01zulfi, line 174: I feel this section doesn't talk about how collisions and use of linked lists affect complexities

01zulfi (changes requested):
Love the lesson 🔥 

Left A LOT of teeny tiny comments below. "Wording suggestion" comments are subjective for the most part and you are welcome to improve on them.

I'm plus one for splitting this lesson into two. The obvious ones are "What is a Hash Map" and "Project: Hash Map". @JoshDevHub @Mclilzee, what do you think?

javascript/computer_science/project_hash_map.md
Mclilzee: I don't follow. The complexity itself surfaces because of the linked list, and `O(n)` is because we are traversing the linked list to insert yet again another Node in the same bucket. Which happens specifically because of collisions.

Any suggestions?

01zulfi: > The complexity itself surfaces because of the linked list, and O(n) is because we are traversing the linked list to insert yet again another Node in the same bucket. Which happens specifically because of collisions.

^ this is exactly what I think should be included when talking about the worst case

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+You might be thinking, wouldn't it be just better to save the whole name as a hash code? That is true. This would make it unique for each name, but there are some good reasons to choose a number:
+
+- We do not save the hash code. The hash code is a calculation to find out which bucket (storage) our value will be placed in. Think about it as a locker number for example.
+
+- We save computation time. If we are to find where our hash is stored, it would be much faster and easier for a computer to compare two numbers rather than a string (or objects) to find the bucket where we have to store our elements (you will learn more about buckets shortly). Finding a bucket using a number would allow us to use that number as our index.
+
+- By using number as hash code, we can have hash codes that is uniform across variables where it doesn't have to only be a string, we can even hash numbers, or objects then we need an algorithm to generate a hash code number for our needs from an object, taking `class Person` object as an example, our hash code can be a combination of `strinngToNummber(name) + age`.
```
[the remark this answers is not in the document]
Mclilzee: At this point I wanted the ease in the switch I made after explaining hash codes using letters, and now we switched to a number, wanted to elaborate to not make that switch feel forced. But I may have written too many distracting words from the point of this lesson too, I'm not against this

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+}
+```
+
+With our new function we will have different hash codes for the names `"Sara"` and `"raSa"` that is because even if both names have the same letters, some of the letters appear in different locations. The hash code started to change because we are multiplying the old hash every new iteration and then adding the letter code.
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Notice the usage of prime number. We could have chosen any number we wanted, but prime numbers are even better to introduce less hash codes that are divisible by the same bucket length, which will make collisions less likely to happen.
+</div>
+
+Even though we reworked our hash function to avoid the `"Sara"/"raSa"` collision, there is always the possibility for collisions. Because we have a finite number of buckets, there is no way to eliminate collisions entirely. But we can try to minimize them as much as possible.
```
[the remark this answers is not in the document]
Mclilzee: I believe you missed a `we` after Since there, I will add this in my next PR

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+Let's start by learning what it means to hash a value. Hashing involves taking an input in and generating an output that represents the initial input. A hash function should be a pure function. Hashing the same input should always return the same hash code, and there should be no random generation component. For example, let's look at a hashing function that takes a name and gives us the first letter of that name:
+
+```javascript
+function hash(name) {
+  return name.charAt(0);
+}
+```
+
+We created our first basic hashing function. It takes a name and returns the first letter of that name.
+You might be familiar with this concept from your cipher exercise earlier in the foundations course, but there is a key difference between hashing and ciphering (encryption): reversibility.
```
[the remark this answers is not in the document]
Mclilzee: The concept of ciphering, where you take an input and produce an output that represent that input, like how people did it in the Caesar cipher exercise. Which I have just realized that it has been archived…

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+  bob.txt
+  beatrice.txt
+  bella.txt
+  benjamin.txt
+  bianca.txt
+```
+
+If we get new student in our school with the name `"Carlos"`, we can run our hash function to find out which folder to place them in. `hash("Carlos") -> "C"` so we put `"Carlos"` in the directory labeled `C`
+
+You might have spotted a problem: what if our school is populated with many people whose names share the same first letter `C`? Then we will have a directory labeled `C` that holds too many names that start with `"C"` while the other directories are empty. To eliminate this duplication and better separate our students, we need to rework our hash function while still maintaining a constant folder structure.
```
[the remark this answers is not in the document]
Mclilzee: Without changing the way we arrange our folders, we just update how we're going to label and insert items into them. But if it is not clear to you, then it probably not clear for others. I will remove that sentence

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+To get a value using a key, we put each entry inside a bucket as a `Node` item, which hold both the key and the value. To retrieve the value we hash a key, we find the bucket number, if the bucket is not empty then we go to that bucket we compare if the Node's key is the same key that is already in the bucket, if it is then we retrieve the Node's value otherwise we return null.
+
+Maybe you are wondering, why are we comparing the keys if we already found the hash code of that bucket? That is because remember, hash code is just the location, different items might generate the same hash code. We need to make sure the key is the same by comparing both keys that are inside the bucket.
+
+This is it, making this will result in a hash table with `search`, `set` and `get`.
```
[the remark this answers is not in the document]
Mclilzee: I believe I meant `has` not `search` my bad

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+To get a value using a key, we put each entry inside a bucket as a `Node` item, which hold both the key and the value. To retrieve the value we hash a key, we find the bucket number, if the bucket is not empty then we go to that bucket we compare if the Node's key is the same key that is already in the bucket, if it is then we retrieve the Node's value otherwise we return null.
+
+Maybe you are wondering, why are we comparing the keys if we already found the hash code of that bucket? That is because remember, hash code is just the location, different items might generate the same hash code. We need to make sure the key is the same by comparing both keys that are inside the bucket.
```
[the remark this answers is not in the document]
Mclilzee: You mean to replace the Bucket section entirely with a note? I believe it's too lengthy to be looked at as a note, unless we shorten it down. I don't think it hurts to keep it as a section, it fits nicely with all the bucket foreshadowing we have been doing in earlier sections.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+<div class="lesson-note lesson-note--warning" markdown="1">
+A hash map does not guarantee insertion order when you iterate over it, Because of the buckets indexing nature of hash coding. That means if you are to retrieve the array of keys and values to iterate over them, then they will not be in order of when you inserted them.
+
+Some libraries implement hash tables with insertion order in mind such as JavaScript's own `Map`. For this project however we will be implementing an unordered hash table.
+Example: if we insert the values `Mao`, `Zach`, `Xari` in this order, we may get back `["Zach", "Mao", "Xari"]` when we call an iterator.
+
+If iterating over the hash map frequently is your goal, then this data structure is not the right choice for the job, a simple array would be better.
+</div>
```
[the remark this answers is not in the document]
Mclilzee: I wanted to explain that since hash codes translation to indexes does not go in linear fashion from first to last index, and rather more chaotic of where the index going to be no matter the insert order.

If we are to add a heading, wouldn't it be better to remove the note entirely and just make it a subsection?

javascript/computer_science/project_hash_map.md
Mclilzee: The Buckets section does not introduce the initial size of our array, I had to add this to clear a confusion of how are we going to translate big numbers when our array is only limited to 16? If I need to move this to the buckets section, then I will have to find a way to introduce the array size there as well.

Any suggestions?

Mclilzee:
> I'm plus one for splitting this lesson into two. The obvious ones are "What is a Hash Map" and "Project: Hash Map". @JoshDevHub @Mclilzee, what do you think?

I'm with splitting this lesson into two. In that case we should replace JavaScript code with psudo code so that it can translate nicely to the Ruby section, then the project itself can be refined into two projects one for Ruby and another for JavaScript.

javascript/computer_science/project_hash_map.md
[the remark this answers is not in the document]
01zulfi: > Maybe you are wondering, why are we comparing the keys if we already found the hash code of that bucket? That is because remember, hash code is just the location, different items might generate the same hash code. We need to make sure the key is the same by comparing both keys that are inside the bucket.

^ I just mean this

javascript/computer_science/project_hash_map.md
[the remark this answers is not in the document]
01zulfi: > hash codes translation to indexes does not go in linear fashion from first to last index, and rather more chaotic of where the index going to be no matter the insert order.

this is more clear than "indexing nature of hash coding", can you add this to the lesson instead please?

> If we are to add a heading, wouldn't it be better to remove the note entirely and just make it a subsection?

Agreed.

01zulfi:
@Mclilzee  I'm of the opinion we keep JavaScript code. It allows learners do run the code for themselves and that can be a valuable learning experience. We can just have a separate Ruby lesson with its own code snippets.

javascript/computer_science/project_hash_map.md
01zulfi: Nice observation. Then, I propose the following wording suggestion:
```diff
- How are we going to insert into those buckets when our hash function generates big numbers like `20353924`? We make use of the modulo `%` operation `given any number modulo by 10 we will get a number between 0 and 9`.
+ Now that we have imposed the restriction of finite buckets, how will we deal with hash codes that exceed the available indexes in our hash map like `20353924`.  We make use of the modulo `%` operator. 
```

JoshDevHub:
Yeah the Ruby and JavaScript paths already maintain different lessons for all the Computer Science stuff. It's a bit annoying, but we kind of have to do it that way. Even outside of the code examples, you'd have to account for things like the JavaScript lesson referencing `Map` and object literal vs. the Ruby lesson referencing Ruby's hash structure. It's very difficult to make the lessons completely language agnostic.

JoshDevHub:
Also great review @01zulfi :raised_hands: 

I absolutely agree that we can have lesson + project here. And great work @Mclilzee :rocket:
This is a really cool addition to the CS section.

I'll go through everything one more time while building up a Ruby translation. I don't expect I'll notice anything major. Will probably just be English language nitpick stuff. But I'll try to get this done in the next 1-2 weeks. Feel free to @ me here if I take longer than that :sweat_smile:

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,230 @@
...
+
+  return hashCode;
+}
+```
+
+With our new function we will have different hash codes for the names `"Sara"` and `"raSa"` that is because even if both names have the same letters, some of the letters appear in different locations. The hash code started to change because we are multiplying the old hash every new iteration and then adding the letter code.
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Notice the usage of prime number. We could have chosen any number we wanted, but prime numbers are even better to introduce less hash codes that are divisible by the same bucket length, which will make collisions less likely to happen.
+</div>
```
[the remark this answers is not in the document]
Mclilzee: It's how I learnt it the first time, long ago, from some sources and video on YouTube.

You might want to explore this lengthy discussion https://stackoverflow.com/questions/299304/why-does-javas-hashcode-in-string-use-31-as-a-multiplier/299748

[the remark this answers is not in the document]
01zulfi: Cool read; this link should definitely be in the lesson

01zulfi:
@Mclilzee Looks like we're all on the same page regarding the split: a lesson and a project. Please make the split change in this PR as well

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,53 @@
+### Introduction
+
+It is time that you create your own hash map. You already know the magic of how it all works out from the previous lesson, now it is time for you to create your own magical implementation.
```
01zulfi, line 3: Wording suggestion: 
```suggestion
You already know the magic behind hash maps, now it's time to write your own implementation!
```

01zulfi (changes requested):

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,53 @@
...
+
+  1. When you create a function that takes a hash code value and returns the bucket index, use this snippet before the function returns to make sure it's accessing inbound array indexes only.
+
+      ```javascript
+      if (index < 0 || index >= buckets.length) {
+        throw new Error("Trying to access index out of bound");
+      }
+      ```
+
+  1. Inside the class DO NOT use Objects to store your key value pairs, and DO NOT use the array method `push()` to insert your nodes into the buckets. Always use indexes.
```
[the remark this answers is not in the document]
JoshDevHub: I think he means using object literals like:
```js
// not okay
{ key: keyToInsert, value: valueToInsert }

// :thumbsup:
new Node(keyToInsert, valueToInsert)  // class
Node(keyToInsert, valueToInsert)  // factory
```
I agree that we need to be precise with the language here. JavaScript mingles the idea of "hash map" and "object" in a way most other languages don't.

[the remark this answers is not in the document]
Mclilzee: It could be renamed to Object Literal `{}` and can be more specific on how not to use that object literal, Josh showcase of what is ok and what is not ok is brilliant we can use that

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,53 @@
...
+
+     - Remember to grow your buckets size when it needs to, by calculating if your bucket has reached the `load factor`.
+
+  1. `get` takes one argument as a key and returns the value that is assigned to this key. If key is not found, return `null`.
+
+  1. `has` takes a key as an argument and returns `true` or `false` based on whether or not the key is in the hash map.
+
+  1. `remove` takes a key as argument and removes it from the hash table.
+
+  1. `length` returns the number of stored keys in the hash map.
```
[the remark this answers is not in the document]
Mclilzee: Capacity is inner details that don't usually get exposed to the outside world, as you don't really need it outside a hash map. However, sometimes hash map creations gets an overloaded constructor where you can specify your starting capacity in case that you already know your hash map going to be big from the start. The same goes for the load factor.

I'm not sure if we should add the overloaded constructor details, or more methods to the project. There are many that we can add like `addIfAbsent` or `getOrDefault` and so on. But those are just extra quality of life kind of things, while our lesson of importance is to understand the inner workings of a hash map, not necessarily the most evolved one.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,53 @@
...
+
+  1. `clear` removes all entries in the hash map.
+
+  1. `keys` returns an array containing all the keys inside the hash map.
+
+  1. `values` returns an array containing all the values.
+
+  1. `entries` returns an array that contains each `key, value` pair. Example: `[[firstKey, firstValue], [secondKey, secondValue]]`
+
+  Remember that a hash map does not preserve insertion order when you are retrieving your arrays' data, it is normal and expected for keys and values to appear out of the order you inserted them in.
```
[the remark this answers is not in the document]
Mclilzee: I see where I got mixed up with words, I think I meant to say `when you are retrieving your hash map's data as an array` since this is for the `entries` method which gives us an array.

Let me know if you think just saying `hash map's` data sounds better, or we can word it differently to present the idea

javascript/computer_science/project_hash_map.md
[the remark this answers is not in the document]
01zulfi: It's clearly said in the lesson that key value pairs are stored in the Linked List's node data structure. Then, why do we have to state that objects cannot be used when the learner will always be using a node data structure.

javascript/computer_science/project_hash_map.md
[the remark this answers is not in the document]
01zulfi: Sounds reasonable. We could probably add the constructor (and other methods) in extra credit section sometime in the future. Good to go for now!

javascript/computer_science/project_hash_map.md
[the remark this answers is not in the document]
01zulfi: I think `hash map's` sound clearer. If you other ideas to present this information, go for it

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,199 @@
...
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Notice the usage of prime number. We could have chosen any number we wanted, but prime numbers are even better to introduce less hash codes that are divisible by the same bucket length, which will make collisions less likely to happen.
+</div>
+
+Even though we reworked our hash function to avoid the `"Sara"/"raSa"` collision, there is always the possibility for collisions. Since we have a finite number of buckets, there is no way to eliminate collisions entirely. Let's try to minimize them.
+
+#### Dealing with collisions
+
+Up until now, our hash map is a one-dimensional data structure. What if each `Node` inside the bucket can store more than one value? Enters `Linked Lists`. Now, each `Node` inside the bucket is also a Linked List. When inserting to a bucket, if it's empty, we insert the head of Linked List. If a `Node` head exists in a bucket, we follow that Linked List to add to the end of it.
```
01zulfi, line 137: > Now, each `Node` inside the bucket is also a Linked List. 

Now that I'm reading it again, this phrase might come off as confusing. We could just say, 

> Now, each bucket will be a Linked List. 

What are your thoughts on this?
```suggestion
Up until now, our hash map is a one-dimensional data structure. What if each `Node` inside the bucket can store more than one value? Enters `Linked Lists`. Now, each bucket will be a Linked List. When inserting to a bucket, if it's empty, we insert the head of Linked List. If a head exists in a bucket, we follow that Linked List to add to the end of it.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### When do we know that it's time to grow our buckets size
+
+To deal with this, our hash map class need to keep track of two new fields, the `capacity` and the `load factor`.
+
+- The `capacity` is the total number of buckets we currently have. Keeping track of this will let us know if our map has reached a certain threshold aka `load factor`,
+
+- The `load factor` is a number that we can assign our hash map to at the start. It's the factor that will determine when is it a good time to grow our buckets, for example a load factor of `0.75` means our hash map will need to grow its buckets when the capacity reaches 75% full. Setting it too low will consume too much memory by having too many empty buckets, while setting it too high will allow our buckets to have many collisions before we grow them. Usually a good balance of `0.75 to 1` is used.
```
01zulfi, line 163: IMO we can remove the phrase. "Usually a good ...". I don't think it serves any purpose, what do you think?
```suggestion
- The `load factor` is a number that we can assign our hash map to at the start. It's the factor that will determine when is it a good time to grow our buckets, for example a load factor of `0.75` means our hash map will need to grow its buckets when the capacity reaches 75% full. Setting it too low will consume too much memory by having too many empty buckets, while setting it too high will allow our buckets to have many collisions before we grow them.
```

01zulfi (changes requested):

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,199 @@
...
+  bob.txt
+  beatrice.txt
+  bella.txt
+  benjamin.txt
+  bianca.txt
+```
+
+If we get new student in our school with the name `"Carlos"`, we can run our hash function to find out which folder to place them in. `hash("Carlos") -> "C"` so we put `"Carlos"` in the directory labeled `C`.
+
+You might have spotted a problem: what if our school is populated with many people whose names share the same first letter `C`? Then we will have a directory labeled `C` that holds too many names while other directories could be empty. To eliminate this duplication and better separate our students, we need to rework our hash function while still maintaining a constant folder structure.
```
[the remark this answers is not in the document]
Mclilzee: We did, must have missed changing it with all the problems that are surfacing. Will get to it when i work on the rest

javascript/computer_science/hash_map_data_structure.md
Mclilzee: It can be reworded, but I feel letting people know that there is a conventional `sweet spot` of where a good balanced load factor can be. Rather than people just deciding a random number to be their load factor

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,199 @@
...
+
+In this lesson, you will learn how it all works, and even implement your own hash map in the next project! To start, here is a brief description of a hash map: a hash map takes in a key value pair, produces a hash code, and stores the pair in a bucket. Hash codes? Buckets? What? Don't fret, we'll learn all about these concepts and more. Buckle up and let's dive in!
+
+### Lesson overview
+
+By the end of this lesson, you will learn about:
+
+- Hash codes and how to generate them.
+- Hash maps and how they work internally.
+- Turning the concept of hashing into a data structure.
```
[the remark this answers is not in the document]
Mclilzee: That can be removed, It actually took work to come up with this sentence because I felt the list was a little short lol

javascript/computer_science/hash_map_data_structure.md
01zulfi: I'm curious about the 0.75-1 figure. I feel it's just easier to reword it as "Many hash map implementations use a default load factor of `0.75`"

Mclilzee: Why wouldn't we recommend the load factor to be between 0.75 and 1? I don't understand the reason we trying to avoid that. Saying many hash map implementation uses 0.75 is also not very accurate, as the number varies, but it's usually between 0.75 and 1.
Rust: 0.9
Java : 0.75
C#: 1

Do you have any reason to believe that this number is not optimal between space and performance complexities?

01zulfi: I just thought giving a range of load factors might overwhelm the learners. But thanks for digging in and providing load factors in different languages. Let's put it in. My wording suggestion: "Hash map implementations across various languages use a load factor between 0.75 and 1" You can tweak the wording of course.

01zulfi: I'm looking at load factors in different languages. Go has 6.5 if I'm reading it right https://github.com/golang/go/blob/fbfc203/src/runtime/hashmap.go#L67

Mclilzee: 6.5 felt way too big and gonna have on average 6 collisions per bucket on average so something felt wrong and I did some digging. I think they use a load factor differently than our implementation they go with 6.5 out of 8 buckets so around 81% you can see the math here with 13 (double of load factor) divided by 16 (double of 8 buckets)
 check out this line here https://github.com/golang/go/blob/9b4b3e5acca2dabe107fa2c3ed963097d78a4562/src/runtime/map.go#L69C1-L69C1
 
 If that is true then in our implementation Go would have a load factor of 0.8

01zulfi: Cool cool, I really can't understand code written in these repos lol Thank God for the comments 😆 

 This article (https://rcoh.me/posts/hash-map-analysis/) does say 6.5 for Go. Interestingly Ruby does load factor of 0.5 and uses an array instead of linked lists for buckets https://github.com/ruby/ruby/blob/master/st.c#L9

Mclilzee: I still find it really hard to believe a language would allow up to 6 pairs per bucket, that is too much performance hit to save memory, especially now days that memory is cheap. Unless their implementation is vastly different than a normal hash map.

javascript/computer_science/project_hash_map.md
JoshDevHub: I like this change, and think it's better than mine :laughing: @01zulfi 

Do you have thoughts on it @Mclilzee 

Another thing I think we may want to specify here is: what types of things we can hash and use as keys in the hash map. With ruby hashes and JavaScript `Map`, you can use just about anything as a key (numbers, arrays, sets, objects). Under the hood, the hashing algorithm will generate different codes for things like `1` vs. `"1".

The other way would be like a JavaScript object implementation, where everything is just coerced to strings.
```js
const obj = {}
obj[1] = "value"; // stored a number key right

console.log(obj["1"]); // wrong
// logs "value"
```
Since the lesson only talked about how to deal with hashing strings, I think taking the 2nd approach would be best? And I also think it's probably a good idea to say that people will not be required to deal with hashing different types.

JoshDevHub (changes requested):
A few more comments. I'll try to have some more soon. Slowly working my way through the Ruby translation.

javascript/computer_science/project_hash_map.md
Mclilzee: It is still the same concept, the only thing that changes is the hash code input, the output is still a number. at this point, students should have enough experience to work around this.

We can add it as extra bonus to update the hash function so that it accepts any object and be able to generate meaningful hash code with it, Thoughts?

01zulfi: IMO this could be a limitation as the project does seem quite lengthy as it stands. Something like:

> Hash maps could accommodate various data types for keys like numbers, strings, objects. But for this project, only handle keys of type strings.

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,53 @@
...
+
+  1. Use the following snippet whenever you access a bucket through an index. We want to throw an error if we try to access an out of bound index:
+
+      ```javascript
+      if (index < 0 || index >= buckets.length) {
+        throw new Error("Trying to access index out of bound");
+      }
+      ```
+
+  1. Inside the class DO NOT use the array method `push()` to insert your nodes into the buckets. Always use indexes.
```
01zulfi, line 17: Won't our learners be implementing linked lists for buckets? So I don't think we have to worry about them using the `push` method.

Mclilzee: Linked Lists are the items inside the buckets. The buckets themselves needs to be created using an Array and be accessed through indexes only.

Mclilzee: That leads me to believe your suggestion for this to be changed was because you believed Linked List were the buckets themselves? 
https://github.com/Mclilzee/curriculum/commit/490c30f43b7ca387c33b47eedc8ddd086c22bde8#diff-2ae1edcda41d250feb4bded479b31f662bd5c7231c17553e9e351b9ba9e02b67L137

I will revert it back so that it is more clear that Nodes as linked lists are _inside_ the buckets

01zulfi: I'm not married to any semantic here, just stating what is clear to _me_. If you feel saying "linked list is _inside_ a bucket" is clearer, go with that. I wonder if @JoshDevHub has a preference here.

Mclilzee: I assumed you thought about buckets being implemented in a Linked List and learners have no access to the `push` method. But since we are on the same page, I don't want learners to use `push` to insert the nodes inside the buckets.

I see your point of view more clearly now, where if you have a linked list inside the buckets, then the push method will not help since you need to traverse that list to find a location for a new node.

Mclilzee:
@JoshDevHub any update on the translation so far? The JavaScript version should be 99% ready by now

01zulfi (approved):
pr is a go from my side 🚀 

will wait on @JoshDevHub's ruby pr before merging

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+
+Buckets are storage that we need to store our elements. Simply, it's an array. For a specific key, we decide which bucket to use for storage through our hash function. Hash function returns a number which serves as the index of the array at which we store this specific key value pair. Let's say we need to store a key "Fred":
+
+1. Pass "Fred" into the hash function to get the hash code which is `508`
+1. Find the bucket at index `508`
+1. Store the key value pair in that bucket.
+
+This is an oversimplified explanation; we'll discuss more internal mechanics later in the lesson.
+
+To get a value using a key, we put each entry inside a bucket as a `Node` item, which hold both the key and the value. To retrieve the value we hash a key, we find the bucket number, if the bucket is not empty then we go to that bucket we compare if the Node's key is the same key that is already in the bucket, if it is then we retrieve the Node's value otherwise we return null.
```
JoshDevHub, line 95:
```suggestion
To get a value using a key, we put each entry inside a bucket as a `Node` item, which holds both the key and the value. To retrieve the value, we hash the key and calculate its bucket number. If the bucket is not empty, then we go to that bucket and compare if the Node's key is the same key that is already in the bucket. If it is, then we can return the Node's value. Otherwise, we return null.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+
+This is it, making this will result in a hash table with `has`, `set` and `get`.
+
+What if we found the hash code, but also the key value is the same as what we already have in the bucket. We check if it's the same item by comparing the keys, then we overwrite the value with our new value. This is how we can only have unique values inside a `Set`, `Set` is similar to a hash map but the key difference (pun intended) is that a `Set` will have Nodes with only keys and no values.
+
+#### Insertion order is not maintained
+
+A hash map does not guarantee insertion order when you iterate over it. Hash codes translation to indexes does not go in linear fashion from first to last index, and rather more chaotic of where the index going to be no matter the insert order. That means if you are to retrieve the array of keys and values to iterate over them, then they will not be in order of when you inserted them.
```
JoshDevHub, line 105:
```suggestion
A hash map does not guarantee insertion order when you iterate over it. The translation of hash codes to indexes does not follow a linear progression from the first to the last index. Instead, it is more unpredictable, irrespective of the order in which items are inserted. That means if you are to retrieve the array of keys and values to iterate over them, then they will not be in order of when you inserted them.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+Example: if we insert the values `Mao`, `Zach`, `Xari` in this order, we may get back `["Zach", "Mao", "Xari"]` when we call an iterator.
+
+If iterating over the hash map frequently is your goal, then this data structure is not the right choice for the job, a simple array would be better.
+
+### Collisions
+
+We have another problem that we need to deal with: collisions. A collision occurs when two different keys generate the exact same hash code. Because they have the same hash code, they will land in the same bucket.
+
+Let's take an example: hashing the name `"Sara"` and the name `"raSa"` will generate the same hash code. That is because the letters in both names are the same, just arranged differently.
+We can rework our `stringToNumber` function so that it can give us unique hash codes which depends on where the letter appear in the name using an algorithm.
```
JoshDevHub, line 117:
```suggestion
We can rework our `stringToNumber` function so that it can give us unique hash codes that depend on where the letters appear in the name using an algorithm.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+  const primeNumber = 31;
+  for (let i = 0; i < string.length; i++) {
+    hashCode = primeNumber * hashCode + string.charCodeAt(i);
+  }
+
+  return hashCode;
+}
+```
+
+With our new function we will have different hash codes for the names `"Sara"` and `"raSa"` that is because even if both names have the same letters, some of the letters appear in different locations. The hash code started to change because we are multiplying the old hash every new iteration and then adding the letter code.
```
JoshDevHub, line 132:
```suggestion
With our new function we will have different hash codes for the names `"Sara"` and `"raSa"`. This is because even if both names have the same letters, some of the letters appear in different locations. The hash code started to change because we are multiplying the old hash every new iteration and then adding the letter code.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Notice the usage of prime number. We could have chosen any number we wanted, but prime numbers are even better to introduce less hash codes that are divisible by the same bucket length, which will make collisions less likely to happen.
```
JoshDevHub, line 135:
```suggestion
  Notice the usage of a prime number. We could have chosen any number we wanted, but prime numbers are preferable. Multiplying by a prime number will reduce the likelihood of hash codes being evenly divisible by the bucket length, which helps minimize the occurrence of collisions.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+
+You probably understand by this point why we must write a good hashing function which eliminates as many collisions as possible. Most likely you will not be writing your own hash functions, as most languages have it built in. But understanding how hash functions work is important.
+
+### Growth of a hash table
+
+Let's talk about the growth of our buckets. We don't have infinite memory, so we can't have an infinite amount of buckets. We need to start somewhere, but starting too big is also a waste of memory if we're only going to have a hash map with a single value in it. So to deal with this issue, we should start with a small array for our buckets. We'll use an array size `16`.
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+  Most programming languages start with the default size of `16` because it's a power of 2, which help with some techniques for performance that require bit manipulation for indexes.
```
JoshDevHub, line 151:
```suggestion
  Most programming languages start with the default size of `16` because it's a power of 2, which helps with some techniques for performance that require bit manipulation for indexes.
```

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -0,0 +1,204 @@
...
+
+#### When do we know that it's time to grow our buckets size
+
+To deal with this, our hash map class need to keep track of two new fields, the `capacity` and the `load factor`.
+
+- The `capacity` is the total number of buckets we currently have. Keeping track of this will let us know if our map has reached a certain threshold aka `load factor`,
+
+- The `load factor` is a number that we can assign our hash map to at the start. It's the factor that will determine when is it a good time to grow our buckets, for example a load factor of `0.75` means our hash map will need to grow its buckets when the capacity reaches 75% full. Setting it too low will consume too much memory by having too many empty buckets, while setting it too high will allow our buckets to have many collisions before we grow them. Hash map implementations across various languages use a load factor between `0.75` and `1`.
```
JoshDevHub, line 168:
```suggestion
- The `load factor` is a number that we can assign our hash map to at the start. It's the factor that will determine when is it a good time to grow our buckets. For example, a load factor of `0.75` means our hash map will need to grow its buckets when the capacity reaches 75% full. Setting it too low will consume too much memory by having too many empty buckets, while setting it too high will allow our buckets to have many collisions before we grow them. Hash map implementations across various languages use a load factor between `0.75` and `1`.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,55 @@
+### Introduction
+
+You already know the magic behind hash maps, now it's time to write your own implementation!
+
+#### Limitation
+
+  Before we get started, we need to lay down some ground rules. JavaScript's dynamic nature of array allow us to insert and retrieve indexes that is outside our array size range. Example: if we create an array of size `16` to be our buckets size, nothing stopping us from storing items at index `500`. This defeats the purpose we are trying to demonstrate, so we need to put some self restriction to work around this.
```
JoshDevHub, line 7:
```suggestion
  Before we get started, we need to lay down some ground rules. JavaScript's dynamic nature of array allow us to insert and retrieve indexes that are outside our array size range. Example: if we create an array of size `16` to be our buckets size, nothing stopping us from storing items at index `500`. This defeats the purpose we are trying to demonstrate, so we need to put some self restriction to work around this.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,55 @@
...
+
+  1. Start by creating a `HashMap` class or factory function. It's up to you which you want to use. Then proceed to create the following methods:
+
+  1. `hash` takes a value and produces a hash code with it. We did implement a fairly good `hash` function in the previous lesson. You are free to use that, or if you wish, you can conduct your own research. Beware, this is a deep deep rabbit hole.
+
+      <div class="lesson-note lesson-note--tip" markdown="1">
+        Hash maps could accommodate various data types for keys like numbers, strings, objects. But for this project, only handle keys of type strings.
+      </div>
+
+  1. `set` takes two arguments, the first is a key and the second is a value that is assigned to this key, if a key already exists then the old value is overwritten.
```
JoshDevHub, line 29:
```suggestion
  1. `set` takes two arguments, the first is a key and the second is a value that is assigned to this key. If a key already exists, then the old value is overwritten.
```

javascript/computer_science/project_hash_map.md
```diff
@@ -0,0 +1,55 @@
...
+
+  1. `clear` removes all entries in the hash map.
+
+  1. `keys` returns an array containing all the keys inside the hash map.
+
+  1. `values` returns an array containing all the values.
+
+  1. `entries` returns an array that contains each `key, value` pair. Example: `[[firstKey, firstValue], [secondKey, secondValue]]`
+
+  Remember that a hash map does not preserve insertion order when you are retrieving your hash map's data, it is normal and expected for keys and values to appear out of the order you inserted them in.
```
JoshDevHub, line 49:
```suggestion
  Remember that a hash map does not preserve insertion order when you are retrieving your hash map's data. It is normal and expected for keys and values to appear out of the order you inserted them in.
```

JoshDevHub (changes requested):
Ran into just a few more small grammar/punctuation nitpicks while working on the Ruby translation. I promise these are the last ones I have :smile: 

But yeah after this, I'll approve. Have also got my own PR for the ruby content nearly ready to go. So we should be good to merge very soon :partying_face: 

One question @Mclilzee : do you want to do the honors with adding these lessons to the main site? You can see how it's done [here](https://github.com/TheOdinProject/theodinproject/wiki/Adding-a-Lesson). If you want me to take care of it, I could. But figured I'd offer you the chance to. Often when people write their first lesson, they like to be the person who adds it in.

Excellent work @Mclilzee :muscle: 
And thanks for the awesome reviews and ideas @01zulfi :1st_place_medal:

Mclilzee:
@JoshDevHub Thank you for holding my hand through all this grammar :rofl: 

It doesn't matter to me who adds the lessons to the site, since you are working on the Ruby version and will do the merges then I would say you go for it so that you can merge when you are done 😄

01zulfi:
@Mclilzee One final task.

Can you replace the relative links for images with statically links in a new PR. [how I created these links if you're interested](https://github.com/TheOdinProject/curriculum/wiki/Adding-Images-to-the-Curriculum)
- https://cdn.statically.io/gh/TheOdinProject/curriculum/7ea463cfb7c05c330d72f5977cc5fe3b0c640b86/javascript/computer_science/hash_map_data_structure/imgs/00.png
- https://cdn.statically.io/gh/TheOdinProject/curriculum/7ea463cfb7c05c330d72f5977cc5fe3b0c640b86/javascript/computer_science/hash_map_data_structure/imgs/01.png

01zulfi:
@Mclilzee I do agree with Josh. You should be the one to add the lessons to the site 😄 (but your call at the end of the day)

Mclilzee:
@01zulfi I don't mind adding it if you and Josh believe that I should do so, then who am I to argue with the majority 😅 

Where would you like the lesson to be, however? Since it makes use of Linked Lists it only makes sense to come after the Linked List lesson, but should it be before or after Binary Search Tree?

I think it should be before the Knights Travail project since that is a nice project to end the computer science section with.

01zulfi:
I might have a slight preference to add it right after linked lists. We could also argue to have it after BST to separate out the difficult BST and Knight Travails projects with a relatively easy one i.e. hashmap.

Either way, I'm fine with both placements. Or, [we could leave it to fate lol](https://wheeldecide.com/index.php?c1=hashmap+before+bst&c2=hashmap+after+bst&col=gbblackp&time=5)

---

Please add both Ruby & JS versions in one PR.

PR #26545, opened 2023-10-22
Update transforms.md

sapienfrom2000s:
Makes false claim about chain transformation

## Because
It makes a false claim about chain transformations.

## This PR
Removes a stanza from the lesson.

## Issue

Closes #XXXXX

## Additional Information
The following sentence is taken from codepen article which makes the false claim._The end result relatively to the original element is exactly the same for both of these interpretations and this is true for any list of transforms, even with more than 2 transforms_.

## Pull Request Requirements


fabulousgk:
So your solution is to completely remove the article?  THis make no sense to me.

fabulousgk:
I also dont see that line as being false.

sapienfrom2000s:
Order of transformation matters in the lesson wheras the article contradicts it by saying it doesn't matter.

fabulousgk:
The lesson also says there is some disagreement on that and provides the article to clarify

PR #26555, opened 2023-10-24
Update project_flickr_api.md

sapienfrom2000s:
## Because
Using gem and env variable is a bit outdated. Rails credentials is the new way to go for storing credentials.

## This PR
Introduces rails credentials to store keys.

## Issue

Closes #XXXXX

## Pull Request Requirements


CouchofTomato (changes requested):
While I'm in favour of this change generally I'm not 100% sure by the article. It is a nice introduction but I wonder if some of the ideas in there are a little dated. https://guides.rubyonrails.org/security.html#environmental-security uses examples of accessing credentials all as method calls in a chain, whereas the article makes use of the `dig` method.

We could perhaps link to both and maybe highlight they show different ways to access the credential values.

What are your thoughts?

sapienfrom2000s:
_In an object-oriented fashion the call to Rails.application.credentials.stripe returns a hash of the key-value pairs we just added._
This is mentioned just before he said that he likes to use _dig_ to access nested values. Although I would like to link docs, it will lead to addition of one more document. I am not sure if it would not make the paragraph more crowded. 
Also, now, students will have 3 resources to add credentials. Do you think we can drop one of them to keep the lesson slim?

CouchofTomato:
@sapienfrom2000s 

Because the links are shortened in the lesson I'm not sure how much overhead it will add.

Perhaps a middle ground is to add a note as you suggested, but instead of giving an example we just link to the docs. Something like...

Although not mentioned in the article, the value of each key can also be accessed by chaining keys as methods as shown in the [Rails Guides](https://guides.rubyonrails.org/security.html#environmental-security) examples

Do you think this is a good middle ground?

sapienfrom2000s (commented):
Yes it's a good middle ground. Are you cool with the current commit, or do you want me to make some changes.

CouchofTomato (approved):

CouchofTomato:
Thanks @sapienfrom2000s 

Looks good to me 🥳

PR #26642, opened 2023-10-31
HTML Boilerplate: Fix formatting of code

schmidhauser:
* fix formatting of links and code fragments according to the general remark below
* replace hyphen (-) by en dash (--) if used between words with spaces
* add missing period at end of sentence

### General remark
With respect to the typography, I'm afraid to say that the entire page is somewhat carelessly written. One finds (say) "the `<body>` element" next to "the body element". I should adopt the former convention, which one may find (e.g.) at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body. One should probably also format the subtitles accordingly – e.g. "The `<html>` element". I'm happy to submit another PR if such a change is desired.

## Pull Request Requirements


foundations/html_css/html-foundations/html-boilerplate.md
```diff
@@ -60,25 +60,25 @@ Noticed the word `lang` here? It represents an HTML attribute which is associate
 The `<head>` element is where we put important meta-information **about** our webpages, and stuff required for our webpages to render correctly in the browser.
 Inside the `<head>`, we **should not** use any element that displays content on the webpage.
 
-#### The charset meta element
+#### The meta charset element
```
rlmoser99, line 63: I think part of the confusing wording in this lesson is that it is the `meta` element and the `charset` attribute. But these two get interchanged. I'd suggest updating the sub-heading to the following
```suggestion
#### The meta element
```

foundations/html_css/html-foundations/html-boilerplate.md
```diff
@@ -60,25 +60,25 @@ Noticed the word `lang` here? It represents an HTML attribute which is associate
...
+If we didn't include a `<title>` element, the webpage's title would default to its file name. In our case that would be `index.html`, which isn't very meaningful for users; this would make it very difficult to find our webpage if the user has many browser tabs open.
 
 There are many more elements that can go within the head of an HTML document. However, for now it's only crucial to know about the two elements we have covered here. We will introduce more elements that go into the head throughout the rest of the curriculum.
 
-Back in our `index.html` file, let's add a head element with a charset meta element and a title within it. The head element goes within the HTML element and should always be the first element under the opening `<html>` tag:
+Back in our `index.html` file, let's add a `<head>` element with a `<meta charset>` element and a title within it. The `<head>` element goes within the `<html>` element and should always be the first element under the opening `<html>` tag:
```
rlmoser99, line 81: To keep in line with `meta` element and `charset` attribute.

```suggestion
Back in our `index.html` file, let's add a `<head>` element with a `<meta>` element and a title within it. The `<head>` element goes within the `<html>` element and should always be the first element under the opening `<html>` tag:
```

foundations/html_css/html-foundations/html-boilerplate.md
```diff
@@ -177,7 +177,7 @@ This section contains questions for you to check your understanding of this less
...
+*   Another option for opening your HTML pages in the browser is using the [live server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) with VSCode. This will open your HTML document and automatically refresh it every time you save the document. However, we recommend not using this extension and instead doing it the old-fashioned way, by opening the page and refreshing the page manually in the browser for your first few HTML projects. This way, you can get used to that process and won't become reliant on extensions right away. One reason is that there may be subtle differences when using extensions. For example, live server will always use UTF-8 character encoding and not the value defined in your `<meta charset>` element. This could potentially hide some characters on your site as they will not be encoded the way you expect.
```
rlmoser99, line 180: In addition to just using `<meta>` element in this line as well.

rlmoser99 (changes requested):
Thank you for taking the time to update this lesson. I've made a few suggestions with wording that I found confusing, but I am happy to discuss them if you disagree.

In addition, I kindly suggest that next time you create a PR you do not insult the people that wrote the lesson, because sometimes small details get lost in the review process.
> the entire page is somewhat carelessly written

Finally, if I am understanding you correctly, you are asking if we should use backticks in the subtitles. 
> One should probably also format the subtitles accordingly – e.g. "The <html> element"

The answer is no. In our style guide that is listed in our PR template, it indicates that no code snippets should be in the headings.
https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#no-code-snippets

foundations/html_css/html-foundations/html-boilerplate.md
schmidhauser: Yes, I agree. "The meta element" seems to be the best solution.

For consistency, one should then probably also rephrase the other titles -- e.g. "The head element", "The title element".

foundations/html_css/html-foundations/html-boilerplate.md
schmidhauser: Following your earlier explanation, I think one should write here

> For example, live server will always use UTF-8 character encoding and not the value defined in your `<meta>` element.

schmidhauser:
Thank you for your comments. I fully agree with them. And apologies for my word choice. It wasn't at all intended in the way you perceived it.

One thing I didn't find in the Style Guide is a rule regarding en dashes (--) and em dashes (---). Do you know if either is preferred for TOP? Maybe one should add a note to the Style Guide.

foundations/html_css/html-foundations/html-boilerplate.md
```diff
@@ -60,25 +60,25 @@ Noticed the word `lang` here? It represents an HTML attribute which is associate
 The `<head>` element is where we put important meta-information **about** our webpages, and stuff required for our webpages to render correctly in the browser.
 Inside the `<head>`, we **should not** use any element that displays content on the webpage.
 
-#### The charset meta element
+#### The meta element
 
-We should always have the meta tag for the charset encoding of the webpage in the head element: `<meta charset="utf-8">`.
+We should always have the `<meta>` tag for the charset encoding of the webpage in the `<head>` element: `<meta charset="utf-8">`.
```
rlmoser99, line 65: What do you think of using "with" instead of "for" here?
```suggestion
We should always have the `<meta>` tag with the charset encoding of the webpage in the `<head>` element: `<meta charset="utf-8">`.
```

foundations/html_css/html-foundations/html-boilerplate.md
rlmoser99: What do you think about adding the `charset` here?
> For example, live server will always use UTF-8 character encoding and not the `charset` value defined in your `<meta>` element.

rlmoser99 (changes requested):
> One thing I didn't find in the Style Guide is a rule regarding en dashes (--) and em dashes (---). Do you know if either is preferred for TOP? Maybe one should add a note to the Style Guide.

> Another point is that TOP sometimes uses "VSCode" (as in the HTML Boilerplate page) and sometimes "VS Code" (as e.g. in https://www.theodinproject.com/lessons/foundations-git-basics). The latter is the official spelling. I think it would be good to use a single form.

I think these two questions would be good to raise, either in an issue on the repo or you could make a PR to the style guide. We probably would refer to other standards that are out there regarding proper uses for dashes.

foundations/html_css/html-foundations/html-boilerplate.md
schmidhauser: Yes, it's more precise.

foundations/html_css/html-foundations/html-boilerplate.md
schmidhauser: Yes, excellent.

rlmoser99: Great! I think this change is the only thing still needed on this PR before I can approve it.

foundations/html_css/html-foundations/html-boilerplate.md
rlmoser99: Oh, I missed this comment. In looking at the lesson this is the only one that uses "The" in the title, so I wonder if we should just change this one to be "Meta element"?

foundations/html_css/html-foundations/html-boilerplate.md
schmidhauser: Done. (At least I hope so.)

foundations/html_css/html-foundations/html-boilerplate.md
schmidhauser: Yes, I think that would be best.

rlmoser99 (approved):
Thanks for taking the time to discuss these changes and update them.

As a heads up, generally the person who requested changes is the one who should mark a comment resolved. This is to help ensure that the changes made meet the changes requested.

PR #26911, opened 2023-12-10
Fix grammar and add new path for knight in file project_knights_trava…

Zirconium-Zr:
…ils.md

Added new alternate path from [0, 0] square to [7, 7] and fixed few grammatical errors.

## Because

It adds new example which was not already present in knights travails path so if any programmers get the same path they will know that they are on a right track.

## This PR

- I changed the grammar in line 3 by adding "Now that you're pro" which I think sounds good to previously "Now you're a pro"

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_knights_travails.md
+++ b/javascript/computer_science/project_knights_travails.md
@@ -6,7 +6,7 @@ For this project, you'll need to use a data structure that's similar (but not id
 
 A knight in chess can move to any square on the standard 8x8 chess board from any other square on the board, given enough turns (don't believe it? See [this animation](https://cdn.statically.io/gh/TheOdinProject/curriculum/284f0cdc998be7e4751e29e8458323ad5d320303/ruby_programming/computer_science/project_knights_travails/imgs/00.png)). Its basic move is two steps forward and one step to the side or one step forward and two steps to the side. It can face any direction.
 
-All the possible places you can end up after one move looks like this:
+All the possible places you can end up after one move look like this:
 
 ![Knights Travails board](https://cdn.statically.io/gh/TheOdinProject/curriculum/d30038e0aaca1f35e58e205e37a21b2c9d31053d/javascript/computer_science/project_knights_travails/imgs/01.png)
 Note: The picture is only to explain the problem, There is no need to create a GUI.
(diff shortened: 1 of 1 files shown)


javascript/computer_science/project_knights_travails.md
```diff
@@ -1,41 +1,43 @@
...
 
-A knight in chess can move to any square on the standard 8x8 chess board from any other square on the board, given enough turns (don't believe it?  See [this animation](https://cdn.statically.io/gh/TheOdinProject/curriculum/284f0cdc998be7e4751e29e8458323ad5d320303/ruby_programming/computer_science/project_knights_travails/imgs/00.png)).  Its basic move is two steps forward and one step to the side.  It can face any direction.
+A knight in chess can move to any square on the standard 8x8 chess board from any other square on the board, given enough turns (don't believe it? See [this animation](https://cdn.statically.io/gh/TheOdinProject/curriculum/284f0cdc998be7e4751e29e8458323ad5d320303/ruby_programming/computer_science/project_knights_travails/imgs/00.png)). Its basic move is two steps forward and one step to the side or one step forward and two steps to the side. It can face any direction.
 
-All the possible places you can end up after one move look like this:
+All the possible places you can end up after one move looks like this:
```
JoshDevHub, line 9:
```suggestion
All the possible places you can end up after one move look like this:
```
The old way was correct. The verb "look" has the subject "places". Because the subject is plural, the verb has no "s"

I realize this is confusing because the singular "move" could seem to potentially be the subject, but it's not. It's part of the "after" prepositional phrase.

javascript/computer_science/project_knights_travails.md
```diff
@@ -1,41 +1,43 @@
...
 Your task is to build a function `knightMoves` that shows the shortest possible way to get from one square to another by outputting all squares the knight will stop on along the way.
 
-You can think of the board as having 2-dimensional coordinates.  Your function would therefore look like:
+You can think of the board as having 2-dimensional coordinates. Your function would therefore look like:
 
-  * `knightMoves([0,0],[1,2]) == [[0,0],[1,2]]`
+- `knightMoves([0,0],[1,2]) == [[0,0],[1,2]]`
 
-  Note: Sometimes _there is more than one fastest path_, an example of this is shown below. Either answer will work.
+Note: Sometimes _there are more than one fastest path_, an example of this is shown below. Either answer will work.
```
JoshDevHub, line 22:
```suggestion
Note: Sometimes _there is more than one fastest path_, an example of this is shown below. Either answer will work.
```
Again the old way is correct. The subject is the singular noun phrase "one fastest path", so the verb should be the singular "is".

JoshDevHub (changes requested):
A few changes. I agree on the section about multiple paths being a bit unclear, but I don't think listing every combination is the way to go about clarifying. Let me know what you think.

javascript/computer_science/project_knights_travails.md
```diff
@@ -1,41 +1,43 @@
...
 
-  * `knightMoves([0,0],[3,3]) == [[0,0],[2,1],[3,3]]` or `knightMoves([0,0],[3,3]) == [[0,0],[1,2],[3,3]]`
-  * `knightMoves([3,3],[0,0]) == [[3,3],[2,1],[0,0]]` or `knightMoves([3,3],[0,0]) == [[3,3],[1,2],[0,0]]`
-  * `knightMoves([0,0],[7,7]) == [[0,0],[2,1],[4,2],[6,3],[4,4],[6,5],[7,7]]` or `knightMoves([0,0],[7,7]) == [[0,0],[2,1],[4,2],[6,3],[7,5],[5,6],[7,7]]`
+- `knightMoves([0,0],[3,3]) == [[0,0],[2,1],[3,3]]` or `knightMoves([0,0],[3,3]) == [[0,0],[1,2],[3,3]]`
+- `knightMoves([3,3],[0,0]) == [[3,3],[2,1],[0,0]]` or `knightMoves([3,3],[0,0]) == [[3,3],[1,2],[0,0]]`
+- `knightMoves([0,0],[7,7]) == [[0,0],[2,1],[4,2],[6,3],[4,4],[6,5],[7,7]]` or `knightMoves([0,0],[7,7]) == [[0,0],[2,1],[4,2],[6,3],[7,5],[5,6],[7,7]]` or `knightMoves([0,0],[7,7]) == [[0,0],[2,1],[4,0],[6,1],[7,3],[6,5],[7,7]]`
```
JoshDevHub, line 26: Is the intention here to list all paths? Because there are more than these 3 for getting between [0, 0] and [7, 7]

I think rather than try to name all paths, we could instead change the language for the "note" section just above that says
> Either answer will work
This suggests that only the two answers named in the examples are valid, which isn't the case.

Zirconium-Zr: Hi, I updated the language for the note section and removed the path that I added. I think you are correct about listing every path is not the way to go about clarifying because there could be more than three paths to another coordinate. So I think saying all answers will work as long as they follow the path and is also the shortest path might be more clarifying.

JoshDevHub (approved):
Thank you for this contribution. I definitely think it makes the situation more clear :smiley:

PR #26930, opened 2023-12-14
Fix: Correct folder name from 'recipe page' to 'recipes directory'

KURELLANAVEEN:
The project incorrectly referred to a folder as a 'recipe page' when it should be named 'recipes directory'. This commit updates the relevant documentation or code to use the correct term, improving clarity and accuracy in the project structure.

## Because

This pull request aims to prevent confusion for users or readers when adding files to the recipes folder. The correction ensures a clear and straightforward structure, enhancing the user experience during file organization. The correction ensures accuracy and clarity in the project's structure, aligning the naming with the actual content and improving overall readability.

## This PR

1) Corrected the folder name from 'recipe page' to 'recipes directory'.

## Issue

Closes #2013

## Pull Request Requirements

-   [ x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [ x] The `Because` section summarizes the reason for this PR
-   [ x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x ] If this PR addresses an open issue, it is linked in the `Issue` section
-   [ x] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [x ] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

### DIFF
--- a/foundations/html_css/html-foundations/project-recipes.md
+++ b/foundations/html_css/html-foundations/project-recipes.md
@@ -69,7 +69,7 @@ Your new recipe page should have the following content:
 
 #### Iteration 4: add more recipes
 
-1. Add two more recipes with identical page structures to the recipe page you've already created.
+1. Add two more recipes with identical page structures to the `recipes` directory you've already created.
 2. Don't forget to link to the new recipes on the index page. Also, consider putting all the links in an unordered list so they aren't all on one line. 
 
 Example:


KevinMulhern:
This project has been live some years and this "problem" isn't something any of our learners have reported having trouble with.

KURELLANAVEEN:
In iteration 2, we have created a directory and named it 'recipes'. So, all the recipes(html files) that we are going to create comes under this directory. Then, we have created our first recipe page (example: chicken_curry.html) within the recipes directory.

Now, in iteration 3, we have given the instructions to follow for the recipe page content.


Now, when we jump into iteration 4, we are adding more recipes(in our case we are adding 2 more recipes). But, where should we add? Yeah, we should add these two recipes(ex: fish_curry.html, biryani.html) into the recipes directory.
More technically, 'recipes' is the parent directory, which holds three children(three recipe html files)

This is what I'm trying to correct. Currently, it is written as: "Add two more recipes with identical page structures to the recipe page". Here, recipe page is something we have seen in iteration 2&3(in our case it is chicken_curry.html). So, my concern is for beginners point of view, beginner programmers(like me) get stuck or confused at this point. It will be more clear and meaningful if we re-write "recipe page" with "recipes directory".



> This project has been live some years and this "problem" isn't something any of our learners have reported having trouble with.
> 
> I think theres maybe some confusion with the intention, 'recipe page' doesn't refer to a folder but the recipe html page created in iteration 3. Why do you think it refers to a directory @KURELLANAVEEN?



> This project has been live some years and this "problem" isn't something any of our learners have reported having trouble with.
> 
> I think theres maybe some confusion with the intention, 'recipe page' doesn't refer to a folder but the recipe html page created in iteration 3. Why do you think it refers to a directory @KURELLANAVEEN?

KevinMulhern:
We make reference to "recipe page" a few times in the previous iterations. The context of what it means should be fairly well established by the time they get to adding their own pages.
Screenshot 2023-12-14 at 14 47 37


I'm sorry @KURELLANAVEEN, But I don't think this is an issue.

PR #26944, opened 2023-12-17
Testing basics & More Testing lesson: Clarify instruction

mekbibDev:
## Because

The purpose of this pull request is to prevent learners from installing unnecessary packages and configuring unnecessary configurations. The original instruction tells users to simply follow the getting started guide on the Jest website but the instruction on that website has info that is irrelevant to learners at this stage. One of them is [Generate a basic configuration file](https://jestjs.io/docs/getting-started#generate-a-basic-configuration-file) which helps users configure their jest setup. However, the setup is unnecessary and may lead to confusion for learners. For example, it asks questions  whether they want to use typescript or not, and whether or not it should clear up mock calls and context. For the latter example take in mind that learners have never written a single mock test so far and might actually take a way their chance to use beforeEach/AfterEach jest code. The other configuration are also unnecessary at this stage. I have also clarified that they should install and configure babel so they can use ES6 modules since learners have so far never used commonJS modules which further on might lead them to mistakenly mix both when doing projects.

The other confusion that learners might come across is the using webpack with jest section in the getting started guide. Learners might be misled into thinking they need to follow this instruction to use jest when testing their upcoming project. However the instruction in that section, talks about how you would set up jest with webpack for CSS and static assets, which is again irrelevant for learners at this stage because they are only going to be doing unit testing and have not been taught about integration testing. This is also the reason that I have removed using jest with webpack in the more testing section.

## This PR

- Adds a sentence to clarify instruction
- Removes a resource form the additional resource section

## Issue

Closes #XXXXX

## Pull Request Requirements


mekbibDev:
I know this is yet to be reviewed but I want to let you know that I made a mistake on one of the points. I forgotten that instructions for using babel with jest is mentioned in the testing practice. so the part about using jest with babel is redundant.

rlmoser99:
> I know this is yet to be reviewed but I want to let you know that I made a mistake on one of the points. I had forgotten that instructions for using babel with jest are mentioned in the testing practice. so the part about using jest with babel that I've added is redundant.

@mekbibDev If you know part of the changes are no longer needed, you should change them back.

mekbibDev:
alright. I was worried that it might interfere with the review flow or that maybe I need to do a whole new pull request.

mekbibDev:
@rlmoser99 @codyloyd  I have made the changes

codyloyd (approved):

PR #27020, opened 2023-12-30
JavaScript Fundamentals Part 1: Added additional reference source for varible types

macyapp:
## Because

I have added a W3Schools link to the JavaScript tutorial which explains the use of 'const', a detail not previously included. This article clarifies that 'const' is used to declare arrays and objects. While the type of the variable won't change, the data within it can be modified if it's an array or an object.

## This PR

- Added links of W3Schools let and const article.

## Issue

Closes #XXXXX

## Pull Request Requirements

-   [ x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
-   [x ] The `Because` section summarizes the reason for this PR
-   [x ] The `This PR` section has a bullet point list describing the changes in this PR


linkonsat:
So looking at the resource that is already there and the W3 article. It doesn't appear to have anything beyond that one line. Given that I'm not sure it makes sense to add another resource for specifically that. The resource that is already there does say when a programmer would use it. As such I don't think the W3 resource is quite as strong as I would want it to be for it to be added. I do however greatly appreciate it!

Issue #26976, opened 2023-12-23
ES6 Modules & Webpack: To improve the Webpack and ES6 lessons, more specifically to add more consolidated information in the lesson content since these are crucial lessons; also to improve the sequence of information in said lessons.


Advaitmenon1106:
### Checks

- [X] This is not a duplicate of an existing issue (please have a look through our [open issues list](https://github.com/TheOdinProject/curriculum/issues) to make sure)
- [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [X] Would you like to work on this issue?

### Key concerns covered in this issue (a TLDR)
  
- Additional points of information need to be added to the Webpacks and ES6 lessons by the means of a revision in order to consolidate information in one single place.
- The sequence of said additional information, and the existing information, possibly needs to be revised in order for it to make more logical sense.
- Potentially: This can be done in a more organised manner if the sole reliance of the curriculum content on the Webpack documentation/disjoint articles on bundlers and webpacks is reduced- in other words,  if the course stitches up the various sources of information on Webpacks to form its own lesson content on the same.

### Elaboration of the issue

- Currently, the Webpack lessons are an amalgamation of several bits of information, with the following flow:-

-> npm information <br/>
-> webpacks (the webpack doc tutorial is here and asks us to create a bundle using just index.js)<br/>
-> es6 import and export (Odin introduces import-export here and using multiple JS files)<br/>
-> webpacks continues (talks about bundling functionality, etc using separate js files) <br/>
-> Move to webpack lesson for loaders etc <br/>

| Part of the flow    | Source of information |
| -------- | ------- |
| NPM  | Lesson + "History of Javascript"    |
| Why Bundlers | Odin + https://snipcart.com/blog/javascript-module-bundler     |
| Webpack tutorial using 1 JS file | https://webpack.js.org/guides/getting-started/   |
| About Src and Dist | Odin |
| ES6: import-export | Odin |
| Loaders | Webpack (next lesson) (Adds a reference to the Webpack documentation once again) |

- The webpack documentation is process-oriented however, the rationale behind using a webpack project is lost in the process, and is only covered in the next section in the same Odin lesson (note another break in continuity, here).
- This problem can be fixed using a better thought-out sequence of information in the lesson.
- More information related to the so-called "dependency graph" built by Webpack can also be added using visual means and examples, in order to give a high-level overview of how Webpack works.
[...]

PR #27101, opened 2024-01-09
JS Linked Lists: Add mention of ES6 modules with Node

### LINKED ISSUE #26679: JS Computer Science: Mention running .js files in Node and import compatibility
### Describe your suggestion

In the JS comp sci section, in the recursion project there is a small instruction for running your code in Node. At this point, that likely won't cause issues as the files will be small so most will likely have everything they need to run in a single file.

For Linked List to Knight Travails, if someone decides to use modules (e.g. separating their Node and LinkedList/BinarySearchTree functions/classes, or perhaps their "driver script" that tests their code), there is a chance they may encounter issues by trying to use ESM syntax where Node does not yet support it by default (either must use `.mjs` file extension or include a `package.json` with `"type": "module"` in it, for example).

Learners will not have learnt about commonJS syntax yet. Previously, some people have asked in Discord about an error they got due to running a `.js` file using ESM.

I propose in the Linked Lists project adding a small lesson note that says something like:

> By default, Node uses something called CommonJS modules and not ES6 modules meaning unless you configure it to do so, it will not be able to understand your ES6 import/export syntax. For simplicity here, keep everything in a single file.

In Node 21.x (current, not yet LTS), this behaviour is looking to change where they now have an experimental flag that will allow Node to automatically switch to ESM if it detects it being used in the file (unless it is told specifically to use CJS e.g. `.cjs`). An interesting thing to look out for in the future!

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-linked-lists

### Checks

### (Optional) Discord Name

Mao

_No response_

MaoShizhong:
## Because
Currently, Node uses CommonJS by default and needs to be told in some way to use ESM instead.
Whilst this is simple, it's not made clear in the curriculum so far and so sometimes people wish to use multiple files in these comp sci projects but encounter issues with Node not recognising ESM by default.
This should add a small instruction for how to tell Node to use ESM and a little link that introduces the situation a little further.

## This PR
- Adds a mention of how to use ESM in Node

## Issue
Closes #26679

## Additional Information
Latest Node (non-LTS) has added experimental support for automatic module-type recognition which will be interesting in the future.

## Pull Request Requirements


wise-king-sullyman (commented):
I think we've largely tried to move away from linking to outside articles inline, I wonder if the linked resource would make sense in the additional resources section.

Also: I wonder if this addition makes the most sense in this lesson. Part of me thinks it should go in the es6 modules lesson, but the other part feels like that's to early. WDYT?

MaoShizhong:
@wise-king-sullyman My feeling is that the ESM lesson might be a little early. My understanding is that the ESM lesson is focused not just on what ESM is generally but primarily in the context of being used in a browser, hence the justification for webpack to bundle. 
This issue is more of a running modules in Node thing, which whilst related to ESM, I don't think is quite as related to the ESM lesson's objectives.

Hence the suggestion to add it where the there's the first likely instance of its relevance, in the Linked List project. I don't think many if any people would find a significant need to use multiple modules in the recursion project.

About the link/additional resources, given that the proposed addition is in the Linked List project itself which doesn't have an additional resources section (don't think any projects do?), I'm not sure if separating the link out from the note will be better? That's going by having the addition in the project content.

kamranzafar4343 (commented):
"Thank you for your contribution! I appreciate the effort you've put into this commit. The code looks clean and well-organized. Keep up the good work!"

wise-king-sullyman:
@MaoShizhong Hmm, I've been thinking about this a lot, and while yes the ESM lesson is more in the context of running in the browser it does also touch on NPM and the package file which ties into your note quite nicely IMO. To me it feels like a much better fit than having it in a project.

I do understand wanting to add it where it will be most relevant though. What would you think about adding the actual info and linking to the CJS vs ESM blogpost in the ESM lesson and having a note that calls back to it in the linked list project?

MaoShizhong:
@wise-king-sullyman Going through the ESM lesson again, what you're saying does make sense.
I'm aware that some ideas about restructuring the ESM/Webpack sections are being explored as per #26976 , which I'm very much in agreement with (which I have commented there about). 
I definitely think if the ESM and Webpack content were separated (with ESM coming first), then it would be far easier to integrate the ESM/CJS thing into the ESM portion, which can then be reminded about come the comp sci section.

I'm a little unsure exactly how I'd want to integrate the same content in the current rendition of the ESM lesson in terms of the flow, keeping it relevant in context, whilst also not risking overwhelming the learner in what's already commonly an overwhelming section (which I also think is in part related to the current structure).
I think I'm on board with you generally but will need some time to dwell on how I might want to make it happen.

wise-king-sullyman:
@MaoShizhong agree it would be a lot easier to fit this into the ESM lesson after that rework, I don't think it would be awful as is though, and I think it would make sense to just add it to the end of the `ES6 modules` section, with the link added to the additional resources.

If you have a better idea than that though I'm not hard committed to it.

MaoShizhong:
Almost all linting errors have been fixed. The remaining ones are ones I'm not sure how to resolve as there are no rules set for them (missing code block language for blocks that do not require a language, and links that link to a page section instead of an external link).

wise-king-sullyman (approved):
🔥

wise-king-sullyman:
Thanks for all your work on this!

PR #27120, opened 2024-01-13
Fix: Correct lesson "Space Complexity"'s explanation on O(1)

autotelico:
## Because
The wording appears to be slightly incorrect, inducing the reader to a mistake in conceptualizing Big O Notation.

## This PR
- Corrects the current statement in which it says two variables are created within the function, when in fact the function only returns a single variable with a value.

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/space_complexity.md
+++ b/javascript/computer_science/space_complexity.md
@@ -59,7 +59,7 @@ function multiply(num1, num2) {
 }
 ```
 
-Here it should hopefully be clear that no matter the arguments we pass to the function call, only a single variable is created. It doesn't change. Therefore, we can consider the space this takes is always O(1).
+Here it should hopefully be clear that no matter the arguments we pass to the function call, only a single value is created (the product of the numbers). It doesn't change. Therefore, we can consider the space this takes is always O(1).
 
 #### O(N) - Linear complexity
 
--- a/javascript/computer_science/space_complexity.md
+++ b/javascript/computer_science/space_complexity.md
@@ -59,7 +59,7 @@ function multiply(num1, num2) {
 }
 ```
 
-Here it should hopefully be clear that no matter the arguments we enter when we call the function, only two variables are created. It doesn't change. Therefore, we can consider the space this takes is always O(1).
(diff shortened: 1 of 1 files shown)


javascript/computer_science/space_complexity.md
```diff
@@ -59,7 +59,7 @@ function multiply(num1, num2) {
 }
 ```
 
-Here it should hopefully be clear that no matter the arguments we enter when we call the function, only two variables are created. It doesn't change. Therefore, we can consider the space this takes is always O(1).
+Here it should hopefully be clear that no matter the arguments we pass to the function call, only a single variable is created. It doesn't change. Therefore, we can consider the space this takes is always O(1).
```
JoshDevHub, line 62: I don't feel great about the word "variable" here. The return value isn't being declared as a variable. I think a wording like this would be preferable:
```suggestion
Here it should hopefully be clear that no matter the arguments we pass to the function call, only a single value is created (the product of the numbers). It doesn't change. Therefore, we can consider the space this takes is always O(1).
```
What do you think?

JoshDevHub (changes requested):
Thank you for making this PR.

Gonna request one change on the wording. Also, the ruby course also has this lesson. Would you be willing to add a commit to this PR that fixes this in the ruby lesson as well? You shouldn't have to know any ruby to do this.

javascript/computer_science/space_complexity.md
autotelico: Since I am just going through this lesson, I didn't want to change the original conveyed message of 'variable', but yeah, I do feel as weird about it as you do. But yeah, functions shouldn't return variables since they were never declared in the first place. It's a simple operation, as you've described.

About the PR on Ruby, sure! That would be a pleasure. I'll do it in a jiffy.

I like your suggestion. Also, considering I'm not very acquainted with replying with new suggestions here on GitHub, I wanted to add to the suggestion you've made by changing the _"It doesn't change"_ part for _"This value doesn't change."_, for extra clarity. If you'd show me how to go about that, that would be lovely.

JoshDevHub (approved):
Awesome stuff. Thank you for contributing :raised_hands:

PR #27121, opened 2024-01-13
Comp Sci: Hash Map Project: modify instructions for `remove` method

JoshDevHub:
## Because

The current instructions for the `remove` method don't contain an implementation detail that's common with hash maps: what should happen when the given key is not in the hash map?

Most implementations provide some return value so the developer can know if the item was deleted or not.

## This PR

- Changes the instructions for JS hash map to use [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete)'s implementation. If the given key is found and removed, return `true`, else return `false`.
- Changes the instructions for the ruby hash map to use [Hash](https://docs.ruby-lang.org/en/3.1/Hash.html#method-i-delete)'s implementation. If the given key is found and removed, return the deleted entry's value, else return `nil`.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_hash_map.md
+++ b/javascript/computer_science/project_hash_map.md
@@ -34,7 +34,7 @@ You already know the magic behind hash maps, now it's time to write your own imp
 
   1. `has` takes a key as an argument and returns `true` or `false` based on whether or not the key is in the hash map.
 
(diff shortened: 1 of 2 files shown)


01zulfi (approved):
lgtm :rocket:

PR #27146, opened 2024-01-17
Installation Lessons: Ignore lint rules regarding heading structure

JoshDevHub:
## Because

The lint rules for lessons don't make sense in the context of the installation guides that appear throughout the curriculum. The installation guides never have an "Assignments" section, and repeating heading content is natural with how they're organized.

## This PR

- Ignores the relevant rules regarding headings for each of the installation guides in the curriculum. This is done with a magic comment in the  markdown files.
- Cleans up various other lint errors that exist in the files that are changed.

## Issue

Closes #XXXXX

## Additional Information

This may not be the best thing to do long term, but I think it's fine for now. Another solution may be to consider installation content as another "type" of file (alongside lessons and projects), and then standardize what format they should take as well. But this seems like it would involve a larger conversation with the team.

## Pull Request Requirements


JoshDevHub:
There are current lint errors around using `` tags in the installations lesson. I wasn't really sure what to do about them. They're used to style the image `display: inline;` so that the icons being referenced can appear among the text:
example of whisker icon

Not sure if we want to carve out an exception for these or change the way this is done.

thatblindgeye (changes requested):
I feel like all of these lessons could probably be restructured to follow the typical lesson structure. The installation sections could be restructured as the "assignment" of the lesson, the lesson overview might just end up being "Installing some Cool Thing', and the knowledge check we might be able to throw in some easy things that could be added as an assignment item, like a resource on basics of whatevers being installed.

That can all be a followup, though. Can you just create an issue to look into all of these lessons further? We should ideally look into a) restructuring the lessons to follow the general lesson structure, if possible, and b) update headings so that we aren't repeating them (instead of 2-3x "Install Chrome' we could update them to "Install Chrome on [Linux/Windows/Mac]" for example).

Also, we can add the `img` tag to the array for rule MD033 if you'd want to go ahead and do that.

thatblindgeye:
@JoshDevHub Actually, looks like we can add classes to markdown images with Kramdown:

```
![Alt text](imgurl){: .classname}
```

@KevinMulhern can you confirm this is accurate? I just did this in the lesson preview link and it works, but confirmation would be good.

KevinMulhern:
Looks accurate to me @thatblindgeye

JoshDevHub:
Yeah I'll make an issue for exploring how to make the installation lessons fit the lesson format.

Re: the `img` tag thing -- I didn't know that Kramdown allowed classes from that. There's still a slight problem though. If I just try adding Tailwind utilities to get the same style, the margin won't work properly:

Example of the broken margins when applying `.inline.m-auto`:
Selection_353

I'm *pretty* sure this is happening because `m-auto` is being caught in the Tailwind purge (a quick `grep` tells me it's  not used anywhere in the main site's code). Now what we really care about is margin above and below, but `my-auto` doesn't apply either. 

`mt-auto` applies:
Selection_354

... but `mb-auto` doesn't, so that's still not going to work right. So to use this format, I think a change is going to have to be made on the main site repo to prevent one of these solution utilities from being purged.

Do you have any thoughts on that @KevinMulhern ?

KevinMulhern:
Nice debugging @JoshDevHub 💪. Any classes we need for the dynamic lesson content, we've put in Tailwinds [safe list here](https://github.com/TheOdinProject/theodinproject/blob/main/tailwind.config.js#L14).

It'd probably be worth creating a new utility class like `inline-img` to combine inline and mb-auto, just to make it simpler on the curriculum content side and have Tailwind decoupled from the content.

JoshDevHub:
Okay cool. The images can be done inline now without using raw HTML, so I converted over to that and it's passing md lint.

I'll go ahead and make the issue for converting the installation lessons over to a different format and pulling the ignore comments. I still think this fix would be a nice temporary solution to have due to the churn happening in some of the installation lessons lately (re: WSL).

thatblindgeye (approved):
Yep, as long as we aren't just disabling rules and forgetting about them it should be good.

PR #27236, opened 2024-01-30
Paragraph re-write stimulus_js.md

jbk2:
An attempt to more clearly explain Stimulus target properties.

### DIFF
--- a/ruby_on_rails/rails_sprinkles/stimulus_js.md
+++ b/ruby_on_rails/rails_sprinkles/stimulus_js.md
@@ -332,8 +332,8 @@ the questions below on your own, clicking the small arrow to the left of the que
 
 - By using abstract controllers, like `toggle`, rather than one specific to the view, like `reveal-comments`
 - By making them configurable with `values` and `classes`
-</details>
 
+</details>
 
 <details markdown="block">
   <summary>How do you trigger actions on an event?</summary>
--- a/ruby_on_rails/rails_sprinkles/stimulus_js.md
+++ b/ruby_on_rails/rails_sprinkles/stimulus_js.md
@@ -140,9 +140,10 @@ export default class extends Controller {
 ```
 
 You need to declare your targets, once you have Stimulus provides the following properties to you controller:
-* `this.nameTarget` - the first DOM element of matching target name.
-* `this.nameTargets` - An array of all targets matching name.
-* `this.hasNameTarget` - Boolean indicator of whether target of name is in scope.
+
+- `this.nameTarget` - the first DOM element of matching target name.
+- `this.nameTargets` - An array of all targets matching name.
+- `this.hasNameTarget` - Boolean indicator of whether target of name is in scope.
 
 ### Keeping state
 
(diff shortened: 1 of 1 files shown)


ruby_on_rails/rails_sprinkles/stimulus_js.md
```diff
@@ -139,10 +139,10 @@ export default class extends Controller {
 }
 ~~~
 
-You need to declare your targets, but once you do, you can get the DOM element by using `this.nameTarget`.
-If you need multiple targets of the same kind, you can get an array of DOM elements with `this.nameTargets`.
-If you need to make your controllers to be smart about targets, you can
-also ask whether certain targets are available in the HTML with `this.hasNameTarget`.
+You need to declare your targets, once you have Stimulus provides the following properties to you controller:
```
JoshDevHub, line 142:
```suggestion
You need to declare your targets. Once you have, Stimulus provides the following properties to you controller:
```

ruby_on_rails/rails_sprinkles/stimulus_js.md
```diff
@@ -139,10 +139,10 @@ export default class extends Controller {
...
+* `this.nameTarget` - the first DOM element of matching target name.
+* `this.nameTargets` - An array of all targets matching name.
+* `this.hasNameTarget` - Boolean indicator of whether target of name is in scope.
```
JoshDevHub, line 145: Per the [layout style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#unordered-lists), hyphen characters (-) should be used for unordered list items rather than asterisks.

JoshDevHub (changes requested):
I like your changes. Just a couple of small suggestions and we'll be good to merge.

Also, this file has several errors in it being flagged by Markdownlint. None of these were introduced by you, and they were already in this file prior to your changes. But would you be willing to go ahead and fix these issues now? You can click the "Files changed" tab and view the right column of the diff to see the different errors.

If you don't want to do that, that's okay. I can just go ahead and merge after you fix the small things with your changes. Just let me know :+1:

jbk2:
Sure Josh, I'll do my best to address those other points also 👍🏻.

jbk2:
Josh, I've corrected all of the flagged linting errors except one which doesn't make any sense to me. It keep flagging a 'Required heading structure' error on line 345 on which there is no heading, and the heading under which that line's content sits seems correctly formatted to me. Over to you to see if you can correct otherwise perhaps just merge anyway.

JoshDevHub:
> Josh, I've corrected all of the flagged linting errors except one which doesn't make any sense to me. It keep flagging a 'Required heading structure' error on line 345 on which there is no heading, and the heading under which that line's content sits seems correctly formatted to me. Over to you to see if you can correct otherwise perhaps just merge anyway.

Yeah the way the tool explains this error is difficult to understand. It's complaining because the `Knowledge checks` section should come before the `Additional resources` section. So if you swap the position of those two sections, everything should pass.

jbk2:
> Yeah the way the tool explains this error is difficult to understand. It's complaining because the `Knowledge checks` section should come before the `Additional resources` section. So if you swap the position of those two sections, everything should pass.

Aha, i see. Checks have passed so should be all sorted now. Over to you to authorise merge, thanks. 👍🏻

JoshDevHub (approved):
Thank you for contributing! And thanks for taking the time to fix all the little lint errors.

PR #27257, opened 2024-02-01
BST project: parameter definitions for methods with args

### LINKED ISSUE #27217: Binary Search Trees: Clearer distinction between node and value
### Checks

### Describe your suggestion

Many learners when they show their code snips for this project tend to think (which they show in their code) that nodes and value as the same thing. They will traverse the tree for a value to delete or get the height of, instead of taking a node. Their post/pre/in-order callbacks will accept a value as their argument, instead of a node. That sort of thing. It leads to a lack of understanding of the point of a tree, which is to arrange objects, not numbers, and to hard to repair code, since its rarely clear to the learner at any given moment what arguments should be of what type 

The lesson muddies the waters on this by stating that the delete function should accept a value instead of a node, when this isn't the case. The only functions that should have arguments of value are those that would construct a new node (insertions), or exactly the function find. All others should work exclusively with the nodes themselves. Then you can leverage find to return references to certain nodes from their value when appropriate (`delete(find(value))`). 
In fact it might even be more appropriate to say that insertions should take a node argument too, and have them be called with an explicitly constructed node, but I'd be willing to compromise on that one. Something like this: `insert(new Node(value))`

This to some extent may also apply to the linked list lesson, but it is most prevalent here. I think a reword of the assignment with this in mind, as well as a note earlier in the lesson talking about the difference between a node and a value could help.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-binary-search-trees

### (Optional) Discord Name

_No response_

_No response_

damon314159:
## Because
It could be clarified at time what types a function is expected to take as its argument, while stopping short of a full function signature

## This PR
Modelled off of the parameters given in the linked list project description:
Replaces `find` with `find(value)` and similar for each of the tree methods that expects an argument. Those that take no inputs are left as they were.

## Issue
Closes #27217

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_binary_search_trees.md
+++ b/javascript/computer_science/project_binary_search_trees.md
@@ -10,11 +10,11 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
 <div class="lesson-content__panel" markdown="1">
 
-1.  Build a `Node` class/factory. It should have an attribute for the data it stores as well as its left and right children.
+1. Build a `Node` class/factory. It should have an attribute for the data it stores as well as its left and right children.
 
-1.  Build a `Tree` class/factory which accepts an array when initialized. The `Tree` class should have a `root` attribute, which uses the return value of `buildTree` which you'll write next.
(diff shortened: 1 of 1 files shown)


JoshDevHub (approved):
Looks great. Thanks for making this.

PR #27268, opened 2024-02-03
Markdownlint - Add custom rules

thatblindgeye:
## Because

The default rules for Markdownlint were missing some things that our style guide mentions

## This PR

- Adds various custom rules, including a modified heading structure rule
  - TOP001 will catch inaccessible link text that commonly includes things such as "this video", "click here", etc. It won't say what to update the link text to, because the context is super important, but flagging it is a huge time saver
  - TOP002 does what it says: complains if we try using inline code in headings
  - TOP003 also does what it says: for sections like "lesson overview" where we are using default text over and over, the rule complains when that content doesn't match up. It'll catch if the post heading verbiage isn't there/is wrong, if there is no unordered list following that verbiage, and if there is extra content after or before the unordered list.
  - TOP004 is just a slightly tweaked version of the default linter's required-headings rule. Where the default rule only had a general wild card `*`, ours allows something like `#### *` to say "any heading is allowed here as long as it's a level 4 heading"

## Issue

Closes #XXXXX

## Additional Information

For the link text rule, it may catch some "false positives", such as something like `[This blog post on flex-grow from CoolBlogSite](...url)`. Originally I went with an arbitrary "throw an error if there's less than 5 words before and/or after instances of `this`", but like I said it's arbitrary and may not be any better. I figure for these cases, we should still update the link text to pull "this" out, so that we end up with `this blog post on [flex-grow from CoolBlogSite](...url)` or something.

## Pull Request Requirements

### DIFF
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -80,5 +80,8 @@
       "style": "asterisk"
     }
   },
-  "customRules": ["./markdownlint/descriptiveLinkText.js"]
+  "customRules": [
+    "./markdownlint/descriptiveLinkText.js",
+    "./markdownlint/noCodeInHeadings.js"
+  ]
 }
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -81,7 +81,8 @@
     }
   },
   "customRules": [
-    "./markdownlint/descriptiveLinkText.js",
-    "./markdownlint/noCodeInHeadings.js"
+    "./markdownlint/TOP001_descriptiveLinkText.js",
+    "./markdownlint/TOP002_noCodeInHeadings.js",
+    "./markdownlint/TOP003_defaultSectionContent.js"
   ]
 }
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -79,5 +79,6 @@
     "MD050": {
       "style": "asterisk"
     }
-  }
+  },
+  "customRules": ["./markdownlint/descriptiveLinkText.js"]
 }
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -2,87 +2,88 @@
 
 {
   "config": {
-    "default": true,
+    "default": false
     // heading-style
     // Enforces ATX style headings, e.g. ### A level 3 heading
-    "MD003": {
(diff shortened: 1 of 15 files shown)


markdownlint/TOP001_descriptiveLinkText.js
```diff
@@ -0,0 +1,35 @@
...
+    );
+    const linkOpenTokens = linesWithLinks
+      .map((line) =>
+        line.children?.filter((child) => child.type === "link_open")
+      )
+      .flat();
+
+    linkOpenTokens.forEach((token) => {
+      const { line } = token;
+      const isInvalid = /\[(.+\s*)?(this|here)(\s*.+)?\]/i.test(line);
```
thatblindgeye, line 17: Link to Regexr to show valid and invalid examples

markdownlint/TOP004_lessonHeadings.js
```diff
@@ -0,0 +1,155 @@
...
+    let hasError = false;
+    let anyHeadings = false;
+    const getExpected = () => requiredHeadings[i++] || "[None]";
+    const handleCase = (str) => (matchCase ? str : str.toLowerCase());
+    forEachHeading(params, (heading, content) => {
+      if (!hasError) {
+        anyHeadings = true;
+        const actual = levels[heading.tag] + " " + content;
+        const expected = getExpected();
+        if (/^#*\s?\*$/.test(expected)) {
```
thatblindgeye, line 101: Same as above here, also store as a new Regex

.markdownlint-cli2.jsonc
```diff
@@ -79,5 +79,11 @@
     "MD050": {
       "style": "asterisk"
     }
-  }
+  },
+  "customRules": [
```
thatblindgeye, line 83: Add some test markdown files that can be checked when custom rules are added/updated

thatblindgeye:
@wise-king-sullyman So bit of a hiccup with the testing MD files, at least in this PR. I tested each file locally and everything that should be flagged did get flagged, and anything that should NOT be flagged wasn't. There's some limitation with some of these rules (custom or default) when the GH action runs, though. I can try opening a separate PR for each of these test files and close them to show they work as intended.

wise-king-sullyman (approved):
I'm good with however you want to proceed. IMO followup PRs just doing whitespace changes or something to the test files that aren't getting flagged properly would be fine.

PR #27280, opened 2024-02-05
HashMap Data Structure Lesson: Improve formatting and clarify a key point

nikitarevenco:
## Because

I feel like the lesson could use some clarity for this example:

> Store the key value pair in that bucket

This was confusing to me when I first completed it. I couldn't understand what the value would be in this situation at first, so I think it wouldn't hurt to clarify this small detail. 

Additionally, I've improved the formatting for this paragraph: 

> To get a value using a key, we put each entry inside a bucket as a `Node` item, which holds both the key and the value. To retrieve the value, we hash the key and calculate its bucket number. If the bucket is not empty, then we go to that bucket and compare if the node's key is the same key that was used for the retrieval. If it is, then we can return the node's value. Otherwise, we return `null`.

When first reading it, I couldn't understand what it was saying. I had to re-read it many times because all of this information is just put into a single paragraph. Plus, putting it in this structure to be like a bullet point list will make it more consistent with the previous example of inserting the key. 

## This PR

Improve formatting of a single paragraph to make it easier to digest and to align with the style of the previous example. Plus, clarified a minor detail which could be the point of confusion. 

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/hash_map_data_structure.md
+++ b/javascript/computer_science/hash_map_data_structure.md
@@ -88,7 +88,7 @@ Buckets are storage that we need to store our elements. Simply, it's an array. F
 
 1. Pass "Fred" into the hash function to get the hash code which is `508`.
 1. Find the bucket at index `508`.
-1. Store the key value pair in that bucket, in this case the key would be "Fred" and the value we've not specified.
+1. Store the key value pair in that bucket, in this case the value would be `undefined` as we haven't specified it.
 
 This is an oversimplified explanation; we'll discuss more internal mechanics later in the lesson.
 
@@ -98,7 +98,7 @@ Now if we wanted to get a value using a key:
 1. To retrieve the value, we hash the key and calculate its bucket number.
 1. If the bucket is not empty, then we go to that bucket.
 1. Now we compare if the node's key is the same key that was used for the retrieval.
-1. If it is, then we can return the node's value. Otherwise, we return `null`. 
+1. If it is, then we can return the node's value. Otherwise, we return `null`.
 
 Maybe you are wondering, why are we comparing the keys if we already found the index of that bucket? Remember, a hash code is just the location. Different keys might generate the same hash code. We need to make sure the key is the same by comparing both keys that are inside the bucket.
 
--- a/javascript/computer_science/hash_map_data_structure.md
+++ b/javascript/computer_science/hash_map_data_structure.md
@@ -88,11 +88,17 @@ Buckets are storage that we need to store our elements. Simply, it's an array. F
 
 1. Pass "Fred" into the hash function to get the hash code which is `508`.
 1. Find the bucket at index `508`.
-1. Store the key value pair in that bucket.
+1. Store the key value pair in that bucket, in this case the key would be "Fred" and the value we've not specified.
 
 This is an oversimplified explanation; we'll discuss more internal mechanics later in the lesson.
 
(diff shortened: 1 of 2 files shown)


javascript/computer_science/hash_map_data_structure.md
```diff
@@ -88,11 +88,17 @@ Buckets are storage that we need to store our elements. Simply, it's an array. F
 
 1. Pass "Fred" into the hash function to get the hash code which is `508`.
 1. Find the bucket at index `508`.
-1. Store the key value pair in that bucket.
+1. Store the key value pair in that bucket, in this case the value would be `undefined` as we haven't specified it.
```
JoshDevHub, line 91: I think we could also change this section to mention a value to store. This would probably be less confusing? Do you have any thoughts @nikitarevenco ?

javascript/computer_science/hash_map_data_structure.md
```diff
@@ -88,11 +88,17 @@ Buckets are storage that we need to store our elements. Simply, it's an array. F
...
 This is an oversimplified explanation; we'll discuss more internal mechanics later in the lesson.
 
-To get a value using a key, we put each entry inside a bucket as a `Node` item, which holds both the key and the value. To retrieve the value, we hash the key and calculate its bucket number. If the bucket is not empty, then we go to that bucket and compare if the node's key is the same key that was used for the retrieval. If it is, then we can return the node's value. Otherwise, we return `null`.
+Now if we wanted to get a value using a key:
+
+1. Put each entry inside a bucket as a `Node` item, which holds both the key and the value.
+1. To retrieve the value, we hash the key and calculate its bucket number.
+1. If the bucket is not empty, then we go to that bucket.
+1. Now we compare if the node's key is the same key that was used for the retrieval.
+1. If it is, then we can return the node's value. Otherwise, we return `null`.
```
JoshDevHub, line 101: I like this section

JoshDevHub (commented):
I like this. Just one little comment I want your thoughts on.

javascript/computer_science/hash_map_data_structure.md
nikitarevenco: Yes, I think that's a good idea. What do you think of this:

Let's say we wanted to store a person's full name as a key "Fred" with a value of "Smith":

1. Pass "Fred" into the hash function to get the hash code which is `508`.
1. Find the bucket at index `508`.
1. Store the key value pair in that bucket. In this case, the key would be `508` and the value would be "Smith".

nikitarevenco: Alright, I've implemented it.

JoshDevHub (changes requested):
Looks great but I forgot to comment on one additional change I think you can go ahead and make.

This lesson also exists in the Ruby course with nearly the same content, just with the code examples given in Ruby instead of JavaScript.

Could you also add these changes to the Ruby lesson? You shouldn't have to know any Ruby code to do this.

nikitarevenco:
> Looks great but I forgot to comment on one additional change I think you can go ahead and make.
> 
> This lesson also exists in the Ruby course with nearly the same content, just with the code examples given in Ruby instead of JavaScript.
> 
> Could you also add these changes to the Ruby lesson? You shouldn't have to know any Ruby code to do this.

Yep, no problem. I added the same changes to the Ruby lesson too. 

Sidenote: I thought if I know JavaScript I'd not be able to read Ruby, but actually Ruby makes perfect sense to me despite having different syntax. I really understand now that it's not really about the language you choose to learn that matters, but rather how you understand general concepts and ideas, then it just comes down to learning the syntax to actually be able to implement something.

JoshDevHub (approved):
Awesome work!

And yeah, JavaScript and Ruby are very closely related and a lot of concepts transfer very well between the languages.

PR #27334, opened 2024-02-10
[Ruby] Basic Data Types: Update link from web archive to the new one

kapustindev:
## Because

This PR is a minor fix for a link that currently points to web.archive.org. It appears that the original link did not provide a 301 redirect to the new page, which uses hyphens instead of underscores in the URL. This fix ensures that the link directs to the correct location without relying on web.archive.org.

## This PR

- The link has been replaced from the web archive to the original source.

## Issue

A simple change that is not part of an existing issue

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [ ] If this PR addresses an open issue, it is linked in the `Issue` section
-   [x] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [x] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

PR #27416, opened 2024-02-22
Foundations Data Types: Warn about DOM manipulation code before taught

MaoShizhong:
## Because
One of the MDN lessons on strings that the learner is instructed to code along to includes DOM manipulation code. At this point in the course, the learner will not yet have been exposed to the need to defer JS execution until after parsing the HTML body, and so will be very easy to [run into the classic `cannot read properties of null` error](https://discord.com/channels/505093832157691914/690590001486102589/1209565728752140400). 

A note has been added to the relevant bullet with instructions for making the JS run after the querySelected elements have loaded, as well as a mention that this will be covered properly in a later lesson.

## This PR
- Adds a note about deferring the DOM manipulation JS from one of the linked tutorial's examples.
- Replaces many links with descriptive text.

## Issue
N/A

## Additional Information
There is a single linting error on the first line regarding rule `TOP003` and an "incorrect value of `lineNumber`". I was unable to find the exact cause of this/a remedy but #27334 mentions how this is an expected error and thus can be left alone in this case.

## Pull Request Requirements


MaoShizhong:
@thatblindgeye I have made those linting and link text changes as requested.
After some thought, I also agree that skipping the `Concatenation in context` section would be less confusing overall without much if any loss to the tutorial content, so I have also replaced the note with a skip instruction for that section. Definitely feels a little more streamlined too.

How are you thinking the [how to run JavaScript code](https://www.theodinproject.com/lessons/foundations-variables-and-operators#how-to-run-javascript-code) lesson's code blocks should be amended? I can open a separate issue/PR on this if you wish.

thatblindgeye:
@MaoShizhong Probably a few different ways:

- Add a comment to the codeblocks themselves, `// Be sure to place your script at the end of the body element, after all your HTML content` or something along those lines
- Place some text inline with the instructions somewhere, like `Type the basic HTML skeleton into a file on your computer somewhere, being sure to place your script after all of your HTML content`
- Add a notebox titled "Location matters" or something, that kind of goes into what you originally had in this PR, at the end of the section

I might lean towards the notebox since it would bring more attention to it, but if you'd like to open another PR I'd trust you to amend it however makes more sense to you

MaoShizhong:
@thatblindgeye Hmmm. Of the three options, I think the notebox style feels more "whole" as more of a "important details!" and less of an inline "btw do this as well".

However now I'm thinking that if we're now telling people to skip the DOM manip part of the MDN strings tutorial, would we even need to mention the "location matters" bit in the previous lesson?

Because the location/deferment, as far as the course goes, only really matters for the purpose of DOM manipulation. And in the DOM manipulation lesson there already is a [noteworthy section about the location/deferment of a script](https://www.theodinproject.com/lessons/foundations-dom-manipulation-and-events#:~:text=Your%20JavaScript%2C%20for,parsed%20and%20created.).
I feel like given this is already mentioned where most relevant, mentioning it again before you do anything where it's relevant just feels like unnecessary duplication?

thatblindgeye:
Oh good point. Yeah I agree, telling users to skip that task in this lesson means they really don't need to yet know about placement of their script tag, so I'm good with just keeping things as-is.

thatblindgeye (approved):

PR #27576, opened 2024-03-07
JS Form Validation: Clarify form validation requirements

### LINKED ISSUE #27316: JS Form Validation: Clarify intended approach with JS-only validation
### Checks

### Describe your suggestion

The JS form validation lesson teaches the constraint validation API.
The assignment in the practice section of that lesson indicates
> The form doesn’t need to actually submit, but you should give an error message if the button is pushed with any active errors or unfilled required fields. For the sake of this lesson, make sure all of the validation occurs in the JavaScript file. If all is well and the form is “submitted”, give the user a high five.

I have seen some confusion over exactly what this expects of the learner, with the following two interpretations being the most common:
1. Set `novalidate` on the `<form>` element, then use the appropriate form inputs in the HTML. Validate via the constraint validation API in the JS file.
2. Use *only* `<input type="text">` i.e. cannot use even stuff like `type="number"`/`type="email"` and *all* validation must be 100% custom in JS. This would mean for something like an email input, instead of using an email input and utilising `reportValidity` and other features of the constraint validation API, you use a text input and manually parse the input string for whether it conforms to an email pattern, then report the validity on that.

I think the current lesson wording is unclear as to which it expects the learner to do. Personally, I'm inclined to believe it intends the former, where it is fine to use different input types, just `novalidate` the `<form>` and use the constraint validation API as taught (so you won't have to manually parse a string to conform to a specific pattern for basic things like number input/email input etc.).

Can this be clarified and whether it is agreed the current wording can be improved to make the intent clearer?

### Path

Ruby / Rails, Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-javascript-form-validation-with-javascript#practice

### (Optional) Discord Name

_No response_

_No response_

MaoShizhong:
## Because
A surprisingly common occurrence I've seen in the community is confusion as to what the [Form Validation with JS lesson](https://www.theodinproject.com/lessons/node-path-javascript-form-validation-with-javascript)'s instruction to do "all" validation in JS actually means.

Some have interpreted this to mean that you cannot even use different input types with the form having the `novalidate` attribute. Meaning they use only text inputs and manually validate every small detail inc. input format.

Instead, I believe the intent is to disable automatic HTML validation and to handle it via JS instead, where the learner can still use different input types (just needing to use the constraint validation API to check and report validity).

## This PR
- Adds a instruction to set `novalidate` on the `<form>` element, and use JS to check and report validity via the constraint validation API as was taught in the lesson.
- Fixes linting errors.
- Combined the `### Practice` block into the `### Assignment` block.

## Issue
Closes #27316

## Additional Information
The decision to combine `### Practice` and `### Assignment` came from mdlint not liking `Practice` being a `###` instead of `####`. However, setting it as a `####` looks odd given it's a block heading where the block already has `####` subheadings.

The `Practice` material is also essentially an assignment anyway, so I felt it might as well be combined with the `Assignment` section as well.

I did this combination in its own commit - so if this would preferably be done in a separate PR, I can always revert it to make the main change for `novalidate` and open a new PR for the section combination. Equally, if it would be preferable to just keep the section structure separated as is and ignore the `### Practice` lint error via a comment, a revert could easily handle that as well.

## Pull Request Requirements

### DIFF
--- a/javascript/javascript_in_the_real_world/form_validation_with_javascript.md
+++ b/javascript/javascript_in_the_real_world/form_validation_with_javascript.md
@@ -22,17 +22,11 @@ This section contains a general overview of topics that you will learn in this l
 
 1. For reference, [W3Schools' page on the JavaScript validation API](https://www.w3schools.com/js/js_validation_api.asp) covers things in a more concise format. These functions were explained in the previous article. Typically, with HTML forms, the inputs are validated upon form submission, but you can use these functions to check validity whenever you like (such as when a user clicks or tabs out of a specific input field).
 
-</div>
-
-### Practice
-
-<div class="lesson-content__panel" markdown="1">
-
 #### Warmup
 
 Go back to your 'Library' project and add validation to that form! Don't let your users submit without filling in all the fields! Don’t forget to use the git branch workflow you learned in [Revisiting Rock Paper Scissors](https://www.theodinproject.com/lessons/foundations-revisiting-rock-paper-scissors) from Foundations to work on a new feature.
 
-### Project
+#### A little more practice
 
 Build a browser form which collects Email, Country, Zip Code, Password and Password Confirmation fields.  It should use live inline validation to inform the user whether a field is properly filled in or not.  That means highlighting a field red and providing a helpful error message until it has been filled in properly.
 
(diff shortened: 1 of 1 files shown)


bycdiaz (approved):
Thanks for looking into this!

I noticed the numbers are all 1 for all the numbered lists. Guessing that's not intentional?

MaoShizhong:
> Thanks for looking into this!
> 
> I noticed the numbers are all 1 for all the numbered lists. Guessing that's not intentional?

That's intentional due to us now using lazy numbering as per our [style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#lazy-numbering), and also enforced by our markdownlint rules.

PR #27607, opened 2024-03-13
Javascript -> Computer Science -> Project: Knight's Travails: Add resource link, update numbering

Slashmyaxl:
## Because

Knight's Travails deals with a data structure (i.e., a graph) which TOP only links to reading instruction for students to learn from before diving right in. While I do appreciate this way of learning, it makes for a very challenging project. For me and at least a few others I've noted, coming up with the algorithm was a point of particular anguish.

My solution? More reading instruction! Albeit with a dash of interactivity and practice should the user follow through to the end of the lesson. The Khan Academy resources already linked in this lesson are great, and this follow-up with the "breadth-first" keyword omitted from the hyperlink allows the student only to utilize it if they too are struggling.

Credit to Youtuber @ACoupleCodes (https://www.youtube.com/@ACoupleCodes) for being the one to lead me in this direction.

## This PR

- Added link to resource in the assignment section's instructions, sub-point to instruction number 3.
- Updated numbering to follow current TOP markdown standard (i.e. 1., 1., 1... instead of 1., 2., 3...)

## Issue

N/A

## Additional Information

N/A

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_knights_travails.md
+++ b/javascript/computer_science/project_knights_travails.md
@@ -29,10 +29,10 @@ Sometimes _there is more than one fastest path_. Examples of this are shown belo
 
 <div class="lesson-content__panel" markdown="1">
 1. Think about the rules of the board and knight, and make sure to follow them.
-2. For every square there is a number of possible moves, choose a data structure that will allow you to work with them. Don't allow any moves to go off the board.
-3. Decide which search algorithm is best to use for this case.  Hint: one of them could be a potentially infinite series.
-  1. If you're having trouble, [here](https://www.khanacademy.org/computing/computer-science/algorithms/breadth-first-search/a/breadth-first-search-and-its-uses) is a great resource to help guide you and allow you to practice with traversing graphs.
-3. Use the chosen search algorithm to find the shortest path between the starting square (or node) and the ending square.  Output what that full path looks like, e.g.:
+1. For every square there is a number of possible moves, choose a data structure that will allow you to work with them. Don't allow any moves to go off the board.
+1. Decide which search algorithm is best to use for this case.  Hint: one of them could be a potentially infinite series.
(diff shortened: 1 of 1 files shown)


MaoShizhong:
@TheOdinProject/ds-a Thoughts?

I don't know (as in actually don't know, not as in doubting) whether we want to give away the search algorithm to be used there. Especially given that the link would have to be made accessible and I'm not sure how it could be worded without including the name of the algo in the link text.

(Also unsure how the box model file ended up included in the first commit despite no changes made in it?)

Slashmyaxl:
@MaoShizhong re: the box model file it might have been similar to my other PR in that I branched off of a previous sub-branch... apologies!

As far as the issue at hand, I was admittedly uncertain at first as it is almost _too_ good of a hint. However, my argument is that even though the link all but gives away the _type_ of search algorithm best suited for the solution, it doesn't fully solve the problem for the student who is stuck and chooses to click the link. The user must still think/learn about how to traverse a graph in this way, as opposed to a more simple binary tree. I believe the link is valuable in that it also allows interactive practice.

In fact, I'm drawing similarities to the "Building a House from the Inside Out" article linked in Tic-Tac-Toe. If a user follows the links in that article, the JS code to create a Connect 4 game is _presented_ to them in full. They need only simplify the code into a game of TTT and design a UI. In the case of Knight's Travails, a user would still need to greatly expand upon what they _figured out_ in the practice problem to yield the desired output.

If it goes through, I'm open to ideas on the link text. Right that "here" doesn't suit the standard. Perhaps something like, "If you're stuck, take a look at this resource on **graph traversal** for guidance?"

CouchofTomato:
I'll seek some feedback from the team if we want to do hints and tips in this way.

nikitarevenco:
This could have the opposite effect, the learner could have a completely different idea about how they want to complete this project, so this could give them doubts and stop them from trying their own ideas

CouchofTomato:
Yeah, the view of the team seems to be that this will also cause people to jump to looking too early rather than really try and think it through. I don't think we have an agreed way to handle tips and tricks so for now I don't think we can merge this. Really do appreciate your contribution though.

PR #27621, opened 2024-03-16
React Router: Introduce outlet context and remove component-based router

### LINKED ISSUE #26502: React router/context: Adjust router component example and mention context earlier
### Describe your suggestion

In the React Router lesson, there is an [example of refactoring the browser router into its own component](https://www.theodinproject.com/lessons/node-path-react-new-react-router#refactoring-the-routing) which will allow it access to hooks that can be passed into the elements inside it. At the time of writing the lesson, it was said that the docs for React Router had this example as the way to do things.

However, it seems that they have changed how things work rather quietly and now explicitly advise that browser routers should be singletons and not placed inside any component at all, meaning they also will not and should not have direct access to hooks. This is confirmed by the React Router team in [this GH discussion](https://github.com/remix-run/react-router/discussions/10440) and as such, the docs have since been updated to keep the browser router as a singleton outside of any component within what will likely be `Main.jsx`. Basically just like the current lesson up until the "refactoring to a component" section.

Because it's not within a component anymore, their advice is to manage state for its child components using something like the Context API, which currently is placed after the shopping cart project. Outlets also have an even easier context provider baked in as a prop, callable via `useOutletContext` (though context will still need to be introduced in some capacity for this).

Because of this now-direct relationship with how the router should be used with context (certainly in regards to the course content), I am proposing that the refactoring to a component part should be removed, a mention that shared state management between elements in the router ought to be handled with Context (which will be covered in a later lesson), then the Context lesson brought back to some point before shopping cart but after router. I am not 100% sure at the moment where I feel the best place for it would be.

Would love to hear some insights on this? I only came to know about this change in the advised implementation from [this discord conversation](https://discord.com/channels/505093832157691914/664184123061174274/1152435834184794112). If not for that, I would not have known and it's likely many others would not have either.
@ManonLef has directly encountered issues with state management within the router following the pattern shown in the router lesson as per the [conversation following this message](https://discord.com/channels/505093832157691914/505093832157691916/1162839867613515918).

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-react-new-react-router and https://www.theodinproject.com/lessons/node-path-react-new-managing-state-with-the-context-api

### Checks

### (Optional) Discord Name

Mao

_No response_

MaoShizhong:
## Because
The current [React Router lesson](https://www.theodinproject.com/lessons/node-path-react-new-react-router) demonstrates creating a Router component that contains a router, which would allow for things like props/state to be passed in to any of the router route's elements.

This is advised against by the React Router team who advise [routers should be singletons where data passed from parent to child in nested routes should be done via outlet context](https://github.com/remix-run/react-router/discussions/10440). The React Router docs have since been updated to match this advice, as opposed to the component version present at the time the lessons were written.

Having a component-based router and using state/props to pass data to nested routes/outlet components has often led to issues caused by the Router component rerendering and thus generating a completely fresh router with new components with new keys.

Therefore, the Router component example needs to be removed, and outlet context needs to be introduced at some point before the Shopping Cart project.

The current [Context lesson](https://www.theodinproject.com/lessons/node-path-react-new-managing-state-with-the-context-api) is written specifically with the Shopping Cart project in mind. Its examples and framing are all based on that project. Bringing the context lesson forward to be before that project would require far more significant changes to that lesson:
 - It can no longer reference the "shopping cart project you just did"
 - Its examples need be changed from Shopping Cart components to also not potentially solve some problems learners might face in that project.

Therefore, to minimise how disruptive any changes would be, the decision was made to leave the lesson order as is but to only introduce outlet context in the React Router lesson. Awareness and research into outlet context is by far one of the most common issues and resolutions in the `#react-help` channel in the discord server. Its usage is also straightforward (a built in `context` prop to `<Outlet />` and a simple `useOutletContext()` call in whatever descendent component within that outlet). There should not need to be much more of a significant change made than just introducing this concept and hook before the Shopping Cart project.

## This PR
- Changes the router refactoring section in the router lesson to extract only the routes array to a separate file, as opposed to creating a new Router component.
- Adds an "Outlets and state" section to the router lesson that introduces outlet context.
- Tweaks the post-Shopping Cart Context lesson's introduction to accommodate for the now-earlier mention of outlet context.
- Fixes linting errors in both affected files
- Converts React Router's knowledge check links from anchor tags to markdown links in line with the style guide and other lessons.

## Issue
Closes #26502

## Additional Information
N/A

## Pull Request Requirements

### DIFF
--- a/react/more_react_concepts/managing_state_with_context_api.md
+++ b/react/more_react_concepts/managing_state_with_context_api.md
@@ -115,9 +115,9 @@ This is a very basic application, but imagine the application grows in size as m
 
 To simplify our application and reduce complexity, we can implement the Context API. There are three key elements in this API that we need to understand:
 
-1. `createContext` - This "creates the context" Duh... But yes, it's how we can create the context. It takes in any value, be it a number, string, or object, which can be referred to as the _default value_ of the context, and returns a context object that can be used to pass down data to components
-2. `useContext` - This hook is used to consume data from a context object created by `createContext`. We can use this hook inside our component to retrieve the data that we need. This hook accepts the context object as an argument
-3. `ContextObject.Provider` - The context object comes with the `Provider` component that accepts a prop called `value`, which is the context value that's going to be passed down to the components no matter how deeply they're nested. In other words, a way to "provide" the context value to these components
+1. `createContext` - This "creates the context" Duh... But yes, it's how we can create the context. It takes in any value, be it a number, string, or object, which can be referred to as the *default value* of the context, and returns a context object that can be used to pass down data to components
+1. `useContext` - This hook is used to consume data from a context object created by `createContext`. We can use this hook inside our component to retrieve the data that we need. This hook accepts the context object as an argument
+1. `ContextObject.Provider` - The context object comes with the `Provider` component that accepts a prop called `value`, which is the context value that's going to be passed down to the components no matter how deeply they're nested. In other words, a way to "provide" the context value to these components
 
 We can start by using the `createContext` function that can be imported from the `react` module.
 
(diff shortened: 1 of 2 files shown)


react/more_react_concepts/managing_state_with_context_api.md
```diff
@@ -1,6 +1,6 @@
 ### Introduction
 
-Throughout the lessons, we've learned how to manage state and pass data/props around components. However, as our application grows in size, not only will this process of managing states and passing data around become repetitive and inconvenient, but it will also be hard to manage. This lesson will cover how we can reduce these complexities in our application by using the Context API.
+Throughout the lessons, we've learned how to manage state and pass data/props around components. However, as our application grows in size, not only will this process of managing states and passing data around become repetitive and inconvenient, but it will also be hard to manage. In the [React Router lesson](https://www.theodinproject.com/lessons/node-path-react-new-react-router#outlets-and-state), we mentioned how we could use outlet context to pass data from a parent component through an `<Outlet />` to a child component. This lesson will cover how we can use the Context API to do similar things outside of outlet scenarios.
```
cakegod, line 3:
```suggestion
Throughout the lessons, we've learned how to manage state and pass data and props between components. However, as our application grows in size, not only will this process of managing states and passing data around become repetitive and inconvenient, but it will also be hard to manage. In the [React Router lesson](https://www.theodinproject.com/lessons/node-path-react-new-react-router#outlets-and-state), we mentioned how we could use outlet context to pass data from a parent component through an `<Outlet />` to a child component. This lesson will cover how we can use the Context API to do similar things outside of outlet scenarios.
```

cakegod (commented):
Some suggestions.

wise-king-sullyman (approved):
🔥

PR #27671, opened 2024-03-23
Objects and Object Constructors: Add article explaining object references

takinabradley:
## Because

Learners don't often discover how object references behave differently until the Tic-Tac-Toe project or the Todo project. I'm unsure if there's much value in keeping this information from learners. 

I find myself introducing this concept to learners for the first time often as a helper.

## This PR

- Adds a recommendation to read JavaScript.Info's [Object references and copying](https://javascript.info/object-copy) article for a short and sweet introduction to the differences between object references and primitive values.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/objects_and_object_constructors.md
+++ b/javascript/organizing_your_javascript_code/objects_and_object_constructors.md
@@ -36,6 +36,8 @@ myObject[variable]; // this is equivalent to myObject['property'] and returns 'V
 
 If you are feeling rusty on using objects, now might be a good time to go back and review the content in our [object basics lesson](https://www.theodinproject.com/lessons/foundations-object-basics) from our JavaScript Basics course.
 
+You should also read JavaScript.Info's article [Object references and copying](https://javascript.info/object-copy) to learn more about how objects in JavaScript behave differently than primitive values you may be used to using up to this point in the curriculum.
+
 ### Lesson overview
 
 This section contains a general overview of topics that you will learn in this lesson.


javascript/organizing_your_javascript_code/objects_and_object_constructors.md
```diff
@@ -36,6 +36,8 @@ myObject[variable]; // this is equivalent to myObject['property'] and returns 'V
 
 If you are feeling rusty on using objects, now might be a good time to go back and review the content in our [object basics lesson](https://www.theodinproject.com/lessons/foundations-object-basics) from our JavaScript Basics course.
 
+You should also read JavaScript.Info's article [Object references and copying](https://javascript.info/object-copy) to learn more about how objects in JavaScript behave differently than primitive values you may be used to using up to this point in the curriculum.
```
thatblindgeye, line 39: We shouldn't really be linking to resources like this in an introduction of a lesson. Due to the context of this lesson this topic may not really be ideal here, but rather the object basics lesson linked on line 37. Though the resource you link goes into a bit more than we may want users to dive into (the content on object assign and structuredClone).

takinabradley: An option might be to add a section to the beginning of the lesson, something like:
```markdown
### Object references

You may or may not have noticed that objects and object-like data types in JS behave differently than the primitive values you may be used to. Read up about object references in JavaScript.Info's [Object references and copying](https://javascript.info/object-copy) article 
```

Or, I could make a PR to add this resource for the lesson in Foundations, if you don't feel it's too early to explain and have users exposed to `Object.assign` and `structuredClone` in the context of this article.

thatblindgeye: I still think the lesson in Foundations would be a better spot for the concept of references. If there's a resource that covers it without going into stuff like assign and structuredClone that'd be even better, but if we explicitly tell users to only read X/not to read Y then it isn't the worst thing.

PR #27676, opened 2024-03-24
Styling React Applications - Add additional resource

nikitarevenco:
## Because

I found this resource that changed my perspective on styled components

## This PR

- Adds additional resource that goes very in depth on styled components

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/react/the_react_ecosystem/styling_react_applications.md
+++ b/react/the_react_ecosystem/styling_react_applications.md
@@ -58,4 +58,5 @@ This section contains questions for you to check your understanding of this less
 
 This section contains helpful links to related content. It isn’t required, so consider it supplemental.
 
-- [Tailwind CSS documentation](https://tailwindcss.com/)
+- [Tailwind CSS documentation](https://tailwindcss.com/).
+- We recommend checking out this article that goes [very in-depth on some amazing things](https://www.joshwcomeau.com/css/styled-components/)https://www.joshwcomeau.com/css/styled-components/ you can do using styled components. 
--- a/react/the_react_ecosystem/styling_react_applications.md
+++ b/react/the_react_ecosystem/styling_react_applications.md
@@ -59,4 +59,4 @@ This section contains questions for you to check your understanding of this less
 This section contains helpful links to related content. It isn’t required, so consider it supplemental.
 
 - [Tailwind CSS documentation](https://tailwindcss.com/).
-- We recommend checking out this article that goes [very in-depth on some amazing things](https://www.joshwcomeau.com/css/styled-components/)https://www.joshwcomeau.com/css/styled-components/ you can do using styled components. 
+- Josh Comeau has a nice article on [getting the most out of styled-components](https://www.joshwcomeau.com/css/styled-components/). 


react/the_react_ecosystem/styling_react_applications.md
```diff
@@ -58,4 +58,5 @@ This section contains questions for you to check your understanding of this less
 
 This section contains helpful links to related content. It isn’t required, so consider it supplemental.
 
-- [Tailwind CSS documentation](https://tailwindcss.com/)
+- [Tailwind CSS documentation](https://tailwindcss.com/).
+- We recommend checking out this article that goes [very in-depth on some amazing things](https://www.joshwcomeau.com/css/styled-components/)https://www.joshwcomeau.com/css/styled-components/ you can do using styled components.
```
MaoShizhong, line 62:
```suggestion
- Josh Comeau has a nice article on [getting the most out of styled-components](https://www.joshwcomeau.com/css/styled-components/). 
```
I think the article is a good read and also reasonable considering we have the Tailwind docs in the additional resources.

I do not think we should be telling learners that "we recommend reading this article" in this case:
- It's an additional resource. We should not be recommending that people read something which we state is not required. And I don't think the article is appropriate for an assignment.
- Given that TOP still for the most part recommends as much vanilla CSS as possible and these are only options of varying abstractions from vanilla CSS (e.g. CSS modules is almost the same but is very slightly different), we should not be using wording that suggests TOP recommending the use of any other particular option.

We ideally should just, especially for additional resources, state what the resource is and what it's for then let the learner decide if it's relevant enough to read.

MaoShizhong (changes requested):
An important comment below about wording for resources we provide.

Would you also mind fixing up the linting errors here too?

react/the_react_ecosystem/styling_react_applications.md
MaoShizhong: @nikitarevenco Just checking in in case you haven't seen this! Let me know if there are any issues.

nikitarevenco: Oh thanks! I must've forgotten about this. I have added your suggestions but it seems that there aren't any linting issues here. What I am seeing instead is some files that are unrelated to this PR and the failing checks for them. 

image

nikitarevenco: Usually there would be errors here. For example you can see that the backtick is used insead of a single quote on line 59. But the linter doesn't seem to be flagging that?

image

MaoShizhong: I'm not entirely sure why the linter isn't picking that specific error up, given that your `patch-69` branch looks like it does contain the TOP003 rule config.

As for the check annotations stuff, that's just a super annoying beta feature of GitHub that we're not sure exactly why it's happening. But sometimes with older PRs, when an update is made, it runs the actions on unrelated files to try and be helpful....but it really isn't helpful here and is just noise.

You can ignore those.

MaoShizhong (approved):
To fix any remaining linting errors in this file, because it seems the linter's gone crazy here with that check annotations feature, it'd be best to do it in a separate PR.

I'd advise you do this locally, rather than making a "patch" branch via `Improve on Github`. The latest repo has a new repo-specific contribution guide with the instructions to run the lint check and fix scripts - you can use that on this file to fix any errors, and highlight any other errors than might need manual fixes.

Then you can open a new PR with those lint fixes, and link this PR in it as justification for it.

Issue #27681, opened 2024-03-25
Foundations: Provide descriptive link text


MaoShizhong:
Links in our lessons should be sufficiently descriptive to improve accessibility. As per the [layout style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#links), link text should both concisely describe its purpose (or where the link will redirect a user), and must not include the words `this` nor `here`.

## Guidance

- Comment on this issue to let us know you'd like to get a section assigned to you. You can also ask for a specific one.
- Please wait until you are assigned by a maintainer before starting to work on this issue to avoid multiple people working on the same lessons.
- When you have been assigned a section, please refer to the [suggested link text changes spreadsheet](https://docs.google.com/spreadsheets/d/1QyXIBLL1pEFtgY7vzi7C_cmJEUKUlrfQW_lQAFrWuzE/edit?usp=sharing) for instructions on what changes need to be made for your assigned lesson(s).
  - For most of these, you should just be able to copy and paste the suggested text to replace the original text shown in the spreadsheet (which may be more than just the link itself). Any additional steps/clarifications will be noted by the appropriate link(s).
  - You are welcome to use a different wording from a given suggestion if you find a better one. If you do, you must ensure the link text is consistent if there are multiple instances of the same href in the same lesson.
  - You are *not* required to fix any unrelated, preexisting linting errors such as list styles/indentation. Existing linting errors may cause certain tests to fail when you open a PR. You can ignore this and a maintainer will let you know if further action is required.
- Make sure to read the [general contributing guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md) as well as the [curriculum repo contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md) before starting your first contribution.
- Please add the following to your PR in the Issues section: `Related to #XXXXX` , since this issue is divided into sections, **individual PRs should NOT INCLUDE the `closes` keyword.**

### Introduction

- [x] @Ac394 [Intro to Web Development](https://www.theodinproject.com/lessons/foundations-introduction-to-web-development) && [Motivation and Mindset](https://www.theodinproject.com/lessons/foundations-motivation-and-mindset) && [Join the Odin Community](https://www.theodinproject.com/lessons/foundations-join-the-odin-community)

### Prerequisites

[...]

PR #27708, opened 2024-03-29
Fix link text

Crn0:
## Because

The link text should be sufficiently descriptive to improve accessibility.

## This PR

- Made the links text more descriptive for better accessibility by following the [layout style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#links)

## Issue

Related to #27681

## Pull Request Requirements

### DIFF
--- a/foundations/html_css/css-foundations/inspecting-html-and-css.md
+++ b/foundations/html_css/css-foundations/inspecting-html-and-css.md
@@ -36,7 +36,7 @@ The Styles panel also allows you to edit styles directly in the browser. You can
 
 1. Go through the following sections of the [official Chrome DevTools docs](https://developers.google.com/web/tools/chrome-devtools):
 
-   - [Devtools overview](https://developer.chrome.com/docs/devtools/overview/): don't navigate to any other pages linked here; just get familiar with _what_ tools are available in the DevTools, rather than how to use all of them right now.
+   - [Devtools overview](https://developer.chrome.com/docs/devtools/overview/): don't navigate to any other pages linked here; just get familiar with *what* tools are available in the DevTools, rather than how to use all of them right now.
    - [Open Chrome DevTools](https://developer.chrome.com/docs/devtools/open/): similar to what we went over above, but with some nice extras.
    - [Get Started With Viewing And Changing The DOM](https://developer.chrome.com/docs/devtools/dom/): skip through any part that uses the JavaScript console.
    - [View and change CSS](https://developer.chrome.com/docs/devtools/css): be sure to follow along with any interactive instructions!
--- a/foundations/html_css/css-foundations/inspecting-html-and-css.md
+++ b/foundations/html_css/css-foundations/inspecting-html-and-css.md
@@ -53,6 +53,6 @@ The following questions are an opportunity to reflect on key topics in this less
 
 ### Additional resources
 
-This section contains helpful links to related content. It isn’t required, so consider it supplemental.
+This section contains helpful links to related content. It isn't required, so consider it supplemental.
 
-- This article about how we can [utilize css overview in the developer tools](https://www.freecodecamp.org/news/how-to-use-css-overview-in-chrome-developer-tools/) to check the colors, font styles, media-queries, etc. used on a particular webpage.
+- This article about how we can [utilize CSS overview in the developer tools](https://www.freecodecamp.org/news/how-to-use-css-overview-in-chrome-developer-tools/) to check the colors, font styles, media-queries, etc. used on a particular webpage.
--- a/foundations/html_css/css-foundations/inspecting-html-and-css.md
+++ b/foundations/html_css/css-foundations/inspecting-html-and-css.md
@@ -35,7 +35,7 @@ The Styles panel also allows you to edit styles directly in the browser. You can
 <div class="lesson-content__panel" markdown="1">
 1. Go through the following sections of the [official Chrome DevTools docs](https://developers.google.com/web/tools/chrome-devtools):
 
-   - [Overview](https://developer.chrome.com/docs/devtools/overview/): don't navigate to any other pages linked here; just get familiar with _what_ tools are available in the DevTools, rather than how to use all of them right now.
+   - [Devtools overview](https://developer.chrome.com/docs/devtools/overview/): don't navigate to any other pages linked here; just get familiar with _what_ tools are available in the DevTools, rather than how to use all of them right now.
    - [Open Chrome DevTools](https://developer.chrome.com/docs/devtools/open/): similar to what we went over above, but with some nice extras.
    - [Get Started With Viewing And Changing The DOM](https://developer.chrome.com/docs/devtools/dom/): skip through any part that uses the JavaScript console.
    - [View and change CSS](https://developer.chrome.com/docs/devtools/css): be sure to follow along with any interactive instructions!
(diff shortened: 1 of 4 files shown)


MaoShizhong (changes requested):
Thanks for getting these changes made!
Couple of things:

1. I think you probably took the replacement text from the spreadsheet and replaced only the links in the lesson files with them, as opposed to replacing all of the text from the `Original text` section with the replacement text. This has led to some text duplication which I've highlighted below to show what needs to be removed. This may not have been the clearest on my part in the spreadsheet/instructions.

1. There's also a link I completely missed somehow but GitHub won't let me comment on the line specifically. In `inspecting-html-and-css`, line 57, move the opening `[` forwards so the link text only includes the words "utilize css overview in the developer tools". I have added this link to the spreadsheet for reference purposes.

1. There are a few small simple linting errors to fix. You don't *have* to fix these, and the issue instructions say you don't have to either. But since we need to make changes due to the first point, and there aren't that many to fix, I thought I'd ask if you wouldn't mind fixing these two so we can kill two birds with one stone:
   - Most of the linting errors, I've commented below exactly what needs doing to be fixed.
   - You can ignore the "unordered list indentation" errors. If you add the blank line after the div as instructed below, these errors will disappear automatically.
   - Due to GitHub's annoying rules about what lines I can and can't add comments to, the following lines need a simple replacement:
     - `block-and-inline` line 69 && `inspecting-html-and-css` line 47 - replace each with
     ```
     The following questions are an opportunity to reflect on key topics in this lesson. If you can't answer a question, click on it to review the material, but keep in mind you are not expected to memorize or master this knowledge.
     ```
     - `block-and-inline` line 80 && `inspecting-html-and-css` line 55 - replace each with
     ```
     This section contains helpful links to related content. It isn't required, so consider it supplemental.
     ```

Please do let me know if you need clarification or assistance on these

Crn0:
1. This is my bad, reading the instruction again with fresh mind it is pretty clear now that i am only to change the link text but my mind at that time is kinda exhausted after reading docs so i guess i took the replace the **Original text** as replace all the text not just the link text. I learned a lesson in this experience that i should not push changes if am exhausted and i should ask more about the instructions.
2. okay
3. sure. i do not mind doing this

I have a question regarding commits. should i commit the change separately or is it fine to just do a one commit for the changes?

MaoShizhong:
@Crn0 No worries at all! That's why we have the review process. If you make any mistakes, *usually* we catch them in the review and can request changes like above. And if we don't, well that's our fault for not reviewing thoroughly enough (mistakes happen all the time anyway).

We tend to squash commits when we merge them so the number you push to this branch won't really matter.
In terms of your own practice though, it'd be better to commit related things and separate unrelated things.

E.g. points 1 and 2 are very closely related so you could handle both of those in a single commit. But I'd personally do point 3 in a separate commit (doesn't matter the order, just separate that's all) because the changes are completely unrelated. If I needed to revert something, I'm less likely to revert changes not related to what I wanted to revert if anything unrelated is in a separate commit.

Crn0:
Thank you, Mao, for being understanding. I already pushed the changes, please let me know if i missed anything.

foundations/html_css/css-foundations/inspecting-html-and-css.md
```diff
@@ -54,4 +55,4 @@ This section contains questions for you to check your understanding of this less
 
 This section contains helpful links to related content. It isn’t required, so consider it supplemental.
 
-- [This article about how we can utilize css overview in the developer tools](https://www.freecodecamp.org/news/how-to-use-css-overview-in-chrome-developer-tools/) to check the colors, font styles, media-queries, etc. used on a particular webpage.
+- This article about how we can [utilize css overview in the developer tools](https://www.freecodecamp.org/news/how-to-use-css-overview-in-chrome-developer-tools/) to check the colors, font styles, media-queries, etc. used on a particular webpage.
```
MaoShizhong, line 58:
```suggestion
- This article about how we can [utilize CSS overview in the developer tools](https://www.freecodecamp.org/news/how-to-use-css-overview-in-chrome-developer-tools/) to check the colors, font styles, media-queries, etc. used on a particular webpage.
```
The linter caught this before but didn't show it in the `files changed` tab (only the lint actions log which I didn't check thoroughly...)

MaoShizhong (changes requested):
There are a few more lint errors that were flagged before (and showed in the lint actions log), but for some reason just didn't show in the `Files changed` tab. Should've checked the log myself instead of relying on the `Files changed` flags.

Besides the CSS one I've commented on below, the others are just the "Section default content" ones like in point 3 from my earlier comment above.

Note that the additional resources default section content requires a specific type of apostrophe, as for some reason, some people's systems use a different apostrophe character. You can copy+paste the default section content from the linting rule itself (found in `markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionsContent.js`).

You can run lint checks yourself locally to avoid discovering them via the actions only once you've pushed commits here. 
[Linting instructions](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md#curriculum-linting) from our contributing guide.

If you follow that, you can run lint checks locally so that you only commit and push once the local linter is happy. It'd be *very* rare if the local linter is happy but the GH actions linter catches something else (which would usually indicate we have a bug somewhere), assuming you linted locally correctly.

Crn0:
That's strange. This linting error did not show at first when i am changing the lints, now after changing and pushing it showed in my local files. should i also do this Mao? seems like a easy fix, as it is just adding space right?

MaoShizhong:
Was just about to say this 😅 They only just popped up because that lint rule was one I wrote and was added just a few hours ago.

Yes, the fix is dead simple. It's just making sure those 2 sets of p tags have a single blank line on either side of them.

Sorry to have to make you go back and forth on this!

Crn0:
Hello, Mao! in the the-cascade.md file line 107 is having a linting issues and also line 108 after i fix 107 linting issues, which is saying to add space like the <p> tag, should i also do it?

 

> Sorry to have to make you go back and forth on this!

Don't be, it is a good experience for me and i appreciate you for being patient and kind with me.

MaoShizhong:
No, you can ignore that as it's outside of the 3 files you were assigned to work on.

I just noticed that somehow, `introduction-to-css` made its way in the this PR with fixes despite not being part of the assignment.

In this case, no biggie as it's just the default content fixes and it's not flagging any other errors, and also won't interfere with anyone else's assignments as part of this.

The only changes you need to make now are the surrounding blank lines for the 2 pairs of p tags in `block-and-inline.md`.

Crn0:
Uh. Mao, i think i made a mistake. I accidentally commit, and to fix it i did git restore --stage foundations/html_css/css-foundations/block-and-inline.md, by doing this i thought it will just restore it before i did the commit but instead it restores the file before i did any changes; like the link texts, and section contents is before.

My question, is there a way to get back the changes i did before? or should i just manually change and commit it again?

MaoShizhong:
If you view your `git log`, do you still see a commit with the hash `91cbbf1`? It should have the title "Merge branch into main..." Etc.

From the sounds of it, you should. And since you haven't pushed anything, shouldn't be too bad.

If you do, then you can `git reset --hard 91cbbf1` which will rewind time and reset the branch to as if you just made that commit and nothing after. Then, you can just make the 4 new line changes to `block-and-inline.md` and commit/push that one file to resolve everything.

MaoShizhong (approved):
We're ready to ship! :rocket:
Thanks for working through these!

PR #27674, opened 2024-03-23
Markdownlint: Add custom rule  "Blanks around multiline HTML tags"

MaoShizhong:
## Because
Discovered in #27648, until a blank line is reached, any text following an opening HTML tag gets ignored by the linter almost all of the time. This allows for easy misses due to uncaught rule errors.

The cause of this is [markdown-it](https://github.com/markdown-it/markdown-it)'s method for parsing HTML blocks, where any content following an opening HTML tag, until a blank line is reached, gets bundled into the same single `html_block` token. Thus, they don't get their own tokens and so bypass lint checks.

By enforcing blank lines (or code block delimiters) surrounding isolated HTML tags, the above is avoided and all HTML block text content can be linted appropriately. HTML tags within text (e.g. spans/anchors for link fragments) are ignored by this rule.

## This PR
- Adds a new custom rule TOP005
- Adds a test `.md` file for TOP005 demonstrating valid and invalid cases
- Adds a TOP005 documentation file
- Adds a TOP005 file path to the linter config file's custom rules property
- Disables the linter for the first line of all custom rule doc files, which require a level 1 heading instead of a level 3 heading (as those files will not be rendered on the TOP website)

## Issue
Closes #27673

## Additional Information
Failed lint checks all come from the `TOP005_test.md` file which are all intended errors.

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x] If this PR addresses an open issue, it is linked in the `Issue` section
-   [ ] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [ ] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

PR #27720, opened 2024-03-31
Layout Style Guide: Add new rules

MaoShizhong:
## Because
With the addition of the `TOP005` rule in #27674, the layout style guide needs to be updated to include that rule.

Two other rules were missing from the style guide:
1. Use markdown links instead of HTML anchor tags
2. Use the long name for codeblock languages, for any languages that have a long and short form (e.g. `javascript` and `js`)

## This PR
- Adds the rule covered by `TOP005` to the layout style guide
- Also adds a clarification to use markdown links instead of HTML anchor tags
- Adds clarification for using the long language name for codeblocks if a short option is also acceptable
- Fixed TOP005_test.md to abide by the long language name rule

## Issue
N/A

## Additional Information
The only linting errors that should be present are the expected TOP005 errors from `TOP005_test.md`.

## Pull Request Requirements

### DIFF
--- a/LAYOUT_STYLE_GUIDE.md
+++ b/LAYOUT_STYLE_GUIDE.md
@@ -384,6 +384,7 @@ cd Documents
 If you need a codeblock within a list, you should follow the same indenting rules for [multi-line list items](#multi-line-list-items), with the codeblock being indented with 2 spaces for a bulleted list item and 3 spaces for a numbered list item. The following Markdown:
 
 <pre>
+
 - Bullet.
 
   ```javascript
@@ -394,6 +395,7 @@ If you need a codeblock within a list, you should follow the same indenting rule
   ```
 
 - Next bullet.
+
 </pre>
 
 Will result in the following output:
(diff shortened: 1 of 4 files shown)


LAYOUT_STYLE_GUIDE.md
```diff
@@ -361,9 +362,11 @@ const obj = {
 
 #### Declare the language
 
-It is best practice to explicitly declare the language immediately after the opening tilde marks, so that neither the
+It is best practice to explicitly declare the language immediately after the opening backticks, so that neither the
 syntax highlighter nor the next editor must guess.
 
+If a language has both a long and short form that markdown will accept, for example `javascript` will also be accepted as `js`, and `text` will also be accepted as `txt`, the long form should be used.
```
thatblindgeye, line 368:
```suggestion
If a language has both a long and short form that markdown will accept, for example `javascript` will also be accepted as `js`, and `text` will also be accepted as `txt`, the long form must be used.
```

thatblindgeye (changes requested):
Just one quick change below. Rather than " you *should* do this..." we should (ha) probably say "You *must* do this". Make it clear that we will only accept long form names, especially with a rule being made for it/a possible fixer being done.

thatblindgeye (approved):

PR #27723, opened 2024-03-31
Markdownlint: Add custom rule "Use full language name for fenced code blocks"

### LINKED ISSUE #27721: New Markdownlint rule: Use long language name for codeblocks
### Checks

### Describe your suggestion

#27720 will have the layout style guide specify that any codeblock with languages set must use the long form of the language name if there is both a long and short form that work. For example, using `javascript` instead of `js`, even if `js` will work fine.

There is no built-in rule for this, other than an array of languages to accept, which is much clunkier than an option to exclude specific languages.

Therefore, it would be better to write a custom rule that detects short-form languages for those we might expect to see in our lesson files (rb, js, md, txt) and flags, advising the use of the respective long form.instead (ruby, javascript, markdown, text).

### Path

Other / NA

### Lesson Url

N/A

### (Optional) Discord Name

_No response_

_No response_

MaoShizhong:
## Because
#27720 will have the layout style guide specify that any code block with languages set must use the full language name if there is both a full and abbreviated form that work. For example, using `javascript` instead of `js`, even if `js` will work fine.

There is no built-in rule for this, other than an array of languages to accept, which is much clunkier than an option to exclude specific languages.

Therefore, it would be better to write a custom rule that detects abbreviated languages for those we might expect to see in our lesson files (`rb`, `js`, `md`, `txt`) and flags, advising the use of the respective full name instead (`ruby`, `javascript`, `markdown`, `text`).

Languages like `html` and `css`, which do not have a valid full name option (and would be hell to type anyway), are exempt from this rule.

## This PR
- Adds TOP006 rule
- Adds TOP006_test file, demonstrating how the rule works
- Adds TOP006 doc file
- Adds TOP006 as a custom rule to the markdownlint config file

## Issue
Closes #27721

## Additional Information
N/A

## Pull Request Requirements

### DIFF
--- a/markdownlint/TOP006_fullFencedCodeLanguage/tests/TOP006_test.md
+++ b/markdownlint/TOP006_fullFencedCodeLanguage/tests/TOP006_test.md
@@ -47,8 +47,18 @@ As does txt.
 ```
 
 ```html
-HTML is not considered as only the abbreviated name is a valid option.
-The same applies to similar languages like CSS.
+<p>HTML is not considered as only the abbreviated name is a valid option.</p>
+<p>The same applies to similar languages like CSS and JSX.</p>
+```
+
+```css
+.error {
+    display: none;
+}
+```
+
+```jsx
+{isExempt && <p>No error here!</p>}
 ```
 
 ### Knowledge check
--- a/markdownlint/TOP006_fullFencedCodeLanguage/tests/TOP006_test.md
+++ b/markdownlint/TOP006_fullFencedCodeLanguage/tests/TOP006_test.md
@@ -27,7 +27,7 @@ console.log('This code block is valid as it uses the appropriate full name.');
 ```
 
 ```md
-The rule catches the following languages, as they are they ones that will be seen in this repo's files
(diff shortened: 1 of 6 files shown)


markdownlint/TOP006_fullFencedCodeLanguage/tests/TOP006_test.md
```diff
@@ -0,0 +1,64 @@
...
+
+```txt
+As does txt.
+```
+
+```html
+HTML is not considered as only the abbreviated name is a valid option.
+The same applies to similar languages like CSS.
+```
+
```
thatblindgeye, line 53: Can we just add in a css code block as well?

thatblindgeye (changes requested):

markdownlint/TOP006_fullFencedCodeLanguage/tests/TOP006_test.md
MaoShizhong: Threw in a jsx one too for good measure!

thatblindgeye (approved):

Issue #27867, opened 2024-04-26
Styling React Applications(React): clarifications regarding Tailwind CSS as a styling option


Siriuszx:
### Checks

- [X] This is not a duplicate of an existing issue (please have a look through our [open issues list](https://github.com/TheOdinProject/curriculum/issues) to make sure)
- [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [X] Would you like to work on this issue?

## The problem
Hi there! In the lesson it's clearly stated that when it comes to the *implementation of our component styles* throughout the course we should stick to Vanilla CSS or CSS-in-JS. However, when you scroll up a little bit, you see Tailwind CSS categorized as Vanilla CSS option. So I asked guys in the chat back then and got the response in the spirit of "people doing this course should stick to writing their own CSS and stay away from frameworks as regular CSS is something that all of those frameworks are built upon" **( 1 )**, which I completely agree with.

So we've got the general consensus that students should avoid using CSS frameworks (and to "implement your own component's styling from scratch" **( 2 )**) at least until the end of the TOP and the lesson telling people to stick to Vanilla CSS and listing Tailwind CSS as a valid Vanilla CSS option.

Which basically contradicts the general advice because:
  a) Tailwind isn't vanilla CSS (even though you still end up writing vanilla-ish CSS...) and IS a framework which violates **( 1 )** 
  b) Tailwind provides a whole system which dictates the overall style of the app which violates **( 2 )**
## The solution
I would like to use more specific wording instead of "Vanila CSS" since it's rather vague (and Tailwind **could be** considered as Vanilla CSS in that context) and remove Tailwind CSS documentation reference from "Additional resources" section but leave it as "Vanilla CSS" option in the list to avoid confusion and funnel people into writing regular CSS.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-react-new-styling-react-applications

### (Optional) Discord Name

SiriusMain#6886

## Additional infromation
- [My initial question](https://discord.com/channels/505093832157691914/505093832157691916/1192217501275271218)
- [Overall response that I got](https://discord.com/channels/505093832157691914/505093832157691916/1192224217677189120)

wise-king-sullyman:
Sounds good, thanks!

PR #27874, opened 2024-04-27
Styling React Applications: Update styling recommendations

Siriuszx:
## Because

The lesson tells people that Tailwind CSS is a valid Vanilla CSS option which isn't completely untrue, however the general consensus is that students should refrain from utilizing CSS frameworks or component libraries and try writing their *own* css.

## This PR

- reworded lesson tip
- removed Tailwind CSS docs link from Additional Resources
- added navigation links that point to the list of styling options available for students to choose from

## Issue

Closes [#27867](https://github.com/TheOdinProject/curriculum/issues/27867)

## Pull Request Requirements

### DIFF
--- a/react/the_react_ecosystem/styling_react_applications.md
+++ b/react/the_react_ecosystem/styling_react_applications.md
@@ -1,6 +1,6 @@
 ### Introduction
 
-Up until now you've been using vanilla CSS to style your React projects, but there are other options available. This lesson is intended to be more like a guide or a list of options for you to explore.  
+In the previous courses, you've learned a lot of CSS and all of those skills are still applicable to React, however there are a couple of things we'd like to highlight. As you've probably already noticed, all of the styles we write share the global scope, which means that as our application grows, it becomes increasingly difficult to manage our CSS. Some of the tools mentioned below will help solve this problem, however it's important to note that the main purpose of this lesson is to be more of a guide or a list of available options for you to explore when it comes to styling React applications.
 
 ### Lesson overview
 
@@ -8,18 +8,18 @@ This section contains a general overview of topics that you will learn in this l
 
 - Know about ways to style React applications
 
-### Vanilla CSS
-
-Vanilla CSS is the simplest way to style. In the previous courses, you've learned a lot of CSS and all of those skills are applicable to React. There are a couple of things we'd like to highlight.
-
-1. `CSS Modules` let you write CSS style declarations that are scoped locally.
-1. `CSS Utility Frameworks` are a popular choice for styling React applications. They provide a set of pre-defined classes that you can directly use in your HTML, or JSX in our case. [Tailwind CSS](https://tailwindcss.com) is by far the most popular choice.
+### CSS Modules
+Regular CSS is the simplest way to style. CSS Modules let you write CSS style declarations that are scoped locally, which means that we finally no longer have to worry about our class names potentially conflicting with other classes in the global scope, i.e. we can now have two (or more) classes with the same name but different styles, and avoid naming your classes `.list-wrapper`, `.item-wrapper` etc. and just use `.wrapper` with its respective component.
 
 ### CSS in JS
 
 Why even write CSS in CSS when you can write it in JavaScript? Just kidding, of course!
 
-CSS-in-JS is a paradigm for styling front-end projects. It allows you to entirely take control of CSS with JavaScript and extends it with various features. Additionally, it also helps to apply styling in a logical fashion i.e. based on state. There are various CSS-in-JS solutions. One of the most popular ones in the React ecosystem is [styled-components](https://styled-components.com/).
+CSS-in-JS is a paradigm for styling front-end projects. It allows you to entirely take control of CSS with JavaScript and extends it with various features. Additionally, it also helps to apply styling in a logical fashion i.e. based on state, and also supports modular CSS in the same way that CSS Modules do. There are various CSS-in-JS solutions. One of the most popular ones in the React ecosystem is [styled-components](https://styled-components.com/).
+
+### CSS Utility Frameworks
+
+CSS Utility Frameworks are a popular choice for styling React applications. They provide a set of pre-defined classes that you can directly use in your HTML, or JSX in our case. [Tailwind CSS](https://tailwindcss.com) is by far the most popular choice.
 
 ### Component libraries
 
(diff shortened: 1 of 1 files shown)


react/the_react_ecosystem/styling_react_applications.md
```diff
@@ -30,7 +30,7 @@ What if everything's already done for you? Styling, behavior, and accessibility
 
 <div class="lesson-note lesson-note--tip" markdown="1" >
 
-For learning purposes throughout this course, we advise you to implement your component's styling from scratch i.e. use vanilla CSS or a CSS-in-JS option. 
+For the learning purposes throughout this course, we recommend that you avoid using CSS frameworks or component libraries, and instead implement your component's styling from scratch i.e. use [CSS Modules](#vanilla-css) or a [CSS-in-JS](#css-in-js) option.
```
MaoShizhong, line 33:
```suggestion
For learning purposes throughout this course, we recommend that you avoid using CSS frameworks or component libraries, and instead implement your component's styling from scratch i.e. use [CSS Modules](#vanilla-css) or a [CSS-in-JS](#css-in-js) option. 
```

Nit: "For learning purposes" (omitting "the") is more appropriate here.

MaoShizhong (changes requested):
Thanks for doing this PR. Would you mind fixing up the lint issues as well? They should be pretty self explanatory. Let me know if you run into any issues with this.

Siriuszx:
Sorry for the force push, I thought it would overwrite my last commit and the message with typo if I use `git commit --amend`

MaoShizhong (commented):
In all honesty, the more I think about it, the more unsure I am about having Tailwind within `Vanilla CSS`, because it isn't vanilla CSS by any means. CSS modules can be argued as "for all intents and purposes, vanilla CSS" since it's almost the same, just with very tiny differences like local scoping. 

I feel Tailwind ought to go in its own `### Utility frameworks` between `### CSS in JS` and `### Component libraries`.
- The Vanilla CSS section would need a very slight rewording as it's no long highlighting "a couple of things", and it won't need an ordered list anymore. CSS modules can just be a normal sentence/paragraph in it.
- The Tailwind list item can be made its own paragraph in its own `### Utility frameworks` section.

What do you think, @Siriuszx?

Siriuszx:
I think the definition of "Vanilla CSS" is too vague and **IS** purely subjective in that context.

To avoid confusion in the future, I think it would be better if we just get rid of the term "Vanilla CSS" altogether and split the lesson into:
- CSS Modules
- CSS Utility Frameworks
- CSS-in-JS
- Component Libraries

Siriuszx:
If we are on the same page about this could you please merge this PR into main and I will open a separate issue addressing this lesson's structure? It would be amazing if I get the chance to work on the next PR as well 🙂

MaoShizhong:
@Siriuszx I think this is pretty directly related to the original issue and PR purpose, so I think it can be done in this PR instead of a separate one.

I like your idea about the sections so let's go with that.
I think the Introduction section could do with a slight update to include something about vanilla CSS potentially getting harder to manage in a larger project due to every file sharing a global scope. Therefore there are some popular alternatives in common use.

I think that'll help set the scene for why something like css modules is so useful because of the local scoping, while being otherwise almost vanilla css.

I'll leave that with you, and we can always refine in review if we need to.

Siriuszx:
@MaoShizhong sorry if this is taking too long, it says that you've "self-assigned this". Should I still work on this issue or not?

MaoShizhong:
Oh yeah, carry on with it @Siriuszx. The PR assignment is a maintainer thing, you can ignore it

react/the_react_ecosystem/styling_react_applications.md
```diff
@@ -1,36 +1,36 @@
...
 This section contains a general overview of topics that you will learn in this lesson.
 
-- Know about ways to style React applications 
+- Know about ways to style React applications
 
-### Vanilla CSS 
+### CSS Modules
 
-Vanilla CSS is the simplest way to style. In the previous courses, you've learned a lot of CSS and all of those skills are applicable to React. There are a couple of things we'd like to highlight.
+Regular CSS is the simplest way to style. CSS Modules let you write CSS style declarations that are scoped locally, which means that we finally no longer have to worry about our class names potentially conflicting with other classes in the global scope, i.e. we can now have two (or more) classes with the same name but different styles, and avoid naming your classes `.list-wrapper`, `.item-wrapper` etc. and just use `.wrapper` with its respective component.
```
MaoShizhong, line 13:
```suggestion
Regular CSS is the simplest way to style. CSS Modules let you write CSS style declarations that are scoped locally, which means that we finally no longer have to worry about our class names potentially conflicting with other classes in the global scope. For example, we can have several CSS module files containing a `.wrapper` class, but each module's `.wrapper` will be treated as different to each other.
```
The above is purely a suggestion, so you don't have to take it at all. Just mainly feeling like it'd be best to just briefly explain how a class in one CSS module is different to the same class name in another CSS module.
I'm not sure the current example is 100% applicable (I'm being very nitpicky here!), since you may still want to use class names like `list-wrapper` and `item-wrapper` between/within CSS modules for various reasons.

MaoShizhong (changes requested):
Love this! Just a few nits/suggestions below.

react/the_react_ecosystem/styling_react_applications.md
Siriuszx: You are right, I guess it could be even redundant. People will find explanations for what it really is when they eventually get to the assignment.

MaoShizhong: Agreed, especially when the documentation for CSS modules is already linked as an assigned reading.
We could probably just leave it at something like
> Regular CSS is the simplest way to style. CSS Modules let you write locally scoped CSS style declarations, preventing many issues with clashing class names in the global scope.

Siriuszx: Already done! :)

> Agreed, especially when the documentation for CSS modules is already linked as an assigned reading. We could probably just leave it at something like
> 
> > Regular CSS is the simplest way to style. CSS Modules let you write locally scoped CSS style declarations, preventing many issues with clashing class names in the global scope.

Siriuszx:
It seems there is something wrong with my pr. Please don't merge @MaoShizhong, I will fix it.

MaoShizhong (approved):
:rocket:

PR #27889, opened 2024-04-29
Intro to Flexbox lesson: Remove rogue </span> text

SerbanUntu:
The </span> tag after the first image in the lesson was separated by two newlines, causing it to be displayed as a paragraph of text in the browser, so I removed the two newlines.

## Because
The space between the link to the image and the </span> tag afterwards was too large, it caused the tag to be interpreted as a separate paragraph, leading to the visual bug.

This is a screenshot of the part of the lesson in which the visual bug occurs:
Screenshot of bug
## This PR
- Removes the two newlines before the </span> tag after the link to the first image in the lesson to remove the `</span>` text from the page.

## Pull Request Requirements

### DIFF
--- a/foundations/html_css/flexbox/flexbox-intro.md
+++ b/foundations/html_css/flexbox/flexbox-intro.md
@@ -64,9 +64,7 @@ As you've seen, flexbox is not just a single CSS property but a whole toolbox of
 
 <span id="flex-container-item-knowledge-check">A flex container is any element that has `display: flex` on it. A flex item is any element that lives directly inside of a flex container.</span>
 
-<span id="how-to-create-flex-item-knowledge-check">![container-vs-child](https://cdn.statically.io/gh/TheOdinProject/curriculum/b2a53579fcbec1cfde47646cc5a2b109cd7772cc/foundations/html_css/flexbox/imgs/03.png)
-
-</span>
+<span id="how-to-create-flex-item-knowledge-check">![container-vs-child](https://cdn.statically.io/gh/TheOdinProject/curriculum/b2a53579fcbec1cfde47646cc5a2b109cd7772cc/foundations/html_css/flexbox/imgs/03.png)</span>
 
 Somewhat confusingly, any element can be both a flex container *and* a flex item. Said another way, you can also put `display: flex` on a flex item and then use flexbox to arrange *its* children.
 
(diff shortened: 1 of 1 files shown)


foundations/html_css/flexbox/flexbox-intro.md
```diff
@@ -64,9 +64,7 @@ As you've seen, flexbox is not just a single CSS property but a whole toolbox of
 
 <span id="flex-container-item-knowledge-check">A flex container is any element that has `display: flex` on it. A flex item is any element that lives directly inside of a flex container.</span>
 
-<span id="how-to-create-flex-item-knowledge-check">![container-vs-child](https://cdn.statically.io/gh/TheOdinProject/curriculum/b2a53579fcbec1cfde47646cc5a2b109cd7772cc/foundations/html_css/flexbox/imgs/03.png)
-
-</span>
+<span id="how-to-create-flex-item-knowledge-check">![container-vs-child](https://cdn.statically.io/gh/TheOdinProject/curriculum/b2a53579fcbec1cfde47646cc5a2b109cd7772cc/foundations/html_css/flexbox/imgs/03.png)</span>
```
MaoShizhong, line 67: Good spot, thanks for bringing this up. Since this is an image and the span is purely to have an element with an id for a knowledge check link to scroll to, it would actually be better for us to attach the id to the image itself, which can be done in markdown as follows.

```suggestion
![container-vs-child](https://cdn.statically.io/gh/TheOdinProject/curriculum/b2a53579fcbec1cfde47646cc5a2b109cd7772cc/foundations/html_css/flexbox/imgs/03.png){: #how-to-create-flex-item-knowledge-check}
```

A note for future PRs, you're currently PRing from your fork's `main` branch. While this won't cause problems for this PR, it can cause a lot of headaches if you start working on multiple things, with unrelated commits bleeding into different work. You will want to keep your main branch untouched, other than keeping it in sync with upstream (TOP's repo). All your work should go in a different branch coming off `main` for each individual PR, meaning each PR will have its own commit history, untouched by any other branches' work (which is what we want).

You can look through the [contributing guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md) for instructions on the workflow.

MaoShizhong (changes requested):

foundations/html_css/flexbox/flexbox-intro.md
SerbanUntu: Thank you for the suggestions! I didn't know about the markdown syntax in your change, I was mainly concerned with removing the unnecessary text, but it's a nice improvement. I guess there's many more features in markdown I don't know about.

MaoShizhong (approved):

Issue #27819, opened 2024-04-19
Node Pathway: Provide descriptive link text


MaoShizhong:
Links in our lessons should be sufficiently descriptive to improve accessibility. As per the [layout style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#links), link text should both concisely describe its purpose (or where the link will redirect a user), and must not include the words `this` nor `here`.

## Guidance

- Comment on this issue to let us know you'd like to get a section assigned to you. You can also ask for a specific one.
- Please wait until you are assigned by a maintainer before starting to work on this issue to avoid multiple people working on the same lessons.
- When you have been assigned a section, please refer to the [suggested link text changes spreadsheet](https://docs.google.com/spreadsheets/d/1E1OJqUfBODvjdFMNgFMY_GvOLrOzV4ujL9o7Nxzg2zk/edit?usp=sharing) for instructions on what changes need to be made for your assigned lesson(s).
  - For most of these, you should just be able to copy and paste the suggested text to replace the original text shown in the spreadsheet (which may be more than just the link itself). Any additional steps/clarifications will be noted by the appropriate link(s).
  - You are welcome to use a different wording from a given suggestion if you find a better one. If you do, you must ensure the link text is consistent if there are multiple instances of the same href in the same lesson.
  - You are *not* required to fix any unrelated, preexisting linting errors such as list styles/indentation. Existing linting errors may cause certain tests to fail when you open a PR. You can ignore this and a maintainer will let you know if further action is required.
- Make sure to read the [general contributing guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md) as well as the [curriculum repo contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md) before starting your first contribution.
- Please add the following to your PR in the Issues section: `Related to #XXXXX` , since this issue is divided into sections, **individual PRs should NOT INCLUDE the `closes` keyword.**

### Intermediate HTML and CSS

- [x] @Crn0 [Emmet](https://www.theodinproject.com/lessons/node-path-intermediate-html-and-css-emmet) && [SVG](https://www.theodinproject.com/lessons/node-path-intermediate-html-and-css-svg)
[...]

PR #27896, opened 2024-04-29
NodeJS: Add descriptive link text

guskirb:
Made changes to the links in the Express 102, 103, 104 and 105 lessons.

## Because

Provide descriptive link text.

## This PR

* Add descriptive link text to Express 102 lesson.
* Add descriptive link text to Express 103 lesson.
* Add descriptive link text to Express 104 lesson.
* Add descriptive link text to Express 105 lesson.

## Issue

Related to #27819

## Pull Request Requirements

### DIFF
--- a/nodeJS/express_and_mongoose/express_102_crud_and_mvc.md
+++ b/nodeJS/express_and_mongoose/express_102_crud_and_mvc.md
@@ -19,7 +19,7 @@ This section contains a general overview of topics that you will learn in this l
 
 CRUD is a concept that comes up a lot in web development, and it's the type of thing that might show up in interview questions so it's worth taking a little time to make sure you understand what it refers to.
 
-CRUD stands for: <span id="crud">_Create, Read, Update_ and _Delete_</span>. These are the four basic functions that you will be building into your database driven apps. If you are designing a CRUD interface that means that users can expect to be able to do these 4 things to items in the database (providing they have the appropriate permissions of course).
+CRUD stands for: <span id="crud">*Create, Read, Update* and *Delete*</span>. These are the four basic functions that you will be building into your database driven apps. If you are designing a CRUD interface that means that users can expect to be able to do these 4 things to items in the database (providing they have the appropriate permissions of course).
 
 In your library example, this means that we are going to be building the ability for users to `create` entries (add books, authors or genres to the database), `read` entries (or, retrieve lists of books and other things from the database), `update` entries (edit details of an entry), and `delete` entries (remove them from the database).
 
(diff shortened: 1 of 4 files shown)


MaoShizhong (changes requested):
Thanks for sorting out these links, @guskirb!

Would you be good to fix up the remaining lint errors? If you follow the [curriculum repo contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md), it'll show you how to run lint checks and fixes locally.

Most of the lint errors should be fixable by running the correct script on a file (`project_express_105` will need the project scripts, not the lesson scripts). There will be a couple that can't be auto-fixed so will need manually fixing (the error message should be somewhat self explanatory).

For the "Required heading structure" error in `express_103_crud_and_mvc.md` line 62, that should be fixed by removing the `s` from `Knowledge checks` to make `Knowledge check`.

You can ignore the "Required heading structure" error for `express_104_view_templates.md`.

Let me know if you run into any issues with this.

guskirb:
> Small side note - I suspect you may have a formatter like Prettier that runs on save, as the `project 105` file had its `* *` emphasis (which is what we want) switched to `_ _` (which is not what we want). I'd recommend finding the keybinding for `Save without formatting` and getting used to using that when working on any external projects, as you can't guarantee they will use the same style guide as the formatter you have running automatically. Alternatively, you can have it so your formatter does not run on save but instead is bound to a separate keybinding. Then you just only use that when working on projects you know are fine to use your formatter with.

I've fixed the lint errors.

 I did have prettier set to reformat on save, which I've changed, thanks for letting me know.

nodeJS/express_and_mongoose/express_102_crud_and_mvc.md
```diff
@@ -19,58 +19,58 @@ This section contains a general overview of topics that you will learn in this l
 
 CRUD is a concept that comes up a lot in web development, and it's the type of thing that might show up in interview questions so it's worth taking a little time to make sure you understand what it refers to.
 
-CRUD stands for: <span id="crud">_Create, Read, Update_ and _Delete_</span>. These are the four basic functions that you will be building into your database driven apps. If you are designing a CRUD interface that means that users can expect to be able to do these 4 things to items in the database (providing they have the appropriate permissions of course).
+CRUD stands for: <span id="crud">*Create, Read, Update* and *Delete*</span>. These are the four basic functions that you will be building into your database driven apps. If you are designing a CRUD interface that means that users can expect to be able to do these 4 things to items in the database (providing they have the appropriate permissions of course).
```
MaoShizhong, line 22:
```suggestion
CRUD stands for: *Create, Read, Update* and *Delete*. These are the four basic functions that you will be building into your database driven apps. If you are designing a CRUD interface that means that users can expect to be able to do these 4 things to items in the database (providing they have the appropriate permissions of course).
```
Not something you've made but I just noticed that we have a redundant span here, as the `### CRUD` heading itself directly above already generates an `h3#crud`, so the span isn't actually needed for scroll either (since it's not like it's a long way away from the heading). Mind removing it while we're at it?

Everything else looks :man_cook: :pinched_fingers:

MaoShizhong (changes requested):

nodeJS/express_and_mongoose/express_102_crud_and_mvc.md
guskirb: Done!

MaoShizhong (approved):
:rocket:

PR #27953, opened 2024-05-09
ES6 Modules: Rewrite lesson (alongside separate PR for Webpack rewrite)

MaoShizhong:
## Because

The current [Webpack lesson](https://www.theodinproject.com/lessons/node-path-javascript-webpack) is a known pain point in the curriculum journey, primarily due to the sudden increase in complexity and steps of tooling, coupled with how much of the learning is left to external resources. These external resources limit the cohesiveness of teaching material, and can prove troublesome in the learning process when each resource includes a few unnecessary tidbits that can easily throw learners off (such as multiple entry points in the webpack official guides when learning about html-webpack-plugin). They also are frequently aimed at people who may know a little more about bundlers already, or are following a different course/tutorial. Thus, things often feel disjointed with the TOP curriculum.

While improving the Webpack lesson itself is the primary focus, it was also identified that the preceding [ES6 modules lesson](https://www.theodinproject.com/lessons/node-path-javascript-es6-modules) is fairly coupled with it, as Webpack and bundlers are introduced there instead of the Webpack lesson.

The order of the ES6 modules lesson was also identified as a little odd, and the order and flow could be improved by focusing on ES6 modules as a primary topic straight away, then introducing bundlers and Webpack in their own separate lesson.

General plan:
- Rewrite ES6 modules lesson
- Rewrite Webpack lesson
- Write new "Webpack revisited" lesson to go immediately after "Project: Restaurant Page".
  - Move more advanced but "convenience" topics there, like webpack-merge
  - Extract template repos section from "Linting" and place it here instead
- Tweak Project: Restaurant Page to reference any new material from preceding lesson changes if necessary

## This PR

- Rewrites the ES6 modules lesson
- Adds a new lesson on npm

## Issue

Closes #26976
Closes #26977

## Additional Information

This will be something to merge alongside 3 other PRs:
1. (#27961)
2. (#27962)
3. (https://github.com/TheOdinProject/theodinproject/pull/4522)

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project curriculum contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x] If this PR addresses an open issue, it is linked in the `Issue` section
[...]

Issue #26977, opened 2023-12-23
Webpack: Move webpack merge/modes mention to a later lesson


MaoShizhong:
### Checks

- [X] This is not a duplicate of an existing issue (please have a look through our [open issues list](https://github.com/TheOdinProject/curriculum/issues) to make sure)
- [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [X] Would you like to work on this issue?

### Describe your suggestion

In #26381 and #26552 I ended up adding step 4 of the webpack lesson assignment which took the user through the difference between development/production mode and using `webpack-merge` to set a dev and prod config tied to their own npm scripts.

Whilst this has proved convenient for some people (especially after more experience with Webpack) and has reduced the number of questions concerning the warning when no mode is specified, in hindsight, I feel it is not necessary to include that part in the Webpack lesson as it's not strictly necessary to get started with it. I feel it would be more beneficial moved to a later lesson e.g. the linting lesson (though still not 100% sure if this would the best place either?) in order to give the learner more time to use and get a feel for webpack before then introducing convenience measures.

This came about starting with [this Discord discussion](https://discord.com/channels/505093832157691914/505093832157691916/1174867129040912454) and keeping an eye on things since then, whilst fewer Qs about webpack mode came up, there seemed to be more confusing about webpack as a whole than needs be as people were already struggling to get to grips with the more important parts of using webpack let alone dealing with the difference between mode configs (if anything, the less unnecessary config they need to do at the start, the simpler the learning curve would be I feel). `Html-webpack-plugin` and css/style loaders I feel are the primary ones needed for straightforward workflow and understanding the roles of src/dist and should be plenty to get through restaurant page and to-do list. It just feels currently that webpack-merge and modes will serve only to overwhelm at the point they're mentioned.

Thoughts?

### Path

Ruby / Rails, Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-javascript-webpack

### (Optional) Discord Name

Mao

rlmoser99:
Thank you for taking the time to make an issue. Can someone from the @TheOdinProject/javascript team, please review this issue, as well as a closely related one:
https://github.com/TheOdinProject/curriculum/issues/26976

JoshDevHub:
Hey @MaoShizhong 

[...]

PR #27958, opened 2024-05-10
Webpack: Rewrite and split to flank Restaurant Page project

MaoShizhong:
## Because

The current [Webpack lesson](https://www.theodinproject.com/lessons/node-path-javascript-webpack) is a known pain point in the curriculum journey, primarily due to the sudden increase in complexity and steps of tooling, coupled with how much of the learning is left to external resources. These external resources limit the cohesiveness of teaching material, and can prove troublesome in the learning process when each resource includes a few unnecessary tidbits that can easily throw learners off (such as multiple entry points in the webpack official guides when learning about html-webpack-plugin). They also are frequently aimed at people who may know a little more about bundlers already, or are following a different course/tutorial. Thus, things often feel disjointed with the TOP curriculum.

While improving the Webpack lesson itself is the primary focus, it was also identified that the preceding [ES6 modules lesson](https://www.theodinproject.com/lessons/node-path-javascript-es6-modules) is fairly coupled with it, as Webpack and bundlers are introduced there instead of the Webpack lesson.

The order of the ES6 modules lesson was also identified as a little odd, and the order and flow could be improved by focusing on ES6 modules as a primary topic straight away, then introducing bundlers and Webpack in their own separate lesson.

General plan:
- Rewrite ES6 modules lesson
- Rewrite Webpack lesson
- Write new "Webpack revisited" lesson to go immediately after "Project: Restaurant Page".
  - Move more advanced but "convenience" topics there, like webpack-merge
  - Extract template repos section from "Linting" and place it here instead
- Tweak Project: Restaurant Page to reference any new material from preceding lesson changes if necessary

## This PR

- Rewrites the Webpack lesson ([latest live preview](https://www.theodinproject.com/lessons/preview?uuid=f7f80cbd-f79a-41c5-a30d-aa538c9bcdfc))
- Adds a new Webpack lesson after Restaurant Page ([latest live preview](https://www.theodinproject.com/lessons/preview?uuid=ed2c698f-50b8-4ca4-a9f9-1513622f09b7))
- Removes the Template Repo section (and all other mentions of them) from the Linting lesson, as the contents have been moved to the new second Webpack lesson. This will ironically trigger the Linter because the Linting lesson is missing an Assignment section, which I feel should be addressed separately.

## Issue

Related to #26976
Related to #26977

## Additional Information

Only to be merged alongside #27953 (which will close the two related issues), and when a PR has been completed for the Web app to link the new post-Restaurant Page lesson.

The lin

## Pull Request Requirements


MaoShizhong:
Things I would like thoughts on in particular:

- Topics covered in each Webpack lesson
- Order covered both as a whole but also in each lesson individually
  - I've been debating exactly how to cover npm scripts, whether they'd be better at the end of the first lesson or at the start of the second, because the Webpack guides use them throughout. My justification so far is that the first lesson is mainly a custom in-house rewriting of the Webpack tutorial anyway, so the current note about `npm run build` in the first lesson is sufficient, and npm scripts seem to be something we can leave for later to reduce overload.
- Quality and relevance of all aspects of the custom Webpack tutorial
  - The goal was to provide a similar tutorial that only includes the things that would be most relevant in our curriculum, as the Webpack one includes little bits like multiple entry points, runtime optimisations, devServer.static etc. that make sense for a general "here's how different parts of Webpack work", but less in the context of our curriculum.
  - An in-house tutorial would also allow us to explain the *why* behind certain configurations, as they'll often not explained well (if at all) in the Webpack guides. Such explanations often require some preexisting knowledge of bundlers and build tools.
  - Both of the above would allow us to steer learners more towards the happy path of setting up and using Webpack (and at least kinda getting it), while avoiding some of the main (and very very common) pitfalls we've noticed in the community. Oftentimes arising due to using things they don't need but don't realise, because they just copied the Webpack guide without reading side notes etc.

Any other aspects as well. The above is just what stuck around my mind most during writing.
@advait0603 tagging you here as well for whenever you get any time to review. Again, much of this was going off things you raised in the past and planned out.

MaoShizhong:
Closing to restructure branch

PR #27973, opened 2024-05-15
Project: Restaurant Page: Added section on .gitignore templates

Etharialle:
## Because

The Restuarant Page lesson introduces the `.gitignore` file, but does not mention anything about .gitignore templates which are part of the GitHub web interface when creating a repo.  Specifically for JavaScript the common template is actually named `node` which may not be intuitive to students.

## This PR

- Added section `.gitignore templates` and brief description.

## Issue

N/A

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/project_restaurant_page.md
+++ b/javascript/organizing_your_javascript_code/project_restaurant_page.md
@@ -12,8 +12,7 @@ When working with packages that are installed with npm, you don't need to track 
 
 You can make a `.gitignore` file in the root of the project, and by writing file or directory names in it, you can tell git what things you don't want to track. It's customary to add `node_modules` to `.gitignore`, since it can get really big.
 
-#### .gitignore templates
-When creating a new repo on GitHub there is an option to specify a `.gitnore` template.  There are many templates out there that include common files and directories that are not typically tracked based on the type of project or language used.  When looking for a template for JavaScript projects there is a `node` template that includes `node_modules`.
+When creating a new repo on GitHub there is an option to specify a `.gitignore` template.  There are many templates out there that include common files and directories that are not typically tracked based on the type of project or language used.  When looking for a template for JavaScript projects there is a `node` template that includes `node_modules`.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


CouchofTomato (approved):

CouchofTomato:
There is a linting issue that needs to be fixed before this can be merged. A heading needs a blank line above and below. Please amend.

MaoShizhong:
Do you think we *need* to include this in the project content?
All we need is to have `node_modules` and `dist` in a `.gitignore`, which requires introducing what `.gitignore` is anyway, and the `node` template includes a lot more than that. Won't actually cause any problems since it won't ignore stuff we don't want ignored, and it includes both `node_modules` and `dist`. I'm just wondering if it's actually significantly valuable to include beyond what's already introduced.

Side note, there's a typo `.gitnore`

MaoShizhong:
That's fair enough @CouchofTomato @Etharialle 
What do you both think about possible just adding the note on those templates to the end of the preceding `.gitignore` section? As opposed to making a whole new subsection itself. To me it feels better as just an extension of the previous part, like a "When you create a new repo, there's an option to include a gitignore template, where the node template includes `node_modules` and `dist` already."

Also because I just noticed in that previous part, on top of `node_modules` being ignored for getting big, `dist` is also commonly ignored because it's not needed (can be generated by any user by bundling). Just found it odd we only mention `node_modules` there then in the project steps we say both `node_modules` and `dist`.

MaoShizhong:
> I'll make the linting changes and typo correction shortly. I didn't want to go too far down the rabbit hole on gitignore templates, but since it's something students see every time they create a repo I thought it would be worth mentioning (especially since the JavaScript related template is called node instead of JavaScript).
> 
> That said it's not crucial that it is here and if there is a better place or if it is felt that it is unnecessary this PR can be closed.

Nah, no need to close the PR. Just offering an opinion above for discussion, but it's not a blocking opinion

CouchofTomato:
Hey @Etharialle 

Do you think it is worth making the point that for smaller projects it's unlikely to be useful to use a template vs just adding the directories/files to your own gitignore? We don't want people to go crazy using a template when they'd only gitignore node_modules

Etharialle:
I'm not sure.  I typically use a template when creating a project so I don't have to remember to manually create a .gitignore file.  I don't think there is much concern about having extra lines in the .gitignore.

My main concern would be with projects that include a .env file with secrets in it that are accidentally tracked because a a .gitignore wasn't created or it was manually created, but .env wasn't added.

CouchofTomato (approved):

CouchofTomato:
Fair enough. Thanks for your contribution.

PR #28003, opened 2024-05-19
Markdownlint - Added fixers for TOP003 rule

thatblindgeye:
## Because

Fixers will make it easier for updates to be made when applicable.

## This PR

1. Adds fixers to TOP003 rule

## Issue

Closes #XXXXX

## Additional Information

Would be worth refactoring this rule further, but this PR at least adds some fixers in the interim.

## Pull Request Requirements


markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionContent.js
```diff
@@ -27,92 +28,144 @@ function getListSectionErrors(sectionTokens, section) {
...
       )
     );
   }
+  if (defaultContentOpenTokenIndex === -1) {
+    const errorDetail = `Expected: "${listSectionsDefaultContent[section]}"; Actual: "${tokensAfterHeading[0].line}",`;
+
+    listSectionErrors.push(
+      createErrorObject(tokensAfterHeading[0].lineNumber, errorDetail, {
+        lineNumber: tokensAfterHeading[0].lineNumber,
+        insertText: `${listSectionsDefaultContent[section]}\n\n`,
```
MaoShizhong, line 89: Currently, the correct default content is inserted but the incorrect content is left there, meaning the user would still have to manually delete it.

With the below, I've tried to think of edge cases that could break *and* cause huge problems because of this but I can't think of any. As far as I'm aware, the error with this fixInfo would only throw when the first item isn't the default content *and* the correct default content isn't present in the section.
(Also why the template string for insertText? :P)

```js
  lineNumber: tokensAfterHeading[0].lineNumber,
  deleteCount: tokensAfterHeading[0].line.length,
  insertText: listSectionsDefaultContent[section],
```

The only situation where this could cause issues would be if the very first item *was* the unordered list, in which case the above would just replace that first UL item.

A hacky solution could be to check if the offending first item starts with `- ` then insert `` `${listSectionsDefaultContent[section]}\n\n${tokensAfterHeading[0].line}`  ``
e.g.

```js
  if (defaultContentOpenTokenIndex === -1) {
    const errorDetail = `Expected: "${listSectionsDefaultContent[section]}"; Actual: "${tokensAfterHeading[0].line}",`;
    let insertedText = listSectionsDefaultContent[section];

    if (tokensAfterHeading[0].line.startsWith("- ")) {
        insertedText += `\n\n${tokensAfterHeading[0].line}`;
    }

    listSectionErrors.push(
      createErrorObject(tokensAfterHeading[0].lineNumber, errorDetail, {
        lineNumber: tokensAfterHeading[0].lineNumber,
        deleteCount: tokensAfterHeading[0].line.length,
        insertText: `${listSectionsDefaultContent[section]}`,
      })
    );
  }
```

The above will solve that edge case and doesn't seem to break any others from my testing.

markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionContent.js
```diff
@@ -27,92 +28,144 @@ function getListSectionErrors(sectionTokens, section) {
...
+  ) {
     listSectionErrors.push(
       createErrorObject(
-        (tokensAfterFirstContent[0] || tokensAfterHeading[0]).lineNumber,
-        errorDetail
+        tokensAfterFirstContent[0].lineNumber,
+        `Only an unordered list of ${listItemsName} can follow the default content.`,
+        {
+          lineNumber: tokensAfterFirstContent[0].lineNumber,
+          deleteCount: tokensAfterFirstContent[0].line.length,
```
MaoShizhong, line 143: `-1` deletes the line itself (fine when no text to be inserted).
Not a blocker since either way, it might leave blank lines that require a second fix pass, unless you want some really hacky conditionals which probably won't be worth it.

I personally prefer `-1` for this because we can define a symbolic constant at the top `const WHOLE_LINE = -1;` then do `deleteCount: WHOLE_LINE,`

```suggestion
          deleteCount: -1,
```

markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionContent.js
```diff
@@ -27,92 +28,144 @@ function getListSectionErrors(sectionTokens, section) {
...
 
     listSectionErrors.push(
       createErrorObject(
-        tokensAfterBulletListClose.find((token) => token.lineNumber).lineNumber,
-        `There should be no additional content after the unordered list of ${listItemsName}`
+        tokensAfterBulletListClose[0].lineNumber,
+        `There should be no additional content after the unordered list of ${listItemsName}`,
+        {
+          lineNumber: tokensAfterBulletListClose[0].lineNumber,
+          deleteCount: tokensAfterBulletListClose[0].line.length,
```
MaoShizhong, line 167: Exact same thing with this and `-1`/`WHOLE_LINE`

MaoShizhong (changes requested):
From playing around, the below was all I ran into. Everything else *seems* to be in working order and justified.

markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionContent.js
thatblindgeye: Yeah I wasn't sure whether we should remove content and then insert content for this one, just because we may not know what exactly is there to possibly remove. I opted to just insert the default content, then any subsequent runs of the fixer should catch/fix the "additional content between default and list" portion.

If you feel that there shouldn't be any issues with something like the above I can add that change, though, let me know what you think.

MaoShizhong: I think it should be fine.

If it goes
```md
### Heading

Unwanted stuff (inc. ul)

Correct default text
```
then the `Expected default section content to come immediately after the ${section} heading.` error gets flagged instead.

If the correct default text is not there, then the only instance of the first line being something we want would be if the UL is the first item, in which case the 2nd snippet I provided will cover that (for LO/KC/AR sections).

And any other edge case I can think of would result in bigger problems anyway and surely would be caught in review if not by the author (and likely would leave post-fix errors anyway).

Doing the delete + insert would make most realistic situations handled in 1 pass, perhaps 2 just for any extra blank lines left behind that we can't address.

markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionContent.js
```diff
@@ -27,92 +28,153 @@ function getListSectionErrors(sectionTokens, section) {
...
-        `Expected default section content to precede the unordered list of ${listItemsName}: "${listSectionsDefaultContent[section]}"`
+        defaultContentToken.lineNumber,
+        `Expected default section content to come immediately after the ${section} heading.`
       )
     );
   }
+  if (defaultContentOpenTokenIndex === -1) {
+    const sectionStartsWithList = tokensAfterHeading[0].line.startsWith("- ");
+    const errorDetail = sectionStartsWithList
+      ? `Expect default content to precede unorderd list of ${listItemsName}: "${listSectionsDefaultContent[section]}"`
```
MaoShizhong, line 86:
```suggestion
      ? `Expect default content to precede unordered list of ${listItemsName}: "${listSectionsDefaultContent[section]}"`
```

markdownlint/TOP003_defaultSectionContent/TOP003_defaultSectionContent.js
```diff
@@ -27,92 +28,153 @@ function getListSectionErrors(sectionTokens, section) {
...
+  ) {
     listSectionErrors.push(
       createErrorObject(
-        (tokensAfterFirstContent[0] || tokensAfterHeading[0]).lineNumber,
-        errorDetail
+        tokensAfterFirstContent[0].lineNumber,
+        `Only an unordered list of ${listItemsName} can follow the default content.`,
+        {
+          lineNumber: tokensAfterFirstContent[0].lineNumber,
+          deleteCount: -1,
```
MaoShizhong, line 152: Could we define a `WHOLE_LINE` const for this and line 176? Think it'll make it easier to understand than the arbitrary `-1`
```suggestion
          deleteCount: WHOLE_LINE,
```

MaoShizhong (changes requested):

MaoShizhong (approved):

PR #28039, opened 2024-05-23
Markdownlint TOP010 Rule

### LINKED ISSUE #27836: Markdownlint: New custom ordered list lazy numbering rule
### Checks

### Describe your suggestion

This repo has `markdownlint-cli2` as a dependency, which allows us to run lint checks and fixes via our npm scripts in `package.json`, using the built-in rules in our markdownlint config files, as well as our various custom rules in the `markdownlint` directory.

Some rules (not all) have fixers included, so running our `fix:lesson` or `fix:project` scripts will automatically resolve some errors, such as using `_` instead of `*`, or not wrapping headings with blank lines etc. Some rules don't have fixers included, as they are more context-based (e.g. descriptive links) and need to be addressed on a case-by-case basis. That, or the built-in rule with all its various options make a built-in fixer a little trickier.

As per the [layout style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#lazy-numbering), we use lazy numbering with `1`s, meaning any ordered list must be prefixed with `1.` and we let the markdown converter figure out the correct numbering. We enforce this with the built-in [MD029 rule](https://github.com/DavidAnson/markdownlint/blob/v0.32.1/doc/md029.md). However, MD029 does not have a built-in fixer. And we want one. So we'll have to do it ourselves.

# What we want

We need to deactivate the MD029 rule in our base config file, then write a new custom rule that finds all lines that are ordered list items and are not prefixed with `1.`, and report the appropriate linting errors on them.
The custom rule will also need to include fix information so that when either `fix:*` scripts are run, these ordered list items' prefixes are replaced with `1.`.

```markdown
1. ONE
2. TWO
   1. child one
   2. child two
3. THREE
```
`2.`, the nested `2.`, and `3.` should flag errors.
Running the `fix:*` script should fix the above to
```markdown
1. ONE
1. TWO
   1. child one
   1. child two
1. THREE
```

The rule itself will be written in JavaScript.

If you would like to be assigned to work on this, please comment below.

# Acceptance criteria

- [ ] Within the `markdownlint` directory, make a directory for the new custom rule containing the rule `.js` file. Use the [markdownlint custom rules guide](https://github.com/DavidAnson/markdownlint/blob/main/doc/CustomRules.md) and the other custom rules to help you.
  - [ ] It must correctly report the line number, and the "detail" (expected and actual values).
  - [ ] It must include "fixInfo" that will allow the `fix:*` scripts to fix the appropriate lines.
  - [ ] It should ignore the contents of fenced code blocks.
  - [ ] It should only look at ordered lists.
- [ ] In the rule's directory, create a `tests` directory containing an `.md` file showcasing instances where the rule will flag, and when the rule passes. This test file must not flag any errors other than the custom rule.
- [ ] Inside `markdownlint/docs`, create a documentation `.md` file for the new rule so that it can be linked by the custom rule when it errors. It is okay if this documentation file includes unavoidable lint errors - it will not be run through the lint check action when you PR.
- [ ] In `.markdownlint-cli2.jsonc`, deactivate the `MD029` rule by changing its value to `false` (no object, just `false`). Its comment will need amending to say it's being overridden by the custom rule (provide the custom rule code).
- [ ] Append the custom rule's file path (to the `.js` file) to `.markdownlint-cli2.jsonc`'s `"customRules"` array.

---

### Path

Other / NA

### Lesson Url

N/A

### (Optional) Discord Name

_No response_

_No response_

BryanF1nes:
## Because

This PR creates a custom markdownlint command to help with our lazy list numbering system.

## This PR

- Disabled the prebuilt MD029 rule from markdownlint.
- Created a custom markdown lint rule TOP010.
- Created a test document that was used to test this new rule.
- Added the rule to .markdownlint-cli2.jsoinc file

## Issue

Closes #27836 

## Pull Request Requirements


markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -0,0 +1,37 @@
...
+        // https://regexr.com/80oan to test this regex
+        const updatedText = originalText.replace(/^\s*\d+\./, (match) =>
+          match.replace(/\d+/, "1")
+        );
+        const textLength = token.line.length;
+        const lineNumber = token.lineNumber;
+
+        onError({
+          lineNumber: lineNumber,
+          detail: `\n  Expected: "${updatedText}"\n  Actual: "${originalText}"\n`,
```
MaoShizhong, line 27: Ideally, we should only report `Expected: "1."; Actual: "2."` or similar, since we're only concerned with the ol prefix.
If we include the entire line, the following can realistically occur, which is a lot of noise:
image

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
+  names: ["TOP010", "ordered-list-for-lazy-numbering"],
+  description: "Ordered lists should use lazy numbering",
+  information: new URL(
+    "https://github.com/TheOdinProject/curriculum/blob/main/markdownlint/docs/TOP010.md"
+  ),
+  tags: ["lazy-numbering"],
```
MaoShizhong, line 7: [Markdownlint rules tags](https://github.com/DavidAnson/markdownlint/tree/v0.32.1?tab=readme-ov-file#tags).
Not that big of a deal considering we're not really going to be using the tags much, but for consistency, it'll be good to use the existing rule tags.
```suggestion
  tags: ["ol"],
```

markdownlint/docs/TOP010.md
```diff
@@ -0,0 +1,29 @@
+# `TOP010` - Use lazy numbering within ordered lists
+
+Tags: `ordered-list, lazy-numbering`
```
MaoShizhong, line 3: Doc file tag to match the adjust rule tag
```suggestion
Tags: `ol`
```

markdownlint/docs/TOP010.md
```diff
@@ -0,0 +1,29 @@
...
+
+Aliases: `use-lazy-numbering-in-ordered-lists`
```
MaoShizhong, line 5: Should use the name of the rule, which should match the 2nd string in the rule object's `name` array. i.e. `["TOP010", "insert-kebab-case-name-here"]`
Whichever name you feel is best, if you can change things so all reference that same name, that'd be great.

markdownlint/docs/TOP010.md
```diff
@@ -0,0 +1,29 @@
...
+
+This rule is triggered when lazy-numbering is not used in ordered lists.
```
MaoShizhong, line 7: Think we could probably elaborate *a little* more on what "lazy-numbering" is, e.g. we're specifically wanting only `1.` prefixes for ordered lists and why (convenience when changing ordered lists since the markdown converter is smart and will set the right numbers automatically).

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
+  names: ["TOP010", "ordered-list-for-lazy-numbering"],
+  description: "Ordered lists should use lazy numbering",
```
MaoShizhong, line 3: 2 things:
1. As with the above comment, a little more elaboration should make it clearer what the rule actually expects of someone (may not know what lazy numbering alone is, and even then some rules enforce `0.` prefixing).
2. Use "must" instead of "should" given it's an enforced rule.
```suggestion
  description: "Ordered lists must always use 1. as a prefix (lazy numbering)",
```

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -0,0 +1,37 @@
...
+        const textLength = token.line.length;
+        const lineNumber = token.lineNumber;
+
+        onError({
+          lineNumber: lineNumber,
+          detail: `\n  Expected: "${updatedText}"\n  Actual: "${originalText}"\n`,
+          fixInfo: {
+            lineNumber: lineNumber,
+            deleteCount: textLength,
+            insertText: updatedText,
```
MaoShizhong, line 31: Ideally, we only delete the incorrect number and replace with `1` instead. By deleting the whole line and replacing it with the entire adjusted line, it clashes with other fixers.

For example:
- TOP003's fixers replace ordered lists in the LO section with an unordered list. But if TOP010 fixes via replacing the whole line, it causes only the first LO ordered list item to be converted, and the others remain untouched.
- If an ordered list item contains other errors within its text e.g. `_` for emphasis instead of `*`, then markdownlint can prioritise one of the rules then prevent the other from firing. From my testing, it prioritises fixing `_`, so that bit gets fixed but the list item prefix remains untouched.

e.g.
```md
1. Foo
2. _Bar_
```
gets "fixed" to
```md
1. Foo
2. *Bar*
```

The fix for this would be to count the number of leading spaces, get the actual incorrect prefix number (and its length so you know how many chars to delete), and use that info to only delete the number and replace with `1`. Just tested a very rudimentary implementation of this and can confirm this approach does not clash with other rules unlike wiping and replacing the entire line.

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -0,0 +1,37 @@
...
+    params.parsers.markdownit.tokens.forEach((token) => {
+      if (
+        token.tag === "li" &&
+        token.line !== undefined &&
+        /\d/.test(token.line) &&
+        !/^\s*1\./.test(token.line)
+      ) {
+        const originalText = token.line;
+        // https://regexr.com/80oan to test this regex
+        const updatedText = originalText.replace(/^\s*\d+\./, (match) =>
```
MaoShizhong, line 19: Our 3rd regex here `/^\s*\d+\./` could probably saved to a variable outside of the `if` and reused. Then the `if` double regex test condition can be combined into one.

MaoShizhong (changes requested):
Amazing stuff @BryanF1nes :fire:
A few changes needed as detailed below - some just nits/technicalities, and some larger changes due to clashes with other rules (because of this, ideally a rule fixes by deleting as little as possible and inserting only what is necessary, as opposed to wiping a line that could contain unrelated errors).

Let me know if you have any Qs or run into any issues with this!

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
BryanF1nes: This is a good point! I didn't even consider this! I will fix!

markdownlint/docs/TOP010.md
```diff
@@ -1,10 +1,10 @@
 # `TOP010` - Use lazy numbering within ordered lists
 
-Tags: `ordered-list, lazy-numbering`
+Tags: `ol`
 
-Aliases: `use-lazy-numbering-in-ordered-lists`
+Aliases: `TOP010, useLazyNumbering`
```
MaoShizhong, line 5: Only need the 2nd element from the rule's `names` array.
The suggestion uses the one from the suggestion from the second review comment.
```suggestion
Aliases: `lazy-numbering-for-ordered-lists`
```

MaoShizhong (changes requested):
Doesn't seem to clash with the other rules from what I can see for now.
A few nits and technicalities below.

.markdownlint-cli2.jsonc
```diff
@@ -29,8 +29,10 @@
     },
     // ol-prefix
     // Enforces lazy numbering for ordered lists
-    // MD029 Disabled and overriden by TOP010 rule
-    "MD029": false,
+    // MD029 Disabled and overridden by TOP010 rule
+    "MD029": {
+      "style": "one"
+    },
```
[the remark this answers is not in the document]
BryanF1nes: Oh that was my fault, I misunderstood the previous comment 😂

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -1,36 +1,29 @@
...
-      ) {
-        const originalText = token.line;
-        // https://regexr.com/80oan to test this regex
-        const updatedText = originalText.replace(/^\s*\d+\./, (match) =>
-          match.replace(/\d+/, "1")
-        );
-        const textLength = token.line.length;
-        const lineNumber = token.lineNumber;
+      // https://regexr.com/80oan to test this regex
+      const digit = /\d/;
```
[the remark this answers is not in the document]
BryanF1nes: I was concerned about this as well, but never properly checked it. My concern was not only double digits but also in the event we receive double digits with a leading 1. I'll double check the regex and make sure it accounts for that stuff as well!

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
```diff
@@ -1,36 +1,29 @@
...
-            insertText: updatedText,
-          },
-        });
+            detail: `\n  Expected: "${lazyNumbering}"\n  Actual: "${tokenLine[0]}"\n`,
+            fixInfo: {
+              lineNumber: lineNumber,
+              deleteCount: tokenLine[0].length,
+              insertText: lazyNumbering,
+            },
+          });
```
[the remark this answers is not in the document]
BryanF1nes: Which lines exactly? Having trouble seeing it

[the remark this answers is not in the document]
MaoShizhong: Oh, that's one ugly diff... :sweat_smile: Lines 14-26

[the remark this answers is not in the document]
BryanF1nes: I think I got it!

markdownlint/TOP010_useLazyNumbering/TOP010_useLazyNumbering.js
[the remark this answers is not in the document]
MaoShizhong: `10.` etc should still trigger since `token.info` would be `'10'`

We're very unlikely to encounter this anyway. I've an open PR fixing the 4-5 only lessons that have double digit prefixes still, so we'd only encounter double digits in new lessons/changes, which we can fix in review. It's just a funky part of the CommonMark spec that requires more indentation for nested lists if an OL item prefix has more than one digit

MaoShizhong (approved):
We *should* be good to go! Amazing work for this :rocket:

Issue #27871, opened 2024-04-27
Update Lesson: Authentication Basics


01zulfi:
## Lesson overview

Update the current [Authentication Basics lesson](https://www.theodinproject.com/lessons/nodejs-authentication-basics) to align with Node Revamp

## Lesson outline

The following items define the scope of the lesson. The lesson outline is flexible; you can improve, expand, and omit items whilst writing lesson content (make sure the pull request description mentions that).

- Replace MongoDB/Mongoose instructions with equivalent PostgreSQL/node-postgres instructions
- Replace `connect-mongo` for storing sessions with [`connect-pg-simple`](https://www.npmjs.com/package/connect-pg-simple)

## Acceptance criteria

If the requirements here are not met, the work effort is not complete.

- [ ] Lesson conforms to the outline above (unless outline has been modified).
- [ ] Lesson follows our [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md).

### Get the lesson template

Download the lesson template using the following command (replace `lesson_name` with the actual lesson name):

```bash
curl -o <lesson_name>.md https://raw.githubusercontent.com/TheOdinProject/curriculum/main/templates/lesson-template.md 
```

fcasibu:
Hi, I would like to work on this one ✋

PR #28095, opened 2024-06-01
Node Revamp: Update Authentication Basics Lesson

fcasibu:
## Because

https://github.com/TheOdinProject/curriculum/issues/27871

## This PR

- Update authentication basics lesson to make use of postgresql

## Issue

Closes https://github.com/TheOdinProject/curriculum/issues/27871

## Pull Request Requirements


nodeJS/authentication/authentication_basics.md
```diff
@@ -71,9 +71,9 @@ app.get("/", (req, res) => res.render("index"));
 app.listen(3000, () => console.log("app listening on port 3000!"));
 ```
 
-Most of this should look familiar to you by now, except for the new imported middleware for express-session and passport. We are not actually going to be using express-session directly, it is a dependency that is used in the background by passport.js. You can [take a look at what the express-session package does in the Express docs](https://github.com/expressjs/session).
+So most of this should look familiar to you by now, except for the newly required modules `express-session` and `passport`. We are not actually going to be using `express-session` directly, it is a dependency that is used in the background by `passport.js`. You can [take a look at what the express-session package does in the Express docs](https://github.com/expressjs/session).
 
-Our view engine is set up to just look in the main directory, and it's looking for a template called `index.ejs` so go ahead and create that:
+Our view engine is set up to just look in the project directory, and it's looking for a template called `index.ejs` so go ahead and create that:
```
fcasibu, line 76: Maybe just a small improvement word wise, should we make this
"project directory" instead of "main directory"?

nodeJS/authentication/authentication_basics.md
```diff
@@ -348,9 +348,10 @@ You should now be able to log in using the new user you've created (the one with
...
+   - They are using MongoDB and you can replace any instance of it with PostgreSQL.
+   - In [video 3: "Your complete guide to understanding the express-session library"](https://youtu.be/J1qXK66k1y4?list=PLYQSCk-qyTW2ewJ05f_GKHtTIzjynDgjK) and [video 5: "Passport Local Configuration (Node + Passport + Express)"](https://youtu.be/xMEOT9J0IvI?list=PLYQSCk-qyTW2ewJ05f_GKHtTIzjynDgjK), it shows using the `connect-mongo` library to use your MongoDB connection to store sessions, as opposed to storing them in memory. However, since we are using PostgreSQL, we need to replace it with a library called `connect-pg-simple` instead. You can view the implementation for doing this on the [npm page for connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple).
```
fcasibu, line 354: Removed the timestamp since it does not seem to be relevant. And also added video 3 since it's shows the same thing

nodeJS/authentication/authentication_basics.md
```diff
@@ -1,62 +1,62 @@
...
+
+Next, let's set up a very minimal express app. Create a new directory and use `npm init` to start the package.json file then run the following to install all the dependencies we need:
 
 ```bash
-npm install express express-session mongoose passport passport-local ejs
+npm install express express-session pg passport passport-local ejs
 ```
 
-**Mongoose Update**: With the new 7.0.1 version of Mongoose callbacks are no longer supported when querying a database. A promise will be returned instead, meaning that you will now have to use async/await or promises to achieve the same results. If you need a refresher on async/await you can find it in the [Async And Await lesson](https://www.theodinproject.com/lessons/node-path-javascript-async-and-await) from the JavaScript Course. As you progress through this lesson you will see a blend of using async/await with try/catch blocks as well as other functions that use callbacks, which you've seen as you've progressed through the NodeJS course. You can [read more about the v7.0.1 change on the Mongoose website](https://mongoosejs.com/docs/migrating_to_7.html#dropped-callback-support).
```
fcasibu, line 41: Used note boxes instead

nodeJS/authentication/authentication_basics.md
```diff
@@ -71,9 +71,9 @@ app.get("/", (req, res) => res.render("index"));
 app.listen(3000, () => console.log("app listening on port 3000!"));
 ```
 
-Most of this should look familiar to you by now, except for the new imported middleware for express-session and passport. We are not actually going to be using express-session directly, it is a dependency that is used in the background by passport.js. You can [take a look at what the express-session package does in the Express docs](https://github.com/expressjs/session).
+Most of this should look familiar to you by now, except for the newly required modules `express-session` and `passport`. We are not actually going to be using `express-session` directly, it is a dependency that is used in the background by `passport.js`. You can [take a look at what the express-session package does in the Express docs](https://github.com/expressjs/session).
```
fcasibu, line 74: Changed to "newly required modules", since it sounds correct this way?

nodeJS/authentication/authentication_basics.md
```diff
@@ -90,7 +90,7 @@ Our view engine is set up to just look in the main directory, and it's looking f
 
 ### Creating users
 
-The first thing we need is a sign up form so we can actually create users to authenticate! In the Library Tutorial website, you learned about validating and sanitizing inputs. This is a *really good idea*, but for the sake of brevity, we're going to leave that out here. Don't forget to include sanitation and validation when you get to the project.
+The first thing we need is a sign up form so we can actually create users to authenticate! For the sake of brevity, we're going to leave sanitization and validation out here. But don't forget about it when you get to that point, we will have the opportunity to apply them later.
```
fcasibu, line 93: Sanitization is the correct term I believe. And library project is not actual anymore

nodeJS/authentication/authentication_basics.md
```diff
@@ -172,7 +174,7 @@ This function is what will be called when we use the `passport.authenticate()` f
...
+<span id='cookie'>To make sure our user is logged in, and to allow them to *stay* logged in as they move around our app, passport internally calls a function from `express-session` that uses some data to create a cookie called `connect.sid` which is stored in the user's browser</span>. These next two functions define what bit of information passport is looking for when it creates and then decodes the cookie. The reason they require us to define these functions is so that we can make sure that whatever bit of data it’s looking for actually exists in our Database! `passport.serializeUser` takes a callback which contains the information we wish to store in the session data. `passport.deserializeUser` is called when retrieving a session, where it will extract the data we "serialized" in it then ultimately attach something to the `.user` property of the request object (`req.user`) for use in the rest of the request.
```
fcasibu, line 177: Changed to "passport internally calls a function from `express-session` that uses some data to create a cookie called `connect.sid` which is stored in the user's browser" to be more technically correct

nodeJS/authentication/authentication_basics.md
```diff
@@ -226,7 +226,7 @@ app.post(
...
+As you can see, all we have to do is call `passport.authenticate()`. This middleware performs numerous functions behind the scenes. Among other things, it looks at the request body for parameters named `username` and `password` then runs the `LocalStrategy` function that we defined earlier to see if the username and password are in the database. It then creates a session cookie that gets stored in the user's browser and used in all future requests to see whether or not that user is logged in.  It can also redirect you to different routes based on whether the login is a success or a failure.  If we had a separate login page we might want to go back to that if the login failed, or we might want to take the user to their user dashboard if the login is successful.  Since we're keeping everything in the index we want to go back to "/" no matter what.
```
fcasibu, line 229: Changed to "It then creates a session cookie that gets stored in the user's browser and used in all future requests to see whether or not that user is logged in." as it seems to make more sense.

fcasibu (commented):
```javascript
bcrypt.hash(req.body.password, 10, async (err, hashedPassword) => {
  // if err, do something
  // otherwise, store hashedPassword in DB
});
```

Should we use the non-callback version instead since we're doing it with `compare` anyway? It's probably better to use non-callbacks *whenever possible* :thinking:

nodeJS/authentication/authentication_basics.md
```diff
@@ -1,62 +1,62 @@
 ### Introduction
 
-Creating users and allowing them to log in and out of your web apps is a crucial functionality that we are finally ready to learn! There is quite a bit of setup involved here, but thankfully none of it is too tricky. You'll be up and running in no time! In this lesson, we're going to be using [passportJS](https://www.passportjs.org), an excellent middleware to handle our authentication and sessions for us.
+Creating users and allowing them to log in and out of your web apps is a crucial functionality that we are finally ready to learn! There is quite a bit of setup involved here, but thankfully none of it is too tricky. You'll be up and running in no time! In this lesson, we're going to be using [passport.js](https://www.passportjs.org), an excellent middleware to handle our authentication and sessions for us.
```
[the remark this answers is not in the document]
01zulfi: sounds good to me

nodeJS/authentication/authentication_basics.md
```diff
@@ -1,62 +1,62 @@
...
-- Use PassportJS to set up user authentication with Express.
+- Use Passport.js to set up user authentication with Express.
 - Describe what bcrypt is and its use.
 - Describe what a hash is and explain the importance of password hashing.
 - Describe bcrypt's `compare` function.
 
 ### Set up
 
-We're going to be using another Mongo database, so before we begin log in to your mongo provider and create a new database and save its URL string somewhere handy.
+Before we start, create a new database or use your existing one that you've created in the Using PostgreSQL lesson.
```
01zulfi, line 23: I'd rather have them create a new database. More practice for the learners.

nodeJS/authentication/authentication_basics.md
01zulfi: sounds good to me

nodeJS/authentication/authentication_basics.md
01zulfi: love this change

nodeJS/authentication/authentication_basics.md
```diff
@@ -301,6 +301,9 @@ app.use((req, res, next) => {
 
 If you insert this code somewhere between where you instantiate the passport middleware and before you render your views, you will have access to the `currentUser` variable in all of your views, and you won't have to manually pass it into all of the controllers in which you need it.
 
+</div>
+
+
```
01zulfi: unneeded blank line
```suggestion
```

01zulfi (approved):
@fcasibu great stuff 🔥 

I'm assuming you've tested these instructions locally.

01zulfi:
> Should we use the non-callback version instead since we're doing it with compare anyway? It's probably better to use non-callbacks whenever possible 🤔

agreed, but this change is out of scope for this issue. We'll get it done later.

nodeJS/authentication/authentication_basics.md
fcasibu: Done

nodeJS/authentication/authentication_basics.md
fcasibu: Pushed this fix earlier 👍

01zulfi (approved):
lgtm 🚀

fcasibu:
Tested everything except for verifying if `connect-pg-simple` works but assumed it will work the same as `connect-mongo`, but will try to test it.

MaoShizhong (commented):
Awesome stuff @fcasibu!

While we're at it, just a tiny change we might as well make, could you change the value of `saveUninitialized` in the code example on line 65 to `false` please?
Since we'd only be using sessions for login, we don't need to save empty sessions and it makes debugging sessions harder when a bunch of empty sessions creates noise.
Just a minor change but I know has led to some headaches later in the curriculum when people try to implement sessions in db session stores, and don't question what `saveUninitialized` does, but having it set to `true` creates noise that makes debugging harder.

fcasibu:
> Awesome stuff @fcasibu!
> 
> While we're at it, just a tiny change we might as well make, could you change the value of `saveUninitialized` in the code example on line 65 to `false` please? Since we'd only be using sessions for login, we don't need to save empty sessions and it makes debugging sessions harder when a bunch of empty sessions creates noise. Just a minor change but I know has led to some headaches later in the curriculum when people try to implement sessions in db session stores, and don't question what `saveUninitialized` does, but having it set to `true` creates noise that makes debugging harder.

Empty session also happens on log out though @MaoShizhong

MaoShizhong:
@fcasibu I forgot that `req.logout` literally just clears the `.passport` property of the session object and doesn't actually destroy the session or clear the cookie.
I had been using `req.session.destroy()` and `res.clearCookie()` instead.

That specific detail I think would be best to think about when we actually come to revamping the auth section itself, since yes, `req.logout` alone would leave an empty session in the session store.

I still think we should change `saveUninitialized` from `true` to `false` though, because it's actively something that isn't needed. Login sessions don't really need nor want uninitialised sessions to be saved, and since `express-session` tells us we should explicitly set `true` or `false` for that value, we should make sure it's set to the value that matches our use case. So currently having it set to `true` doesn't really make sense, as there's no reason we want them to be saved.

fcasibu:
Thanks for the insight @MaoShizhong

MaoShizhong (approved):
Was just a nit I thought we could crack out for this while we're at it, then the rest can be addressed in the proper revamp phase. Otherwise, :fire: for this update

Issue #27620, opened 2023-05-30
 Project: Binary Search Trees - Instructions could be revised


MLGabe:
Complete the following REQUIRED checkboxes:
-   [x] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [x] The title of this issue follows the `Bug - location of bug: brief description of bug` format, e.g. `Bug - Exercises: File type incorrect for all test files`

The following checkbox is OPTIONAL:

-   [x] I would like to be assigned this issue to work on it

<hr>

**1. Description of the Bug:**
The instructions are a little misleading, as it first tells the learner to create a Tree class with an array, but then the array itself is used in a func. This can cause confusing as the approach is different if going for functional or for object oriented, as the tree only needs the root.

MaoShizhong:
Issue transferred to appropriate repo

thatblindgeye:
@CouchofTomato do you think this is worth keeping open for others to work on, or could it be closed out?

CouchofTomato:
@thatblindgeye I've added a help wanted label. I think if nobody has picked it up by the next time it's marked as stale we can close it

kyak15:
@thatblindgeye @CouchofTomato Hello! Is there any possibility I can work on this issue?

thatblindgeye:
@kyak15 yep, you're assigned

MaoShizhong:
Closing this issue (was not originally linked in the PR above).
I'm not convinced there's an issue with providing the array for constructing the Tree, regardless of factory func/class approach.

The array provides a list of values, and that array is used to generate the balanced tree, for which the root node is returned, to be stored however appropriate. If you don't construct the tree with that array, then any "root" node will not be connected to anything.

It makes sense to allow instantiation of a Tree with initial values provided, just like you can declare an array or object with initial values, rather than being forced to construct an empty array/object, then manually push each and every value you wanted to it afterwards.

I've definitely not noticed this being an issue in the community, certainly not a repeated pain point. There are other repeated and justified pain points, but they're covered in existing open issues.

PR #28142, opened 2024-06-07
Project - Binary Search Trees: Instructions revised for binary search tree implementation

kyak15:
kyak15 changes to confusing wording/requirements for the BST assignment.

## Because

The PR is to fix confusing wording/requirements in the Binary Search Tree assignment. The original issue asked the user to create a Tree class with an array, but to also use that array in the buildTree function. Thus creating confusion on an object or class directed approach for BST implementation. In my changes, I instruct the user to follow a class directed approach. The use of creating a BST with an unsorted or sorted array is not an option either. The 'Tree' class is initialized with a root of null. The user is then instructed to create the following methods inside the 'Tree' class:

-  findValue
-  insert
-  delete
-  height
- depth
- inOrder traversal
- postOrder traversal
- preOrder traversal
- levelOrder 
- isBalanced 
- reBalance 
- buildTree

The first 9 methods allow the user to focus on core BST methods and operations, without needing to worry about inheriting an array to sort and check for duplicates. After successfully implementing the core methods, the user can then use an array of the current tree values to check for balance issues. If the BST is unbalanced, reBalance uses a traversal method to get an array of values from the tree, and buildTree is called to re-balance the tree. 

## This PR
- re-wrote certain portions of the instructions to direct the user to take a class approached implementation. 
- re-writes included instructions to create methods within the 'tree' class. 
- re-writes included linking the user to the MDN page regarding classes and methods in JavaScript
- re-ordered some of the existing instructions to allow the user to focus on core BST methods first. 

## Issue
Related to #27620 

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_binary_search_trees.md
+++ b/javascript/computer_science/project_binary_search_trees.md
@@ -10,13 +10,21 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
 <div class="lesson-content__panel" markdown="1">
 
-1. Build a `Node` class/factory. It should have an attribute for the data it stores as well as its left and right children.
+1. Build a `Node` class. It should have an attribute for the data it stores as well as its left and right children.
 
-1. Build a `Tree` class/factory which accepts an array when initialized. The `Tree` class should have a `root` attribute, which uses the return value of `buildTree` which you'll write next.
+1. Build a `Tree` class which accepts an array when initialized. The `Tree` class should have a `root` attribute, which is initialized as null. 
 
-1. Write a `buildTree(array)` function that takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `buildTree` function should return the level-0 root node.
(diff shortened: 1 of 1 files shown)


MaoShizhong (commented):
Thanks @kyak15. This is a big PR that really should have started out as an issue, so you could discuss ideas with the team and we can collectively decide what work needs to be done, if any. This avoids you doing a bunch of work that we wouldn't accept for reasons that could have been discussed in the issue.

I will defer most of this to other maintainers who are more involved with the DSA side of the curriculum, but there are a couple of things I'd like to mention and some I'll push back on.

- We purposely use `class/factory` because either are fine. We do not need to enforce the use of a class here.
- If the instance accepts an array when instantiated, why set the root to `null`? That would make the array redundant. The point of passing in an array is that the `buildTree` method takes that array, builds the tree from it, then returns the root node which is set as the tree's root, which is our intention. In class terms, this would happen in the constructor. 
  ```javascript
  class Tree {
    constructor(array) {
      this.root = this.buildTree(array);
    }
    // ...
  ```
  If this step's wording has been a noticeable pain point in the community, then a more appropriate resolution would be to discuss improving its wording to be clearer on what's expected, of course without giving too much away.

MaoShizhong:
Going to close this one as on further thought, I have not observed any serious and repeated pain points that necessitate this sort of change, and even if there were, I do not feel that these specific changes would be the appropriate resolution for the reasons above.

There have been pain points around specific steps, but we have an open issue about that with a potential incoming PR.

If you've noticed in the community that there are big issues with the project instructions causing unnecessary confusion, you're welcome to open an issue so we can discuss those problems.

PR #28205, opened 2024-06-12
Update keys_in_react.md

asimn7:
feat: Added v4 to uuidv4 as uuid is not supported in the latest versions of Javascript.

## This PR

- changed from 'uuid' to 'uuidv4' and added import 

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/react/getting_started_with_react/keys_in_react.md
+++ b/react/getting_started_with_react/keys_in_react.md
@@ -46,10 +46,11 @@ Keys are passed into the component or a DOM element as a prop. You should alread
 
 ```jsx
 // a list of todos, each todo object has a task and an id
+import { v4 as uuidv4} from 'uuid';
 const todos = [
-  { task: "mow the yard", id: uuid() },
-  { task: "Work on Odin Projects", id: uuid() },
-  { task: "feed the cat", id: uuid() },
+  { task: "mow the yard", id: uuidv4() },
+  { task: "Work on Odin Projects", id: uuidv4() },
+  { task: "feed the cat", id: uuidv4() },
 ];
 
 function TodoList() {


asimn7:
Sorry, if this is a minor change, I got stuck here for quite a bit, so I thought of adding it.

asimn7:
I don;t really know how the pull thing works, I think I have to reopen though.

Asartea:
> I don;t really know how the pull thing works, I think I have to reopen though.

Yes; you've opened a pull request, and now need to wait for one of the maintainers to review it. If they approve it, they will merge it into the main branch, at which point your changes will be applied to the live version of the curriculum

xandora:
We could do something clever and reduce it to just:
```jsx
import { v4 as uuid } from 'uuid';
```
It will update the code to use the v4 `uuid`s, but keep the rest of the function calls as they currently are.

Asartea:
Actually, if all we're doing is creating UUID's, do we even need the library? I'm fairly the build in `crypto` library has a function for that

xandora:
> Actually, if all we're doing is creating UUID's, do we even need the library? I'm fairly the build in `crypto` library has a function for that

You read that too? :D
I agree though, we could do it that way. Might need some of the other maintainers to check there's no actual reason why we're using `uuid` specifically first.

Asartea:
> You read that too? :D

I plead the Fifth

> I agree though, we could do it that way. Might need some of the other maintainers to check there's no actual reason why we're using uuid specifically first.

I am not intimately familiar with this part of the curriculum, but the primary reason I could think of was that we might use it more later on, but a quick GitHub search seems to suggest this [is the only occurrence of it](https://github.com/search?q=repo:TheOdinProject/curriculum+uuid+language:Markdown+&type=code)

MaoShizhong:
We don't need to import the `uuid` library for this example; we can use the built-in [`crypto.randomUUID()` function](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) for each of those instead.

xandora:
@MaoShizhong So there's no technical or lesson-based reason we're using the `uuid` package then? Sounds like we can drop a package, which is always nice. :)

MaoShizhong (changes requested):
Thanks @asimn7 for this PR. As discussed above, we do not need the `uuid` package to be imported at all, an can instead use a built-in function to do the same thing.

The `todos` array in both examples on lines 49-53 and 86-90 can be changed to the following:

```js
const todos = [
  { task: "mow the yard", id: crypto.randomUUID() },
  { task: "Work on Odin Projects", id: crypto.randomUUID() },
  { task: "feed the cat", id: crypto.randomUUID() },
];
```

And the bad example on line 97 can be adjusted to have

```jsx
<li key={crypto.randomUUID()}>{todo.task}</li>
```

cakegod:
An auto increment would also work fine.

MaoShizhong:
@asimn7 is this something you still wish to work on?

PR #28245, opened 2024-06-19
Project HashMap: Added new section to test user hash map

### LINKED ISSUE #28117: HashMap: <Implement examples/tests to test out Hash Map code>
### Checks

### Describe your suggestion

There should be some examples or tests that users can test out their Hash Maps to see if they've implemented it correctly.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-hashmap

### (Optional) Discord Name

_No response_

_No response_

elishamutang:
## Because

Allows users to populate their hash maps where it will exceed their load factor and ensuring that their hash map methods will still work as expected after growing their buckets.

## This PR

- Added a new section to test our user hash map implementation after `Assignment` section.
- Added dummy nodes to populate user hash map in the new section. 

## Issue

Closes #28117

## Additional Information

N/A

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_hash_map.md
+++ b/javascript/computer_science/project_hash_map.md
@@ -98,9 +98,11 @@ Remember that a hash map does not preserve insertion order when you are retrievi
     test.set('lion', 'golden')
     ```
 
-1. After populating your hash map with the data above, your hash map's actual capacity should now be at `0.75`.
+1. After populating your hash map with the data above, your hash map's actual capacity should now be at `0.75` (full capacity).
 
-1. Now populate your hash map with the last node below (doing this will make your hash map exceed your current load factor, hence expanding your buckets and growing your hash map):
+1. Now with a full hash map, try overwriting a few nodes using `set(key, value)`. By right, this should only over-write the existing `values` of your nodes and not add new ones.
+
+1. After that, populate your hash map with the last node below (doing this will make your hash map exceed your current load factor, hence expanding your buckets and growing your hash map):
 
     ```javascript
     test.set('moon', 'silver')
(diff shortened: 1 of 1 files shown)


javascript/computer_science/project_hash_map.md
```diff
@@ -69,6 +71,43 @@ if (index < 0 || index >= buckets.length) {
 
 Remember that a hash map does not preserve insertion order when you are retrieving your hash map's data. It is normal and expected for keys and values to appear out of the order you inserted them in.
 
+#### Test Your Hash Map
+
+1. Create a new JavaScript file.
+
+1. Create a new instance of your hash map and set the load factor to be `0.75`.
```
JoshDevHub, line 78: How about adding some code below this like:
```suggestion
1. Create a new instance of your hash map and set the load factor to be `0.75`.

    ```javascript
    const test = new HashMap() // or HashMap() if using a factory
    ```
```
Something like this would just make it more apparent what the `test` object is that you're calling `set()` on in the next step.

JoshDevHub (changes requested):
This is an awesome idea. Thanks for making this contribution :smile: 

I have one small suggestion below, but the bigger change is that this very same project is also in the Ruby course. It's functionally the exact same lesson, just with Ruby syntax rather than JavaScript. If we're going to add this content here, we should also add it in the Ruby version of this lesson. You can find this lesson at `ruby/computer_science/project_hash_map.md`.

Don't worry if you don't know any Ruby. You can copy over pretty much everything you've written here with very few changes. The only exceptions are:
1. When you're naming a method in ruby like `set(key, value)`, write it like `#set(key, value)` instead (note the `#` at the start)
2. When you're naming a method that doesn't take arguments, omit parentheses. So if you're talking about the "values method", you'd write `#values` instead of `values()`
3. If you follow my suggestion below for creating the test hash map object, you'll want to write that as `test = HashMap.new`
4. And of course the code blocks will need to specify "ruby" instead of "javascript"

Let me know if you have any questions about any of that. I'll be glad to help.

Mclilzee:
This is an awesome addition indeed! My only suggestion would be to add the overwrite test after filling up the map and before expanding it. So that it can test if your overwriting methods are also incorrectly increasing your capacity. Your capacity should stay the same while overwriting but should extend when you add the final new value.

javascript/computer_science/project_hash_map.md
elishamutang: Oops, you're right! I will add this in. I didn't realise this. Thanks!

And for the Ruby section, I'm more than happy to make the same changes to that as well and will raise a separate PR for that.

Thanks for the feedback!

elishamutang:
> This is an awesome addition indeed! My only suggestion would be to add the overwrite test after filling up the map and before expanding it. So that it can test if your overwriting methods are also incorrectly increasing your capacity. Your capacity should stay the same while overwriting but should extend when you add the final new value.

Good catch, I will add this in. Thanks!

JoshDevHub (approved):
Approving this. Thank you for contributing :raised_hands: 

I'll also be assigning you to work on #28257

elishamutang:
No worries will do, thanks 😃

PR #28390, opened 2024-07-12
Node revamp: Add Express forms and validation lesson

### LINKED ISSUE #27642: New Lesson: Forms and Data Handling
## Lesson overview

Strategy on using forms for create/update and data sanitisation and validation

## Lesson outline

The following items define the scope of the lesson. The lesson outline is flexible; you can improve, expand, and omit items whilst writing lesson content (make sure the pull request description mentions that).

- Discuss forms and how to handle create and update entity forms in Express
- How to deal with error messages
- Discuss data sanitisation & validation via [Express validator](https://express-validator.github.io/docs/) 
- Code examples for above

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/forms is a good guide for this lesson

## Acceptance criteria

If the requirements here are not met, the work effort is not complete.

- [ ] Lesson conforms to the outline above (unless outline has been modified).
- [ ] Lesson follows our [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md).

### Get the lesson template

Download the lesson template using the following command (replace `lesson_name` with the actual lesson name):

```bash
curl -o <lesson_name>.md https://raw.githubusercontent.com/TheOdinProject/curriculum/main/templates/lesson-template.md 
```

cmedina-dev:
## Because

Node revamp lesson for the introduction to form handling in Express. Discusses how forms interact between client and server through routes and controllers, and demonstrates this with a simple CRUD application.

## This PR

* Adds a new Node.js path lesson for Express on form handling and validation.

## Issue

Closes #27642 

## Additional Information

As this lesson is incredibly lengthy due to the need to demonstrate a CRUD application via a form, I'm open to any suggestions for changes. This lesson primarily follows [the guidance from MDN found here](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/forms). A big thank you to @LexiphanicStroon for the initial write-up!

## Pull Request Requirements

### DIFF
--- a/nodeJS/express/express_forms_validation.md
+++ b/nodeJS/express/express_forms_validation.md
@@ -1,0 +1,524 @@
+### Introduction
+
+This lesson will cover how to secure your Node.js / Express applications by using the `express-validator` module to perform validation and sanitization.
+Proper form handling is how we maintain data integrity and prevent security vulnerabilities in our web applications.
+
+### Lesson overview
+
+This section contains a general overview of topics that you will learn in this lesson.
+
+- Understanding HTML form elements and their actions.
+
+- The form handling process in a client-server application.
+
+- Using `express-validator` for validation and sanitization.
+
+- Best practices for form design and data integrity.
+
+### HTML forms overview
+
+First, a brief refresher on the actual HTML.
+
+Let's create a simple HTML form, with a single text field for collecting a full name, and its associated label element.
+
+The HTML structure would look something like this:
+
+```html
+<!-- The action /users/create corresponds to a route you'll define in your Express application -->
+<form action="/users/create" method="POST" >
+  <label for="fullName">Full Name:</label>
+  <input placeholder="John Doe" type="text" name="fullName" id="fullName">
+  <button type="submit">Submit</button>
+</form>
+```
+
+We have the HTML form element itself, with an action pointing to some resource on our server, and a method defined. 
+Notice how the method corresponds to an HTTP verb, typically either `GET` or `POST`.
+
+The input needs a corresponding label element, which is what the end user actually sees on the page.
+
+Then, we have our input. 
+It has an `id` and a `name` attribute, which must match the `for` attribute defined on our label above.
+The `name` attribute is especially important, as it defines how our input will be identified in the form data sent to our server.
+
+It has a type of `text` which tells the browser what type of widget to display, what values to accept, and is used for JavaScript and CSS selectors.
+
+The `type="submit"` button can be pressed by the user to upload the data entered into all other input elements to the server (in this case there's only one).
+
+The `form` attributes define how to communicate with the server:
+
+- `action`: The resource/URL where data is sent for processing when the form is submitted. If this is not set, or is an empty string, the form will be submitted back to the current page URL
+
+- `method`: Defines the HTTP method to use (`POST` or `GET`).
+
+`POST` is more secure for sensitive information, and is used when changing or creating data in the server's database.
+`GET` is for forms that don't modify data, such as search forms, or when you want the form submission to be bookmarkable or shareable via URL.
+
+#### Form handling process
+
+The form's `action` will point to some endpoint on our server, which will allow our controller the request. The controller then communicates with the database to handle the data.
+We then generate a new or updated view with the controller's response and redirect the client. This is known as the [Post/Redirect/Get (PRG)](https://en.wikipedia.org/wiki/Post/Redirect/Get) design pattern which helps prevent duplicate `POST` requests.
+
+It might help to visualize what this entire process looks like:
+
+<pre class="mermaid">
+graph TD
+    A[Client] -->|1. POST /users/create| B(Server)
+    B -->|2. Process Data| C[(Database)]
+    B -->|"3. Redirect (302)"| D[Client]
+    D -->|4. GET /users| E(Server)
+    E -->|5. Fetch Data| F[(Database)]
+    E -->|"6. Render view (200 OK)"| G[Client]
+</pre>
+
+### Validation and sanitization
+
+Before the data from a form is sent off to our server, we must first do two things
+
+- *Validation* ensures user input meets the specified criteria (e.g. required fields, correct format).
+
+- *Sanitization* cleans user input to prevent malicious data from being processed by removing or encoding potentially malicious characters.
+
+The library we'll be using is called `express-validator`. While it simplifies both of these processes for us, 
+it's important to understand the underlying concepts of these two operations.
+
+#### Installation
+
+We start as usual by installing the correct package in the *root* folder of our project.
+
+```bash
+npm i express-validator
+```
+
+To get started using the package, we first need to *require* it in our project
+
+```javascript
+const { body, validationResult } = require('express-validator');
+```
+
+`express-validator` comes with many functions for just about every form operation you could think of, but for this lesson, we'll be sticking with `body()` and `validationResult()`.
+
+### Understanding the body() function
+
+The `body()` function allows you to specify which fields in the request body should be validated and sanitized, as well as how to handle it.
+
+```javascript
+[
+  body("birthdate", "Must be a valid date.")
+          .optional({ values: "falsy" })
+          .isISO8601() // Enforce a YYYY-MM-DD format.
+];
+```
+
+This example marks `birthdate` field as optional, 
+but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy }"` means values that aren't `undefined`, `null`, `false`, or `0` will still be validated.
+
+### Chaining validations
+
+You can also chain multiple validation methods, with unique error messages if the checks fail.
+
+```javascript
+[
+    body("name")
+    .escape()
+    .trim()
+    .isLength({ min: 1 })
+    .withMessage("Name can not be empty.")
+    .isAlpha()
+    .withMessage("Name must only contain alphabet letters."),  
+];
+```
+
+This ensures that name is not only present and trimmed, but also only contains alphabet letters.
+
+Reading [OWASP.ORG](https://blog.presidentbeef.com/blog/2020/01/14/injection-prevention-sanitizing-vs-escaping/) will give you a good idea on what escaping means and why we do it.
+
+### Validation results
+
+Once the validation rules are applied, you can use 'validationResult' to handle any validation errors.
+We use `asyncHandler` to automatically catch these errors in our async route handlers and pass them to the middleware:
+
+```javascript
+// asyncHandler lets us wrap async express routes to handle errors.
+asyncHandler(async (req, res, next) => {
+  const errors = validationResult(req);
+  
+  if (!errors.isEmpty()) {
+    return res.status(400).json({ errors: errors.array() });
+  } else {
+    res.send("Data is valid, form submitted!");
+  }
+});
+```
+
+This setup checks for any failed validation checks, and if there are any (the errors array is NOT empty), 
+then it sends a 400 code, along with any errors that may be present.
+
+### Forms and Express routes
+
+One final thing to cover is how to handle routes in Express. 
+After all, our form needs somewhere to send the data to.
+
+Since we don't want our routes handling logic, we instead use a controller.
+
+```javascript
+const asyncHandler = require("express-async-handler");
+
+exports.user_update_get = asyncHandler(async (req, res, next) => {});
+exports.user_update_post = asyncHandler(async (req, res, next) => {});
+```
+
+Inside our router, we can then assign routes which correspond to the controller's functions:
+
+```javascript
+const { Router } = require("express");
+const usersRouter = Router();
+const usersController = require("../controllers/usersController");
+
+// User update routes
+usersRouter.get("/users/:id/update", usersController.user_update_get);
+usersRouter.post("/users/:id/update", usersController.user_update_post);
+
+module.exports = usersRouter;
+```
+
+You'll notice there's a respective `GET` and `POST` route.
+* `GET` routes typically display forms or confirmation pages.
+* `POST` routes handle form submissions and data processing.
+
+You may also notice the `:id` parameter in the route.
+This lets us pass a value to those routes that we can use to tell the controller which `User` object
+we're referencing.
+
+And in our form, the action would look something like this:
+
+```html
+<!-- Example using EJS with POST to submit an update to our Express server. -->
+<form action="/users/<%= user.userId %>/update" method="POST"></form>
+```
+
+`/users/:id/update` is an *endpoint* we've created on our Express server.
+
+### Handling errors
+
+What happens if we don't give our route a valid `:id` to delete or update? We need a way to tell our users an error occurred.
+
+```javascript
+// Handles POST requests for the /users/:id/delete route.
+// Status code 400 can indicate malformed request syntax, or a missing :id in our case.
+exports.user_delete_post = asyncHandler(async (req, res, next) => {
+  if (!req.params.id) {
+    return res.status(400).send("No user id provided.");
+  }
+});
+```
+
+By responding with a status code and a message, we give the client a way to both handle the error code,
+and a way to display an appropriate error message. We could also render a new or updated view which we'll see below.
+
+### Putting it together
+
+Let's create a quick Express application to demonstrate how clients send data to servers with forms.
+
+Set up a new Express app with EJS templating:
+
+```bash
+npm init -y
+npm i express ejs express-async-handler express-validator
+```
+
+Create folders for `routes`, `views`, `controllers`, and an `app.js` file:
+
+```javascript
+// app.js
+const express = require("express");
+const app = express();
+const usersRouter = require("./routes/usersRouter");
+
+app.set("view engine", "ejs");
+app.use(express.urlencoded({ extended: true }));
+app.use("/users", usersRouter);
+app.get("/", (req, res) => res.render("index", { title: "Userbase" }));
+
+const PORT = process.env.PORT || 3000;
+app.listen(PORT, () => console.log(`Express app listening on port ${PORT}!`));
+```
+
+Then create a new router called `usersRouter.js` in the routes folder:
+
+```javascript
+// routes/usersRouter.js
+const { Router } = require("express");
+const usersController = require("../controllers/usersController");
+const usersRouter = Router();
+
+usersRouter.get("/", usersController.users_list);
+usersRouter.post("/create", usersController.users_create_post);
+
+module.exports = usersRouter;
+```
+
+Next we'll create two views:
+* `index.ejs` will display our main form.
+* `users.ejs` will list all the users we've created.
+
+```html
+<!-- views/index.ejs -->
+<!DOCTYPE html>
+<html>
+  <head>
+    <title><%= title %></title>
+  </head>
+  <body>
+    <h1><%= title %></h1>
+    <form action="/users/create" method="POST">
+      <label for="firstName">First Name: </label>
+      <input type="text" name="firstName" required>
+      <label for="lastName">Last Name: </label>
+      <input type="text" name="lastName" required>
+      <button type="submit">Submit</button>
+    </form>
+    <a href="/users">View all users</a>
+  </body>
+</html>
+```
+
+```html
+<!-- views/users.ejs -->
+<!DOCTYPE html>
+<html>
+  <body>
+    <h1><%= title %></h1>
+    <ul>
+      <% users.forEach(user => { %>
+        <li>
+          ID: <%= user.userId %>, Name: <%= user.firstName %> <%= user.lastName %> 
+        </li>
+      <% }); %> 
+    </ul>
+    <a href="/">Back to Home</a>
+  </body>
+</html>
+```
+
+And we'll finally add the logic for this router in our controller:
+
+```javascript
+// controllers/usersController.js
+// User is a factory function which lets us model what we would see in a database.
+// We will explore this concept more as we dive deeper into databases.
+const User = (userId, firstName, lastName) => ({ userId, firstName, lastName });
+const asyncHandler = require("express-async-handler");
+
+let userStorage = {};
+let userId = 0;
+
+exports.users_list = asyncHandler(async (req, res) => {
+  res.render("users", { title: "User List", users: Object.values(userStorage) });
+});
+
+exports.users_create_post = asyncHandler(async (req, res) => {
+  const { firstName, lastName } = req.body;
+  userStorage[userId] = User(userId, firstName, lastName);
+  userId++;
+  res.redirect("/users");
+});
+```
+
+We can test our server with: `node --watch app.js`.
+
+You'll see we're able to add new users at `http://localhost:3000`, as well as list all existing users at `http://localhost:3000/users`.
+
+Let's add a few methods to our controller for validating and sanitizing our form to get the type of data we want.
+
+```javascript
+const {body, validationResult} = require("express-validator")
+
+const validate_user = [
+  body("firstName").trim().isAlpha().isLength({ min: 1, max: 10 }).escape(),
+  body("lastName").trim().isAlpha().isLength({ min: 1, max: 10 }).escape(),
+];
+
+// We can pass an entire array of middleware validations to our controller.
+exports.users_create_post = [
+  validate_user,
+  asyncHandler(async (req, res) => {
+    const errors = validationResult(req);
+    if (!errors.isEmpty()) {
+      return res.status(400).render("index", { title: "Userbase", errors: errors.array() });
+    } else {
+      // Original logic
+    }
+  })
+];
+```
+
+And we need to update our `index.ejs` view to render these errors. Let's create a new partial: `errors.ejs`:
+
+```html
+<!-- views/partials/errors.ejs -->
+<% if (locals.errors) {%>
+  <ul>
+    <% errors.forEach(function(error) { %>
+    <li><%= error.msg %></li>
+    <% }); %>
+  </ul>
+<% } %>
+```
+
+And we'll include the partial just above our form in `index.ejs`.
+
+```html
+<%- include('partials/errors.ejs') %>
+```
+
+If the form is filled out incorrectly, you'll see a page containing all the errors.
+
+Now that we can create users, we also need a way to modify them.
+Let's create a form that lets us specify an ID and the values we want to update by making a new view: `update.ejs`.
+
+```html
+<!-- views/edit.ejs -->
+<%- include('partials/errors.ejs') %>
+<form action="/users/<%= user.userId %>/update" method="POST">
+  <input type="text" name="firstName" value="<%= user.firstName %>" required>
+  <input type="text" name="lastName" value="<%= user.lastName %>" required>
+  <button type="submit">Update User</button>
+</form>
+```
+
+We'll modify `users.ejs` to include a new "Update" button to make our `GET` request to the route:
+
+```html
+<li>
+  ID: <%= user.userId %>, Name: <%= user.firstName %><%= user.lastName %>
+  <a href="/users/<%= user.userId %>/update">Update</a>
+</li>
+```
+
+Just like how we added a route for creating users, we'll add routes for updating them:
+
+```javascript
+usersRouter.get("/:userId/update", usersController.users_update_get);
+usersRouter.post("/:userId/update", usersController.users_update_post);
+```
+
+Then we'll add the logic for the `GET` request into our controller:
+
+```javascript
+exports.users_update_get = asyncHandler(async (req, res) => {
+  const user = userStorage[req.params.userId];
+  if (!user) return res.status(404).send("User not found.");
+  res.render("update", { user, errors: [] });
+});
+
+exports.users_update_post = [
+  validate_user,
+  asyncHandler(async (req, res) => {
+    const { userId } = req.params;
+    const errors = validationResult(req);
+    if (!userStorage[userId]) return res.status(404).send("User not found.");
+    if (!errors.isEmpty()) return res.status(400).render("update", { 
+        errors: errors.array(), user: userStorage[userId] 
+    });
+    
+    const { firstName, lastName } = req.body;
+    userStorage[userId] = User(parseInt(userId), firstName, lastName);
+    res.redirect("/users");
+  })
+];
+```
+
+You'll see you're now able to update users, and this looks very similar to how we create our users.
+
+Finally, let's add a way to delete users as well by starting with another form inside our `users.ejs` view:
+
+```html
+<!-- In views/users.ejs -->
+<ul>
+  <% users.forEach(function(user) { %>
+  <li>ID: <%= user.userId %>, Name: <%= user.firstName %> <%= user.lastName %></li>
+  <a href="/users/<%= user.userId %>/update">Update</a>
+  <!-- This time we're sending a POST request to our route, so we need a form. -->
+  <form action="/users/<%= user.userId %>/delete" method="POST" style="display:inline;">
+    <button type="submit" onclick="return confirm('Are you sure you want to delete this user?');">Delete</button>
+  </form>
+  <% }); %>
+</ul>
+```
+
+Then we add the logic to handle the request into our controller:
+
+```javascript
+// Tell the server to delete a matching user, if any. Otherwise, respond with an error.
+exports.users_delete_post = asyncHandler(async (req, res) => {
+  const userId = req.params.userId;
+  if (!userStorage[userId]) return res.status(404).send("User not found.");
+  delete userStorage[userId];
+  res.redirect("/users");
+});
+```
+
+Don't forget to add the new route to your router!
+
+```javascript
+usersRouter.post("/:userId/delete", usersController.users_delete_post);
+```
+
+You'll notice we didn't need a `GET` route here since we're already redirecting to `/users`.
+
+We could go much deeper into working safely with forms, but we'll stop there. 
+By now you can already see how helpful `express-validator` is,
+and how you can do almost anything you want with the `req.body` object.
+
+### Assignment
+
+<div class="lesson-content__panel" markdown="1">
+
+#### Add user details
+
+Expand the `User` model we created to include more details. Implement the following fields and validations to your model:
+* Email (required, must be formatted properly)
+* Age (optional, must be a number between 18 and 120)
+* Bio (optional, maximum 200 characters)
+
+Don't forget to update the view to display these new fields!
+
+#### Implement searching
+
+What if we want to search for a specific user in a list of thousands?
+We'll need a new route and view that lets clients search our list of users.
+1. Add a form (in `users.ejs` or another view) which accepts a `name` or `email` (or both!)
+1. Create a new route `/users/search` which accepts `GET` and `POST` requests.
+1. Add the search logic to your controller which searches your list for a matching user.
+   * Your `POST` request should handle searching for the user.
+   * Your `GET` request should then render the search result.
+1. Display the search results in a new view: `search.ejs`.
+</div>
+
+### Knowledge check
+
+The following questions are an opportunity to reflect on key topics in this lesson. If you can't answer a question, click on it to review the material, but keep in mind you are not expected to memorize or master this knowledge.
+
+- [What are the attributes of a form element and what do they do?](#html-forms-overview)
+
+- [How do you install and import express-validator in your project?](#validation-and-sanitization)
+
+- [How do you validate and sanitize form input using express-validator?](#understanding-the-body-function)
+
+- [What is the difference between validation and sanitization?](#validation-and-sanitization)
+
+- [How do you handle validation errors in Express routes?](#validation-results)
+
+- [What methods can you chain with the body function to validate form fields?](#chaining-methods)
+
+- [What is the importance of escaping HTML characters in a form?](https://owasp.org/www-community/attacks/SQL_Injection)
+
+### Additional resources
+
+This section contains helpful links to related content. It isn't required, so consider it supplemental.
+
+- [Express request object](https://expressjs.com/en/api.html#req)
+- [Express response object](https://expressjs.com/en/api.html#res)
+  It's certainly not required viewing, but web dev simplified makes some of the best tutorials on Express. Take a watch if you need a refresher.
+- [Web dev simplified](https://youtu.be/SccSCuHhOw0?si=2dZ5Y4dvxyh7jpcy)
--- a/nodeJS/express/express_forms_validation.md
+++ b/nodeJS/express/express_forms_validation.md
@@ -65,7 +65,7 @@ The library we'll be using is called `express-validator`. While it simplifies bo
 We start as usual by installing the correct package in the *root* folder of our project.
 
 ```bash
-npm i express-validator
+npm install express-validator
 ```
 
 To get started using the package, we first need to *require* it in our project
@@ -196,7 +196,7 @@ Set up a new Express app with EJS templating:
 
 ```bash
 npm init -y
-npm i express ejs express-async-handler express-validator
+npm install express ejs express-async-handler express-validator
 ```
 
 Create folders for `routes`, `views`, `controllers`, and an `app.js` file:
--- a/nodeJS/express/express_forms_validation.md
+++ b/nodeJS/express/express_forms_validation.md
@@ -57,19 +57,8 @@ The `form` attributes define how to communicate with the server:
 #### Form handling process
 
 The form's `action` will point to some endpoint on our server, which will allow our controller the request. The controller then communicates with the database to handle the data.
+
 We then generate a new or updated view with the controller's response and redirect the client. This is known as the [Post/Redirect/Get (PRG)](https://en.wikipedia.org/wiki/Post/Redirect/Get) design pattern which helps prevent duplicate `POST` requests.
-
-It might help to visualize what this entire process looks like:
-
-<pre class="mermaid">
-graph TD
-    A[Client] -->|1. POST /users/create| B(Server)
-    B -->|2. Process Data| C[(Database)]
-    B -->|"3. Redirect (302)"| D[Client]
-    D -->|4. GET /users| E(Server)
-    E -->|5. Fetch Data| F[(Database)]
-    E -->|"6. Render view (200 OK)"| G[Client]
-</pre>
 
 ### Validation and sanitization
 
--- a/nodeJS/express/express_forms_validation.md
+++ b/nodeJS/express/express_forms_validation.md
@@ -32,12 +32,12 @@ The HTML structure would look something like this:
 </form>
 ```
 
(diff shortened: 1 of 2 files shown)


MaoShizhong (changes requested):
Haven't had time for a thorough read of the contents itself, but some general things that would need addressing anyway:

- You'd need to make sure you're following our [layout style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md), as there are quite a lot of lint errors for style. We have [instructions for local linting and fixing](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md#curriculum-linting). The errors should be somewhat self-explanatory and some can be fixed with the `fix:lesson` script, while some would need manually addressing.
- We don't need blank lines between list items of the same list, as that ends up rendering as `<li><p>...</p></li>` rather than just `<li>...</li>`, adding unnecessary whitespace.
- To get the correct syntax highlighting, if a code block contains EJS then its language needs to be `ejs` instead of `html`.
- For consistency between all of our lessons, we should ideally use the full npm commands e.g. `npm install express-validator`
- There's the odd JS code block that has single quotes which should be double quotes for file/course consistency.
- In the "Chaining validations" section, you demonstrate escaping the `name` value, then link to the (very nice) article just below to explain what it is and why it's done.

  That article indicates that escaping is best done for a value at the point of use, not at the point of receipt. i.e. store the data pure, then when we output that data, escape it then (so using `<%= %>` for escaped output in EJS). If we escape it upon receipt, when we go to output it, we're forced to either use unescaped output `<%- %>` (which is far riskier and docs generally recommend against it), or manually unescape with some library so we can escape it again upon output with `<%= %>`.

  That's redundant since the value isn't dangerous until it's used, and it'll only be dangerous then if used in a context where the value contains context-specific dangerous characters. So rather than escape then have to unescape it every time we need to use it (then escape again on output), or use unescaped output instead, as per that article (and other articles like https://benhoyt.com/writings/dont-sanitize-do-escape/), we should keep the data pure on input, then escape on output via `<%= %>`.

  Ideally then, the lesson goes through why we make sure to use `<%= %>` for output as it'll handle the escaping of "dangerous HTML characters" for us.
  
Otherwise, an initial scan through is :fire:

MaoShizhong:
Oh, also just noticed that you do a lot of manual line breaks - you won't need to do this as those single line breaks won't be rendered any differently. Keep individual paragraphs on the same line in the file, else it'll be hella annoying to keep fixing manual line breaks if you have to reword anything.

cmedina-dev:
Thank you for all the feedback Mao! I'm not sure how I missed that linter, but it was very helpful!

> * In the "Chaining validations" section, you demonstrate escaping the `name` value, then link to the (very nice) article just below to explain what it is and why it's done.
>       That article indicates that escaping is best done for a value at the point of use, not at the point of receipt. i.e. store the data pure, then when we output that data, escape it then (so using `<%= %>` for escaped output in EJS). If we escape it upon receipt, when we go to output it, we're forced to either use unescaped output `<%- %>` (which is far riskier and docs generally recommend against it), or manually unescape with some library so we can escape it again upon output with `<%= %>`.
>       That's redundant since the value isn't dangerous until it's used, and it'll only be dangerous then if used in a context where the value contains context-specific dangerous characters. So rather than escape then have to unescape it every time we need to use it (then escape again on output), or use unescaped output instead, as per that article (and other articles like https://benhoyt.com/writings/dont-sanitize-do-escape/), we should keep the data pure on input, then escape on output via `<%= %>`.
>       Ideally then, the lesson goes through why we make sure to use `<%= %>` for output as it'll handle the escaping of "dangerous HTML characters" for us.

This is a great point. I rewrote this chaining portion a few times trying to strike a balance between depth and brevity, and felt it could definitely be improved. I've adjusted that section to better cover escaping data on use. Hopefully this new commit hits the level of detail you're looking for.

MaoShizhong:
A cursory read tbrough is looking great! It'll be a while befote I get time again to properly sit down and review the contents more thoroughly. Currently only got time for bits here and there.

And yeah, after looking more into the escaping thing and talking with others in the field, I came to find the MDN tutorial's way of escaping on input, then using unescaped output with a "be careful that unescaped output can be dangerous" warning rather strange. As a result, lots of people have ended up unnecessarily escaping input data, then having the unescape it again for use, often only to escape it again via escaped output :shrug:

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
+### Introduction
+
+This lesson will cover how to secure your Node.js / Express applications by using the `express-validator` module to perform validation and sanitization. Proper form handling is how we maintain data integrity and prevent security vulnerabilities in our web applications.
```
01zulfi, line 3: Intro should start with "We'll learn about forms..." before saying "we'll learn express-validator".

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+### Lesson overview
+
+This section contains a general overview of topics that you will learn in this lesson.
+
+- Understanding HTML form elements and their actions.
+- The form handling process in a client-server application.
```
01zulfi, line 10: seems redundant

```suggestion
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+### HTML forms overview
+
+First, a brief refresher on the actual HTML.
+
+Let's create a simple HTML form, with a single text field for collecting a full name, and its associated label element.
+
+The HTML structure would look something like this:
+
+```html
+<!-- The action /users/create corresponds to a route you'll define in your Express application -->
```
01zulfi, line 23: We already explain what `action` attribute is below, don't think we need to do it here.
```suggestion
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+It has a type of `text` which tells the browser what type of widget to display, what values to accept, and is used for JavaScript and CSS selectors.
+
+The `type="submit"` button can be pressed by the user to upload the data entered into all other input elements to the server (in this case there's only one).
+
+The `form` attributes define how to communicate with the server:
+
+- `action`: The resource/URL where data is sent for processing when the form is submitted. If this is not set, or is an empty string, the form will be submitted back to the current page URL
+- `method`: Defines the HTTP method to use (`POST` or `GET`).
+
+`POST` is more secure for sensitive information, and is used when changing or creating data in the server's database. `GET` is for forms that don't modify data, such as search forms, or when you want the form submission to be bookmarkable or shareable via URL.
```
01zulfi, line 46: > `POST` is more secure for sensitive information

Oh interesting, could you share a reference link or something for this?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+#### Form handling process
+
+The form's `action` will point to some endpoint on our server, which will allow our controller the request. The controller then communicates with the database to handle the data.
+
+We then generate a new or updated view with the controller's response and redirect the client. This is known as the [Post/Redirect/Get (PRG)](https://en.wikipedia.org/wiki/Post/Redirect/Get) design pattern which helps prevent duplicate `POST` requests.
```
01zulfi, line 52: Nice inclusion of PRG

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+```javascript
+[
+  body("birthdate", "Must be a valid date.")
+          .optional({ values: "falsy" })
+          .isISO8601() // Enforce a YYYY-MM-DD format.
+];
+```
+
+This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy }"` means values that aren't `undefined`, `null`, `false`, or `0` will still be validated.
```
01zulfi, line 91:
```suggestion
This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy" }` means values that aren't `undefined`, `null`, `false`, or `0` will still be validated.
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+When unescaped, this would be rendered into HTML as:
+
+```html
+<div>
+  About Me: <script>alert("Hacked!");</script>!
+</div>
+```
+
+To prevent this cross-site scripting (XSS) attack, we can *escape* the output (you may also see this referred to as *encoding*). Escaped HTML replaces special characters, like `<`, with their respective HTML entities, in this case `&lt;`. In EJS, we can escape the output using `<%= %>`.
```
01zulfi, line 129: nice inclusion of XSS

can we include a reference link for cross-site scripting attack, a MDN link or Wikipedia will do

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+```ejs
+<div>
+  About Me: <%= username %>!
+</div>
+
+// The escaped output is now rendered harmless:
+// About Me: &lt;script&gt;alert(&quot;Hacked!&quot;);&lt;/script&gt;!
+```
+
+So why we don't chain `.escape()` in our Express server's `body()` function instead? We would then need to unescape the input when we go to output it again, or use unescaped output with `<%- %>`, which is undesirable as shown above.
```
01zulfi, line 140: Can we add a link to `.escape()` and show a simple code example of its usage. I say this because it's all new to our learners.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+Reading [OWASP.ORG](https://blog.presidentbeef.com/blog/2020/01/14/injection-prevention-sanitizing-vs-escaping/) will give you a good idea on what escaping means and why we do it.
```
01zulfi, line 142: This needs to be an assignment reading. Can you remove it from here and add it there.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+One final thing to cover is how to handle routes in Express. After all, our form needs somewhere to send the data to.
+
+Since we don't want our routes handling logic, we instead use a controller.
+
+```javascript
+const asyncHandler = require("express-async-handler");
+
+exports.user_update_get = asyncHandler(async (req, res, next) => {});
+exports.user_update_post = asyncHandler(async (req, res, next) => {});
```
01zulfi, line 173: Can we use camelCase for all function names?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+// Handles POST requests for the /users/:id/delete route.
+// Status code 400 can indicate malformed request syntax, or a missing :id in our case.
+exports.user_delete_post = asyncHandler(async (req, res, next) => {
+  if (!req.params.id) {
+    return res.status(400).send("No user id provided.");
+  }
+});
+```
+
+By responding with a status code and a message, we give the client a way to both handle the error code, and a way to display an appropriate error message. We could also render a new or updated view which we'll see below.
```
01zulfi, line 220: Not sure if the lesson needs to handle params error handling here. What do you think?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+usersRouter.get("/", usersController.users_list);
+usersRouter.post("/create", usersController.users_create_post);
+
+module.exports = usersRouter;
+```
+
+Next we'll create two views:
+
+- `index.ejs` will display our main form.
+- `users.ejs` will list all the users we've created.
```
01zulfi, line 267: I'd rather have this switched. To me it makes more sense for the index page to display data and a link to the form page. What do you think?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+```javascript
+// controllers/usersController.js
+// User is a factory function which lets us model what we would see in a database.
+// We will explore this concept more as we dive deeper into databases.
+const User = (userId, firstName, lastName) => ({ userId, firstName, lastName });
+const asyncHandler = require("express-async-handler");
+
+let userStorage = {};
+let userId = 0;
```
01zulfi, line 318: Lack of databases nicely handled!

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+// We can pass an entire array of middleware validations to our controller.
+exports.users_create_post = [
+  validate_user,
+  asyncHandler(async (req, res) => {
+    const errors = validationResult(req);
+    if (!errors.isEmpty()) {
+      return res.status(400).render("index", { title: "Userbase", errors: errors.array() });
+    } else {
+      // Original logic
```
01zulfi, line 354: This comment makes more sense
```suggestion
      // Create a new user in the database from the validated and sanitised data
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+```ejs
+<!-- views/partials/errors.ejs -->
+<% if (locals.errors) {%>
+  <ul>
+    <% errors.forEach(function(error) { %>
+    <li><%= error.msg %></li>
+    <% }); %>
+  </ul>
+<% } %>
+```
```
01zulfi, line 371: Nice inclusion of partials

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+And we'll include the partial just above our form in `index.ejs`.
+
+```ejs
+<%- include("partials/errors.ejs") %>
+```
+
+If the form is filled out incorrectly, you'll see a page containing all the errors.
```
01zulfi, line 379: > you'll see a page containing all the errors.

won't the errors show up on the existing form page?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+Create folders for `routes`, `views`, `controllers`, and an `app.js` file:
+
+```javascript
+// app.js
+const express = require("express");
+const app = express();
+const usersRouter = require("./routes/usersRouter");
+
+app.set("view engine", "ejs");
+app.use(express.urlencoded({ extended: true }));
```
01zulfi, line 242: The lesson needs to explain the `express.urlencoded` middleware.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+
+```javascript
+usersRouter.post("/:userId/delete", usersController.users_delete_post);
+```
+
+You'll notice we didn't need a `GET` route here since we're already redirecting to `/users`.
+
+We could go much deeper into working safely with forms, but we'll stop there. By now you can already see how helpful `express-validator` is, and how you can do almost anything you want with the `req.body` object.
+
+### Assignment
```
01zulfi, line 475: Assignment needs a link to the express validator docs and include some of them as required readings https://express-validator.github.io/docs/

My suggestions:
https://express-validator.github.io/docs/guides/getting-started
https://express-validator.github.io/docs/guides/validation-chain

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+- [What is the difference between validation and sanitization?](#validation-and-sanitization)
+- [How do you handle validation errors in Express routes?](#validation-results)
+- [What is the importance of escaping HTML characters in a form?](https://owasp.org/www-community/attacks/SQL_Injection)
+
+### Additional resources
+
+This section contains helpful links to related content. It isn't required, so consider it supplemental.
+
+- [Express request object](https://expressjs.com/en/api.html#req)
+- [Express response object](https://expressjs.com/en/api.html#res)
```
01zulfi, line 518: Unnecessary additional resources
```suggestion
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+  It's certainly not required viewing, but web dev simplified makes some of the best tutorials on Express. Take a watch if you need a refresher.
+- [Web dev simplified](https://youtu.be/SccSCuHhOw0?si=2dZ5Y4dvxyh7jpcy)
```
01zulfi, line 520: Not sure if we need to include the complete video, do you have timestamp suggestion?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,520 @@
...
+The following questions are an opportunity to reflect on key topics in this lesson. If you can't answer a question, click on it to review the material, but keep in mind you are not expected to memorize or master this knowledge.
+
+- [What are the attributes of a form element and what do they do?](#html-forms-overview)
+- [How do you install and import express-validator in your project?](#validation-and-sanitization)
+- [How do you validate and sanitize form input using express-validator?](#understanding-the-body-function)
+- [What is the difference between validation and sanitization?](#validation-and-sanitization)
+- [How do you handle validation errors in Express routes?](#validation-results)
+- [What is the importance of escaping HTML characters in a form?](https://owasp.org/www-community/attacks/SQL_Injection)
+
+### Additional resources
```
01zulfi, line 513: https://express-validator.github.io/docs/guides/customizing#implementing-a-custom-validator

^ custom validators can be an additional resource

nodeJS/express/express_forms_validation.md
01zulfi, line 475: I like this diagram a lot. What do you think about showing it right before the Assignment section and credit MDN?

https://developer.mozilla.org/en-US/docs/Learn/Server-side/Express_Nodejs/forms/web_server_form_handling.png

01zulfi (changes requested):
@cmedina-dev great stuff, can't wait to see the lesson live 🔥🔥

It's a pretty chonky lesson, so left a chonky review below 😆

nodeJS/express/express_forms_validation.md
MaoShizhong: Just want to tack on that the list item seems to have the incorrect formatting with split lines. We'd and also need a more descriptive link for accessibility, e.g. `Web Dev Simplified has a [tutorial on submitting forms and parsing form data in Express] if you want a refresher.` or whatever's most appropriate.

nodeJS/express/express_forms_validation.md
cmedina-dev: Maybe `secure` is the wrong choice of word here? The idea is that `POST` is more about "security through obscurity." A `GET` request could expose credentials in the URL as seen here: https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url

I'm open to other ways of rewording this though

nodeJS/express/express_forms_validation.md
cmedina-dev: How might we want this to look in terms of usage for the code snippet? @MaoShizhong made a good point in that it's not really best practice to encourage escaping server-side. I can't really think of an instance that `escape()` should or would be used.

MaoShizhong: Perhaps something like the following (very rough and wordy suggestion):

> So why not escape the data when we receive it by adding `.escape()` to the end of our `body()` validation chain instead? "Dangerous characters" are only dangerous at the point of use, and also only in certain contexts. What's "dangerous" for HTML may not be dangerous for SQL and vice versa, and they won't pose risks until they get used in those contexts.
> 
> Also, if we have data with HTML entities after escaping them, if we used escaped output with (`<%= %>`), then `&lt;` wouldn't become `<` but output literally as the text `&lt;`. We'd need to either unescape them then re-escape them via escaped output, or end up using unescaped output (`<%- %>`), which, like using `.innerHTML`, is undesirable as shown above.

nodeJS/express/express_forms_validation.md
cmedina-dev: I have no strong opinion on it. I think it's fine to remove since the `### Validation results``` covers handling validation errors. Just lmk and I'll take it out.

nodeJS/express/express_forms_validation.md
cmedina-dev: Yeah makes sense to me. I think the logic came from an earlier iteration I wrote, but I'm good to switch it 👍

nodeJS/express/express_forms_validation.md
cmedina-dev: Yeah that was just clumsy writing on my part. It should be `the page contains`.

nodeJS/express/express_forms_validation.md
cmedina-dev: Totally agree! I actually had a mermaid diagram because I really wanted some sort of visual representation and wasn't sure if directly embedding an image was something we could do (wasn't sure of the logistics wrt crediting, licensing, etc.). I'll add this in.

nodeJS/express/express_forms_validation.md
01zulfi: Cool, we can reword it like you have above.

nodeJS/express/express_forms_validation.md
01zulfi: Yeah, let's remove this.

nodeJS/express/express_forms_validation.md
01zulfi: I think our site supports mermaid diagrams, we could go with that or this image, whichever you prefer.

nodeJS/express/express_forms_validation.md
cmedina-dev: I've updated the phrasing but left the full video in. I'll leave this open just in case we decide we want a timestamp, but this came from Nathaniel so I don't have a recommendation for a timestamp. For what it's worth the video has timestamps built into the video, so I don't think this is too big of an issue.

cmedina-dev:
First round of feedback addressed!

As a heads up, the "Putting it together" section had a bit of a rewrite to incorporate switching `index.ejs` and `users.ejs` so a full review of this would be appreciated 👍

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+    .withMessage("Name can not be empty.")
+    .isAlpha()
+    .withMessage("Name must only contain alphabet letters."),  
+];
+```
+
+This ensures that `name` is not only present and trimmed, but also only contains alphabet letters.
+
+While this might work for outputs we know won't have special characters, like names or ages, we also have to consider situations that do allow those characters. For example, when writing their "About Me" description, what would happen if the client decides to inject JavaScript code instead?
+
```
01zulfi, line 109: We can add a heading here

```suggestion

### Escaping user input

```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+  
+  if (!errors.isEmpty()) {
+    return res.status(400).json({ errors: errors.array() });
+  } else {
+    res.send("Data is valid, form submitted!");
+  }
+});
+```
+
+This setup checks for any failed validation checks, and if there are any (the errors array is NOT empty), then it sends a 400 code, along with any errors that may be present.
```
01zulfi, line 159: We could add a reference link for 400:
```suggestion
This setup checks for any failed validation checks, and if there are any (the errors array is NOT empty), then the server sends a [400 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400), along with any errors that may be present.
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+### Forms and Express routes
+
+One final thing to cover is how to handle routes in Express. After all, our form needs somewhere to send the data to.
+
+Since we don't want our routes handling logic, we instead use a controller.
```
01zulfi, line 165: Redundant sentence, we assume learner already learned about routes and controller responsibilities in the previous lessons.

```suggestion
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+app.set("view engine", "ejs");
+app.use(express.urlencoded({ extended: true }));
+app.use("/users", usersRouter);
+app.get("/", (req, res) => res.render("index", { title: "User List", users: usersStorage.getUsers() }));
+
+const PORT = process.env.PORT || 3000;
+app.listen(PORT, () => console.log(`Express app listening on port ${PORT}!`));
+```
+
+Most simple forms will use the `application/x-www-form-urlencoded` `Content-Type` when sending data to the server. Express, however, can't natively parse that data. We can use the `express.urlencoded()` middleware to handle this for us. When `extended` is `false`,  our server will only accept a `string` or an array of data, so we set it to `true` for some added flexibility. Note that if the `Content-Type` doesn't match `application/x-www-form-urlencoded`, then your server will show the data as an empty object `{}`.
```
01zulfi, line 233: wording suggestion:
```suggestion
Most simple forms will use the `Content-Type: application/x-www-form-urlencoded` HTTP header when sending data to the server. Express, however, can't natively parse that data. We can use the `express.urlencoded()` middleware to handle this for us and automatically set form's data to the `req.body` field. When `extended` is `false`,  our server will only accept a `string` or an array of data, so we set it to `true` for some added flexibility. Note that if the `Content-Type` doesn't match `application/x-www-form-urlencoded`, then your server will show the data as an empty object `{}`.
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+The logic for this router will go inside of our controller:
+
+```javascript
+// controllers/usersController.js
+// User is a factory function which acts as a model for our simulated database.
+// We will explore this concept more as we later dive deeper into databases.
+const User = (firstName, lastName) => ({ firstName, lastName });
+const asyncHandler = require("express-async-handler");
+const usersStorage = require("../storages/usersStorage");
```
01zulfi, line 303: Since you've refactored to use a `usersStorage` class, I reckon the `User` factory can be an internal implementation detail for that class. No need to have it be external. What do you think?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+    updateUser(id, {firstName, lastName}) {
+        this.storage[id] = { id, firstName, lastName };
+    }
+
+    deleteUser(id) {
+        delete this.storage[id];
+    }
+}
+// Rather than exporting the class, we can export an instance of the class by instantiating it.
+// This ensures only one instance of this class can exist, also known as the "singleton" pattern.
```
01zulfi, line 350: Super cool addition. Content like this really makes our lessons stand out. Good job!

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+1. Display the search results in a new view: `search.ejs`.
+
+</div>
+
+#### Further Reading
+
+- [OWASP.ORG](https://blog.presidentbeef.com/blog/2020/01/14/injection-prevention-sanitizing-vs-escaping/) will give you a good idea on what escaping means and how it can help further secure your web applications.
+- [express-validator](https://express-validator.github.io/docs/) contains the full documentation for `validator`, with some important sections being:
+  - [Getting Started](https://express-validator.github.io/docs/guides/getting-started)
+  - [Validation Chains](https://express-validator.github.io/docs/guides/validation-chain)
```
01zulfi, line 547: This section needs to nested within the Assignment block.

01zulfi (changes requested):

nodeJS/express/express_forms_validation.md
01zulfi, line 1: just saw this, could you rename the file to `forms_and_data_handling.md`

nodeJS/express/express_forms_validation.md
cmedina-dev: Good idea 👍

nodeJS/express/express_forms_validation.md
cmedina-dev: I did give this some thought and wasn't sure how relevant I wanted the User factory to be, since it's not really the focal point. But given that the `usersStorage` class was split off, I was considering either a `models/user.js` or like you said just embedding it within the class directly. How do you feel about either approach?

nodeJS/express/express_forms_validation.md
cmedina-dev: I assume you mean moving the `</div>` to below this section? Just want to double check before I mark this as resolved.

01zulfi: yep

nodeJS/express/express_forms_validation.md
01zulfi: Learners haven't been introduced to the "models" as a concept yet, so I wouldn't want `modles/user.js`. Just making it internal to the `userStorage` class should be good enough.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+We have the HTML form element itself, with an action pointing to some resource on our server, and a method defined. Notice how the method corresponds to an HTTP verb, typically either `GET` or `POST`.
+
+The input needs a corresponding label element, which is what the end user actually sees on the page.
+
+Then, we have our input. It has an `id` and a `name` attribute, which must match the `for` attribute defined on our label above. The `name` attribute is especially important, as it defines how our input will be identified in the form data sent to our server.
+
+It has a type of `text` which tells the browser what type of widget to display, what values to accept, and is used for JavaScript and CSS selectors.
+
+The `type="submit"` button can be pressed by the user to upload the data entered into all other input elements to the server (in this case there's only one).
```
MaoShizhong, line 37: I feel like the small details here are unnecessary. Learners will have used forms, labels and inputs already, so we don't need to explain about the purpose of the label, its `for` attribute etc.

We can streamline this bit to just say we have a form with a text input, a label for it, and a submit button. Then we can include the details on the importance of the `name` field, since that's the most important thing we haven't really used in the curriculum yet.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+The form's `action` will point to some endpoint on our server, which will allow our controller the request. The controller then communicates with the database to handle the data.
+
+We then generate a new or updated view with the controller's response and redirect the client. This is known as the [Post/Redirect/Get (PRG)](https://en.wikipedia.org/wiki/Post/Redirect/Get) design pattern which helps prevent duplicate `POST` requests.
+
+### Validation and sanitization
+
+Before the data from a form is sent off to our server, we must first do two things
+
+- *Validation* ensures user input meets the specified criteria (e.g. required fields, correct format).
+- *Sanitization* cleans user input to prevent malicious data from being processed by removing or encoding potentially malicious characters.
```
MaoShizhong, line 57: I'm a little iffy on the wording of this part. Currently, it reads like we all our form data must be *both* validated and sanitised. Of course we must validate all form data as necessary, but sanitisation (especially at receipt) is only needed if absolutely needed (as discussed before and changes made already).

e.g. outputting unescaped data is bad, but we handle that by escaping when outputting, not handling the sanitisation at receipt of input.

I can personally see this bit confusing learners who are told they *must* sanitise all form data, but then are shown examples where sanitisation does not occur and we only escape on output.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+The library we'll be using is called `express-validator`. While it simplifies both of these processes for us, it's important to understand the underlying concepts of these two operations.
+
+#### Installation
```
MaoShizhong, line 61: For accessibility
```suggestion
#### Installation express-validator
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+We start as usual by installing the correct package in the *root* folder of our project.
+
+```bash
+npm install express-validator
+```
+
+To get started using the package, we first need to *require* it in our project
```
MaoShizhong, line 69: Nit
```suggestion
To get started using the package, we first need to `require` it in our project
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+### Understanding the body() function
+
+The `body()` function allows you to specify which fields in the request body should be validated and sanitized, as well as how to handle it.
+
+```javascript
+[
+  body("birthdate", "Must be a valid date.")
+          .optional({ values: "falsy" })
+          .isISO8601() // Enforce a YYYY-MM-DD format.
+];
```
MaoShizhong, line 86: Consistency of 2-space indentation
```suggestion
[
  body("birthdate", "Must be a valid date.")
    .optional({ values: "falsy" })
    .isISO8601() // Enforce a YYYY-MM-DD format.
];
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+```javascript
+[
+    body("name")
+    .trim()
+    .isLength({ min: 1 })
+    .withMessage("Name can not be empty.")
+    .isAlpha()
+    .withMessage("Name must only contain alphabet letters."),  
+];
```
MaoShizhong, line 103: Indentation, and also suggestion to use `notEmpty()` instead of `isLength({ min: 1 })`, thought that part isn't a blocker.
```suggestion
[
  body("name")
    .trim()
    .notEmpty()
    .withMessage("Name can not be empty.")
    .isAlpha()
    .withMessage("Name must only contain alphabet letters."),  
];
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+  About Me: <%= username %>!
+</div>
+
+// The escaped output is now rendered harmless:
+// About Me: &lt;script&gt;alert(&quot;Hacked!&quot;);&lt;/script&gt;!
+```
+
+So why not escape the data when we receive it by adding .escape() to the end of our body() validation chain instead? "Dangerous characters" are only dangerous at the point of use, and also only in certain contexts. What's "dangerous" for HTML may not be dangerous for SQL and vice versa, and they won't pose risks until they get used in those contexts.
+
+Also, if we have data with HTML entities after escaping them, if we used escaped output with (<%= %>), then &lt; wouldn't become < but output literally as the text &lt;. We'd need to either unescape them then re-escape them via escaped output, or end up using unescaped output (<%- %>), which, like using .innerHTML, is undesirable as shown above.
```
MaoShizhong, line 140: Just adding backticks as necessary, which HTML entities will also require, else they'd render as the character and not the text on the website.

```suggestion
Also, if we have data with HTML entities after escaping them, if we used escaped output with (`<%= %>`), then `&lt;` wouldn't become `<` but output literally as the text `&lt;`. We'd need to either unescape them then re-escape them via escaped output, or end up using unescaped output (`<%- %>`), which, like using `.innerHTML`, is undesirable as shown above.
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+### Validation results
+
+Once the validation rules are applied, you can use 'validationResult' to handle any validation errors. We use `asyncHandler` to automatically catch these errors in our async route handlers and pass them to the middleware:
```
MaoShizhong, line 144:
```suggestion
Once the validation rules are applied, you can use `validationResult` to handle any validation errors. We use `asyncHandler` to automatically catch these errors in our async route handlers and pass them to the middleware:
```

nodeJS/express/express_forms_validation.md
MaoShizhong, line 159: Need to change to use `res.render` to render the appropriate page (e.g. the form page but with the errors listed, or if successful, perhaps a redirect to the homepage or wherever is most appropriate).
Sending JSON or a string would be more relevant for APIs later but not when we're still using view templates.

The last paragraph will also need adjusting to reflect the change.

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+module.exports = usersRouter;
+```
+
+You'll notice there's a respective `GET` and `POST` route.
+
+- `GET` routes typically display forms or confirmation pages.
+- `POST` routes handle form submissions and data processing.
+
+You may also notice the `:id` parameter in the route. This lets us pass a value to those routes that we can use to tell the controller which `User` object we're referencing.
```
MaoShizhong, line 193: Also a redundant sentence since route params are explained in the Routes lesson

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+});
+
+exports.usersCreatePost = asyncHandler(async (req, res) => {
+  const { firstName, lastName } = req.body;
+  usersStorage.addUser(User(firstName, lastName));
+  res.redirect("/");
+});
+```
+
+And we'll use a storage class to hold the users we create. In real-world scenarios, you would almost certainly be using a database for this, which you'll explore further in upcoming lessons.
```
MaoShizhong, line 316: Nit suggestion to reinforce the point
```suggestion
And we'll use a storage class to hold the users we create. In real-world scenarios, you would almost certainly be using a database for this, which you'll explore further in upcoming lessons. This class is just for demonstration purposes before we get there.
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    updateUser(id, {firstName, lastName}) {
+        this.storage[id] = { id, firstName, lastName };
+    }
+
+    deleteUser(id) {
+        delete this.storage[id];
+    }
+}
```
MaoShizhong, line 348: Could this use 2-space indentation please?

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+const lengthErr = "must be between 1 and 10 characters.";
+
+const validateUser = [
+    body("firstName").trim()
+      .isAlpha().withMessage(`First name ${alphaErr}`)
+      .isLength({ min: 1, max: 10 }).withMessage(`First name ${lengthErr}`),
+    body("lastName").trim()
+      .isAlpha().withMessage(`Last name ${alphaErr}`)
+      .isLength({ min: 1, max: 10 }).withMessage(`Last name ${lengthErr}`),
+];
```
MaoShizhong, line 373:
```suggestion
const {body, validationResult} = require("express-validator");

const alphaErr = "must only contain letters.";
const lengthErr = "must be between 1 and 10 characters.";

const validateUser = [
  body("firstName").trim()
    .isAlpha().withMessage(`First name ${alphaErr}`)
    .isLength({ min: 1, max: 10 }).withMessage(`First name ${lengthErr}`),
  body("lastName").trim()
    .isAlpha().withMessage(`Last name ${alphaErr}`)
    .isLength({ min: 1, max: 10 }).withMessage(`Last name ${lengthErr}`),
];
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+
+// We can pass an entire array of middleware validations to our controller.
+exports.usersCreatePost = [
+  validateUser,
+  asyncHandler(async (req, res) => {
+    const errors = validationResult(req);
+    if (!errors.isEmpty()) return res.status(400).render("users", { 
+      title: "User List", errors: errors.array() 
+    });
```
MaoShizhong, line 382:
```suggestion
    if (!errors.isEmpty()) {
      return res.status(400).render("users", {
        title: "User List",
        errors: errors.array(),
      });
    }
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+```
+
+And we need to update our `users.ejs` view to render these errors. Let's create a new partial: `errors.ejs`:
+
+```ejs
+<!-- views/partials/errors.ejs -->
+<% if (locals.errors) {%>
+  <ul>
+    <% errors.forEach(function(error) { %>
+    <li><%= error.msg %></li>
```
MaoShizhong, line 398: Nit
```suggestion
      <li><%= error.msg %></li>
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+        }
+        if (!errors.isEmpty()) return res.status(400).render("update", {
+            errors: errors.array(), user: user
+        });
+        const { firstName, lastName } = req.body;
+        usersStorage.updateUser(req.params.id, User(firstName, lastName));
+        res.redirect("/");
+    })
+];
+```
```
MaoShizhong, line 467: 2-space indentation for these as well please

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+</ul>
+```
+
+Then we add the logic to handle the request into our controller:
+
+```javascript
+// Tell the server to delete a matching user, if any. Otherwise, respond with an error.
+exports.usersDeletePost = asyncHandler(async (req, res) => {
+  const user = usersStorage.getUser(req.params.id);
+  if (!user) return res.status(404).send("User not found.");
```
MaoShizhong, line 495: Just maintaining consistency with what users read across this lesson and other lessons
```suggestion
  if (!user) {
    return res.status(404).send("User not found.");
  }
```

nodeJS/express/express_forms_validation.md
```diff
@@ -0,0 +1,565 @@
...
+   - Your `GET` request should then render the search result.
+1. Display the search results in a new view: `search.ejs`.
+
+</div>
+
+#### Further Reading
+
+- [OWASP.ORG](https://blog.presidentbeef.com/blog/2020/01/14/injection-prevention-sanitizing-vs-escaping/) will give you a good idea on what escaping means and how it can help further secure your web applications.
+- [express-validator](https://express-validator.github.io/docs/) contains the full documentation for `validator`, with some important sections being:
+  - [Getting Started](https://express-validator.github.io/docs/guides/getting-started)
```
MaoShizhong, line 546: Nit for accessibility again, else navigating by link, a SR will just read "Getting Started - link"
```suggestion
  - [Getting Started with express-validator](https://express-validator.github.io/docs/guides/getting-started)
```

MaoShizhong (changes requested):
Some more comments, many of which are just consistency/style-related things.

cmedina-dev:
Second round of feedback addressed. Some notes:

I removed references to `res.send()` and `res.json()`. When it comes to validation, these weren't practical and I felt weren't useful to the code samples because learners would have to go out of their way to actually encounter these in the example rather than being shown.

I think everything else should be addressed. Let me know if any other last changes need to be made.

nodeJS/express/express_forms_validation.md
MaoShizhong: Oops @cmedina-dev this should be `Installing express-validator` :eyes: Forgot to change the first word too

01zulfi (approved):
let's go 🚀

MaoShizhong (approved):
Ship it.

PR #28483, opened 2024-07-22
Views: Update views with locals information

jasonHYLam:
## Because

Some learners expressed confusion over a `locals.errors` variable in the subsequent Express Form section. This information about the locals object and defining variables in it has been added to the Views lesson.  

## This PR

- Added specific context about needing to define variables in the `locals` object when rendering a template file with `res.render`.
- Added note about undefined variables resulting in a reference error, with a code example.

## Issue

Closes #XXXXX

## Pull Request Requirements


nodeJS/express/views.md
```diff
@@ -78,9 +78,29 @@ Start the server and go to the `/` route in the browser. You should see:
...
 
-When you hit the `/` route, `res.render("index", { message: "EJS rocks!" });` is the line that sends back the response. Since we've already defined the `views` and `view engine` app properties, the first argument of `res.render` is programmed to look for "a template called index in the specified folder", while the second argument is an object of variables that are to be made available to that specific template.
+When you hit the `/` route, `res.render("index", { message: "EJS rocks!" });` is the line that sends back the response. Since we've already defined the `views` and `view engine` app properties, the first argument of `res.render` is programmed to look for "a template called index in the specified folder", while the second argument is an object of variables that are to be made available to that specific template. This object of variables is known as the `locals` object which is always generated by EJS. Say you define a variable called `foo` in the `locals` object. This variable can be accessed in the view file with either `locals.foo` or `foo`. A few more details can be found in the [Express docs](https://expressjs.com/en/5x/api.html#res.locals).
```
MaoShizhong, line 83: A couple of bits:

- This `locals` thing is an EJS thing. [EJS docs](https://ejs.co/) - go to the `Options` section then to the `localsName` option. It's the default name for the object in EJS where locals (be that set via `res.locals` in Express, or by passing it in via the 2nd arg in `res.render). The Express docs for `res.locals` won't be relevant for this specific point. It can still be linked when discussing `res.locals`, but when discussing the EJS `locals` variable, that's an EJS-specific thing (e.g. Pug doesn't have it, and handles locals variables in a different way).
- We can split the discussion about `locals` into its own subsection with the heading `#### The locals variable in EJS`, rewording as necessary to accommodate the structure change.

nodeJS/express/views.md
```diff
@@ -78,9 +78,29 @@ Start the server and go to the `/` route in the browser. You should see:
...
+
+In the example above, the `message` property exists on the `locals` object as `locals.message`, to which we can access with either `locals.message` or more directly with `message`.
+
+<div class="lesson-note lesson-note--tip" markdown="1">
+
+#### Undefined variables in locals
+
+If we try to access a variable in a rendered template file that was not defined in `res.locals`, this can cause a reference error. For instance if we try to access an undefined `foo` variable, `locals.foo` will return undefined, while `foo` will result in an reference error. Verify this by adding `locals.foo`, and then `foo` below `message` in `index.ejs`:
```
MaoShizhong, line 91: Not just `res.locals` but also if not passed in via the 2nd argument in `res.render`.

nodeJS/express/views.md
jasonHYLam: Thanks for the feedback!

Would it be worth saying that it is possible to implement local variables with both `res.locals` and the `locals` object in `res.render()`, to give the learner more options to consider? And then explicitly say that this is relevant to Express rather than EJS?

I'll add something to convey the point that the `localsName` / `locals` syntax is specific to EJS.

nodeJS/express/views.md
jasonHYLam: I'll change this to include both the `locals` argument of `res.render`, and `res.locals`.

nodeJS/express/views.md
MaoShizhong: No need for `app.locals`. Most people will only need `res.locals` if something like that's appropriate, and if they really need an `app.locals` thing, they can discover that as needed.

I'll address the first part in a separate review comment

nodeJS/express/views.md
```diff
@@ -78,10 +78,41 @@ Start the server and go to the `/` route in the browser. You should see:
...
+In the example above, how did the template file know about the `message` variable? The answer is that we passed a `locals` object as the second argument to `res.render()`, which contained a `message` property. This makes it available for template engines to access local variables. We could have also set local variables onto the [`res.locals`](https://expressjs.com/en/5x/api.html#res.locals) object in our route, like so:
+
+```javascript
+app.get("/", (req, res) => {
+  res.locals.message = "EJS rocks!";
+  res.render("index");
+});
+```
+
+Then, EJS stores this in an object which is named `locals` by default (check the localsName bullet point in the Options section in the [EJS docs](https://ejs.co/#docs) to learn more). Thus the `message` variable is accessible with either `locals.message` or simply `message`.
```
MaoShizhong, line 96: Just a wording suggestion that I think *might* make the point clearer that EJS `locals` just has access to anything in Express's `res.locals` *and* anything we pass in via `res.render`. The current wording feels like you have to pick one or the other. I don't think we'd really need a code block for the `res.locals` thing with this, nor link to the EJS docs about the `localsName` thing, since I *highly* doubt many learners will be trying to configure the EJS engine itself.

The wording below *is* just a suggestion, so feel free to play around with it if you wish, otherwise let me know what you think.

```suggestion
In the example above, how did the template file know about the `message` variable? When we render the view, EJS has access to any properties from the object we pass into `res.render`, as well as any properties on [Express's res.locals object](https://expressjs.com/en/5x/api.html#res.locals) (`res.locals` can be useful if you need to pass values to the view in one middleware function, but won't call `res.render` until later in the middleware chain).

EJS will store these properties in an object called `locals`, which you can access in the view. Similarly to the global `window` object in browsers, this allows you to access the `message` variable in the view via `locals.message`, or simply just `message`.
```

nodeJS/express/views.md
```diff
@@ -78,10 +78,41 @@ Start the server and go to the `/` route in the browser. You should see:
...
+
+```ejs
+<html>
+  <body>
+    <%= message %>
+    <%= locals.foo %>
+    <%= foo %>
+  </body>
+</html>
+```
```
MaoShizhong, line 112: Non-blocking nit - what do you think? Just because with the original, it won't show the output of `message` and `locals.foo`, since it'll just show the error page for `foo`'s reference error straight away.
````suggestion
If we try to access a variable in a rendered template file that was not defined in the `locals` argument of `res.render` or `res.locals`, this can cause a reference error. For instance if we try to access an undefined `foo` variable, `locals.foo` will return undefined, while `foo` will result in an reference error. Verify this by outputting `locals.foo` in `index.ejs`, then replacing it with `foo`:

```ejs
<html>
  <body>
    <%= message %>
    <!-- replace the below with the output of just foo -->
    <%= locals.foo %>
  </body>
</html>
```
````

Comment (jasonHYLam) [reply nodeJS/express/views.md:112]: This is good, your way conveys what I wanted the learner to do much more effectively.

File: nodeJS/express/views.md (hunk 1 above)
Comment (jasonHYLam) [reply nodeJS/express/views.md]: Alright, that sounds reasonable. Do I need to resolve this (and all conversations) for a PR to go through, by the way?

File: nodeJS/express/views.md (hunk 3 above)
Comment (jasonHYLam) [reply nodeJS/express/views.md:96]: Your suggestion sounds good to me, I suppose it's better to focus on EJS rather than the specifics of setting local variables in Express. And it's probably best to not confuse the learner by adding too much information. The analogy with the global `window` object is also a good call.

File: nodeJS/express/views.md [hunk 5]
```diff
@@ -78,10 +78,33 @@ Start the server and go to the `/` route in the browser. You should see:
...
 
 When you hit the `/` route, `res.render("index", { message: "EJS rocks!" });` is the line that sends back the response. Since we've already defined the `views` and `view engine` app properties, the first argument of `res.render` is programmed to look for "a template called index in the specified folder", while the second argument is an object of variables that are to be made available to that specific template.
 
+### The locals variable in EJS
+
+In the example above, how did the template file know about the `message` variable? When we render the view, EJS has access to any properties from the object we pass into `res.render`, as well as any properties on [Express's res.locals object](https://expressjs.com/en/5x/api.html#res.locals) (`res.locals` can be useful if you need to pass values to the view in one middleware function, but won't call `res.render` until later in the middleware chain).
+EJS will store these properties in an object called `locals`, which you can access in the view. Similarly to the global `window` object in browsers, this allows you to access the `message` variable in the view via `locals.message`, or simply just `message`.
```
MaoShizhong, line 88: Blank line required for paragraph separation, else we'd have one long paragraph since single line breaks don't do anything.
```suggestion

EJS will store these properties in an object called `locals`, which you can access in the view. Similarly to the global `window` object in browsers, this allows you to access the `message` variable in the view via `locals.message`, or simply just `message`.
```

nodeJS/express/views.md
MaoShizhong, line 1: Nit - just noticed the lesson overview items don't end with `.` 
Could you add them to both items please? We've got a lint rule in review that'll enforce ending LO items with periods.

MaoShizhong (changes requested):
Just these nits then I'm happy with this addition.

nodeJS/express/views.md
MaoShizhong: Unresolved conversations won't block PRs (but even if they did, we can mark them resolved if we needed to anyway).

nodeJS/express/views.md
jasonHYLam: Thanks for catching that! Blank line now added.

nodeJS/express/views.md
jasonHYLam: Added fullstops, thanks for catching.

MaoShizhong (approved):
:pray: Thanks for adding this to the lesson!

Issue #28488, opened 2024-07-22
SQL Zoo: The Postgres engine on sql zoo is bugged.


DrantDumani:
### Checks

- [X] This is not a duplicate of an existing issue (please have a look through our [open issues list](https://github.com/TheOdinProject/curriculum/issues) to make sure)
- [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [X] Would you like to work on this issue?

### Describe your suggestion

I posted about this in the discord, but I figured this was the proper way to go about this. The POSTGRES option on SqlZoo is bugged. I was doing everything in MySql and I switched to Postgres because I saw the chat. However, it results in some odd behavior.

Some examples:

In the [More Join](https://sqlzoo.net/wiki/More_JOIN_operations) tutorial, there's a question that asks you to find the id of the film titled 'Casablanca'. This is the result in postgres
```sql
SELECT id FROM movie WHERE title = 'Casablanca';
-- Outputs TWO ids!
-- id
--  132689
--  60576357
```
This isn't the case in MySql where you only get one id. The next question, which asks you to list the cast list for Casablanca is also affected. 
```sql
SELECT name FROM actor JOIN casting ON actor.id = actorid
WHERE casting.movieid = 132689;
-- Outputs no confirmation on a correct answer
-- The second id just outputs an empty list
```
And then there's the NULL tutorial, which is what prompted me to do some research and make this post to begin with. The question asks you to list all teacher in the `teacher` table who don't have a department. Here's a link to the tutorial: https://sqlzoo.net/wiki/Using_Null
But there's one problem.
```sql
SELECT name from teacher WHERE dept IS NULL;
--ERROR: relation "teacher" does not exist LINE 1: SELECT name from teacher WHERE dept IS NULL; ^
```
According to this post on postgresql.org, this has been an issue with Sql zoo since 2008. https://www.postgresql.org/message-id/396486430808120859p12f01d57uc050e458f7a1309a@mail.gmail.com
Sorry about bringing all of this up after you guys made the PR to change from mySql to Postgres for this projecy. I was in the middle of doing everything in MySql before it was brought up.

As a solution, I think we can go back to mentioning the use of MySql. But there can also be a note saying that, while future projects would be using Postgres instead, the solutions to problems on SqlZoo do not require anything further than engine agnostic sql syntax, similar to the assignments in the previous lesson that used sqlite.

### Path

[...]

PR #28490, opened 2024-07-23
Revert "SQLZoo: Change Engine to PostgreSQL"

MaoShizhong:
Reverts TheOdinProject/curriculum#28485 due to #28488

Seems to be an SQLZoo bug we have little control over, but is currently blocking people from completing the assignment. Revert to unblock then explore if there's anything we can do otherwise.


MaoShizhong:
Once approved, please merge instead of letting me merge, just in case I head off to sleep and delay this because of that.

Asartea:
LGTM :rocket:

01zulfi (approved):

PR #28503, opened 2024-07-25
Update project_restaurant_page.md

john-walter-munene:
In this commit. I add a direct approach to hosting Webpack applications on GitHub Pages. The existing guide is relatively difficult to follow. 

The Gist I have written is also publicly available should anyone wish to improve it or provide more context to the approach.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/project_restaurant_page.md
+++ b/javascript/organizing_your_javascript_code/project_restaurant_page.md
@@ -65,4 +65,5 @@ When creating a new repo on GitHub there is an option to specify a `.gitignore` 
 
     1. Recall that the **source branch** for GitHub Pages is set in your repository's settings.
 
+Should you have troubles following the above guidelines, please checkout this gist by one of our students on [hosting Webpack application on GitHub Pages](https://gist.github.com/john-walter-munene/d64c84fd11088c6f0f82f713810f2aea)
 </div>


MaoShizhong (commented):
Thanks for this PR.

I'm not particularly sure the instructions in the new gist will solve any problems with the current gist. It includes a lot of things that are not necessary for the Restaurant Page project (or the other projects in the JS course). For example, nothing in the course includes client-side routing, so handling that would be redundant and can confuse learners who don't know why they're being instructed to handle that. Same with CopyWebpackPlugin, as nothing in dist should be stuff that isn't part of the JS dependency graph, or not otherwise loaded via things like html-loader (for filepaths in the HTML template) or css-loader (for urls in stylesheets). We shouldn't be adding in anything to instructions that are specifically for features we are not implementing.

At least from my personal experience, most of the issues with deployment here I've seen have come.from people either missing out a step or two, typing a command incorrectly, or adding in their own steps that the instructions do not say to do.

For reference, #27961 will rewrite the Webpack lesson to be more explicit about what loaders will be needed in different situations.

I'll close this PR, as even if the gist was amended, I don't think we the right approach to people confused about the provided instructions is to give them completely different instructions. We should provide a single set of instructions that covers only what's needed for this project. If you wish to propose amendments to the existing instructions, please do open an issue so ideas can be discussed there.

PR #28535, opened 2024-07-31
Views: Make writing style more imperative throughout

jasonHYLam:
## Because

Learners have addressed frustration in this lesson regarding the writing style. The shift from passive to imperative mood made it vague as to when the code is just a hypothetical example (to not be followed) or should be followed. This PR aims to improve the clarity of the lesson by making the writing style more imperative when the learner should follow the code examples. Additionally the `users.forEach` code example is fleshed out a bit more to make it clear what the learner should do. The order of the lesson content  was also rearranged slightly to make the lesson content be followed in a more logical manner, eg. moving the note on `<%-` to where it is first relevant at the first `includes` example.

## This PR

- Changed writing style from passive to imperative mood for sections that the learner should follow
- Fleshed out code example for `users.forEach`
- Moved the note on `<%-` to where it is first relevant at the first `includes` example
- Modified the nested directories example so it modifies the previous example

## Issue

Closes #XXXXX

## Pull Request Requirements


01zulfi (commented):
I don't have a lot to add here. @MaoShizhong, what is your review?

nodeJS/express/views.md
```diff
@@ -147,48 +161,37 @@ You can insert this component into another EJS file like so:
...
+This will include the navbar with links in `index.ejs`. Reusable templates can be used to include headers and footers in all of your pages.
 
-app.get("/", (req, res) => {
-  res.render("index", { links: links });
-});
-```
+Note the use of the raw output tag `<%-` with the `include` which is used to avoid double-escaping the HTML output.
 
-Here's another example of how to use `includes` to dynamically render a list of variables:
+Let's use `includes` to dynamically render a list of variables a different way. Add `const users = ["Bob", "Bill", "Biff"];` to `app.js` just below the `links` variable. Then create a new view called `user.ejs` in the `views` directory and add the following `<li><%= user %></li>` line. Finally add the following block to `index.ejs`:
```
MaoShizhong, line 168: Nit: drop the `s` from `includes` (3rd word)

Non-nit: the `app.js` and `user.ejs` examples should be separate fenced code blocks like all the others, rather than inline as current. The consistency makes it easier to follow with the same importance to each new part of code. The inlines feel more like "quick references to an prior example" sort of thing. You **must** also instruct to pass the `users` array to `index.ejs` when rendering:

```js
app.get("/", (req, res) => {
  res.render("index", { links: links, users: users });
});
```

(Additional definite nit, non-blocking - what about using `Rose` and `Cake` instead of Bill and Bob? :eyes:)

nodeJS/express/views.md
```diff
@@ -147,48 +161,37 @@ You can insert this component into another EJS file like so:
...
 #### Directories within the views folder
 
-We can have nested directories of EJS template files within the views. For example, to render the template file `./views/user/show.ejs`, we'll need to provide the relative path like so:
+Let's create nested directories of EJS template files within the views. Change the `user.ejs` in the `views` directory to `users/user.ejs`, and in `index.ejs` change the path from `user` to `users/user` in the `users.forEach` block:
 
 ```javascript
-// in res.render
-res.render("user/show");
-
 // in include
```
MaoShizhong, line 187:
```suggestion
<!-- index.ejs -->
```

nodeJS/express/views.md
```diff
@@ -133,7 +133,21 @@ Say you have the following navbar component called `"navbar.ejs"`:
...
+  { href: "/", text: "Home" },
+  { href: "about", text: "About" },
+];
+
+app.get("/", (req, res) => {
+  res.render("index", { links: links });
+});
+```
+
+Then, modify the previously created `index.ejs` to `include` the `navbar` with the `links` variable like so:
```
MaoShizhong, line 150:
```suggestion
Then, modify the previously created `index.ejs` to remove the `<%= message %>` line and instead `include` the `navbar` with the `links` variable like so:
```

nodeJS/express/views.md
```diff
@@ -147,48 +161,37 @@ You can insert this component into another EJS file like so:
...
-
-Note that the navbar expects a `links` value. To pass this data into the navbar, you can pass it when rendering `index.ejs` which contains the navbar. Modify `app.js` such that a `links` object is defined and passed into the `render` function in the `"/"` route handler:
-
-```javascript
-// app.js
-const links = [
-  { href: "/", text: "Home" },
-  { href: "about", text: "About" },
-];
+This will include the navbar with links in `index.ejs`. Reusable templates can be used to include headers and footers in all of your pages.
```
MaoShizhong, line 164: Nitty suggestion
```suggestion
This will include the navbar with links in `index.ejs`. Run your server and see! Reusable templates can be used to include headers and footers in all of your pages.
```

MaoShizhong (changes requested):

nodeJS/express/views.md
MaoShizhong, line 187: Currently it has 
````md
```javascript
// in include
````
Which should now be changed the following, given the contents change:
````md
```ejs
<!-- index.ejs -->
````

Comment (jasonHYLam) [reply nodeJS/express/views.md:187]: Thanks for the catch, I'll try to be more mindful of needing to change the language when changing a codeblock.

File: nodeJS/express/views.md (hunk 1 above)
Comment (jasonHYLam) [reply nodeJS/express/views.md:168]: The Biff was accidental and needed for the alliteration, but including Rose and Cake would definitely be nice!

File: nodeJS/express/views.md (hunk 4 above)
Comment (jasonHYLam) [reply nodeJS/express/views.md]: Have added!

File: nodeJS/express/views.md (hunk 3 above)
Comment (jasonHYLam) [reply nodeJS/express/views.md]: Have now added!

File: nodeJS/express/views.md (hunk 1 above)
Comment (MaoShizhong) [reply nodeJS/express/views.md]: The alliteration was nice but thought we might as well name drop a few more people :P

File: nodeJS/express/views.md [hunk 5]
```diff
@@ -161,11 +161,30 @@ Then, modify the previously created `index.ejs` to `include` the `navbar` with t
...
 
 Note the use of the raw output tag `<%-` with the `include` which is used to avoid double-escaping the HTML output.
 
-Let's use `includes` to dynamically render a list of variables a different way. Add `const users = ["Bob", "Bill", "Biff"];` to `app.js` just below the `links` variable. Then create a new view called `user.ejs` in the `views` directory and add the following `<li><%= user %></li>` line. Finally add the following block to `index.ejs`:
+Let's use `include` to dynamically render a list of variables a different way. In `app.js`, add the following `users` array just below the `links` variable, and pass it to the `render` function when rendering `index`:
+
+```javascript
+// app.js
+
+const links = [...];
```
MaoShizhong, line 173: We should either have the full links thing like before, or have a comment with the omission of something the keep as before. In this scenario, keeping the links array in full as before makes more sense to me. While `[...]` will probably make sense to many, I think it'd require a little thought time to realise, as opposed to just reading it straight up.
```suggestion
const links = [
  { href: "/", text: "Home" },
  { href: "about", text: "About" },
];
```

MaoShizhong (changes requested):
Happy with this once the below change is made :fire: Much appreciated for making these improvements! Real glad we're getting good practical feedback from learners.

nodeJS/express/views.md
jasonHYLam: That sounds good, I've now changed to use the full `links` variable.

MaoShizhong (approved):
Amazing work :fire:

PR #28545, opened 2024-08-01
feat(foundations): clarify that using copyrighted images is dangerous

NikitaRevenco:
## Because

I remember when I first saw this warning in foundations I was like "eh... whatever", not fully realising the implications that not following the license will have.

I think most people understand and will abide by the warning, but I think there's value in making it more "dangerous" so that learners will be careful.

## This PR

Adds a few notes on using images on the web

## Issue

Closes #XXXXX

## Additional Information

https://discord.com/channels/505093832157691914/514204569572605952/1268606210911768586

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project curriculum contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x] If this PR addresses an open issue, it is linked in the `Issue` section
-   [x] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [x] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

PR #28565, opened 2024-08-04
Project: Recipe : Add note about image usage and licensing

XAJX179:
## Because

This lesson is the first time students use github pages which is hosting your website publicly, to ensure they dont use copyrighted images and spread awarness about checking license requirements before using any image on web.

## This PR

- removed suggestion of using normal google search images which could be copyrighted.
- added a note which has:
-  guidelines on using images from web.
-  included sources for free-to-use images and license review instructions.

## Issue

Closes #XXXXX

## Additional Information

https://discord.com/channels/505093832157691914/505093832157691916/1269557778704371725

## Pull Request Requirements


MaoShizhong:
Mentioned this PR in #28545 already as I feel like they're tightly coupled in terms of exactly how the topic is approached.

ManonLef:
This is indeed related to a PR I was just reviewing. Mao linked to it. This PR here seems to put this information in an earlier lesson (so not the landing page, but in the lesson page). Before I review, it might be good to decide where we put this information @TheOdinProject/foundations . I am inclined to put it in this recipes project as opposed to the later landing page. Would you agree?

ManonLef:
One concern I might have is that this is the first project and there's already quite an overload of extra information in here. I will think about this a bit more later because there might be a more appropriate spot

XAJX179:
> One concern I might have is that this is the first project and there's already quite an overload of extra information in here. I will think about this a bit more later because there might be a more appropriate spot

my concern was -
- If we share this information early on at the very first time students like me use images which is the Odin Recipe Project, it would be better since we wont have copyrighted images inside our repos, till we get to a later project and then discover we cant use just any image we find on internet.

ManonLef:
> * If we share this information early on at the very first time students like me use images which is the Odin Recipe Project, it would be better since we wont have copyrighted images inside our repos, till we get to a later project and then discover we cant use just any image we find on internet.

Yes I think that's a very good point and also the reason I'm inclined to put it in the curriculum sooner rather than later. Thanks for elaborating on that!

Eduardo06sp:
@ManonLef Honestly, I wholeheartedly believe that we need a lesson dedicated to just copyright & how it related to what we do in TOP.

It doesn't have to be super detailed and long, but it does need to cover the basics of what we generally can and cannot do in certain situations with our projects.

If this sounds like a ton of new information for an already overwhelmed, new user (because it does sound like it), then I would propose that we provide generic/blank/custom-made images for the first few projects so the user does not have to worry about any of this yet.

We could even use blank boxes made in CSS, since the pictures aren't really important at all for the first few projects. It would be really nice to use pictures though that are free-to-use.

XAJX179:
> @ManonLef  Honestly, I wholeheartedly believe that we need a lesson dedicated to just copyright & how it relates to what we do in TOP.
> 
> It doesn't have to be super detailed and long, but it does need to cover the basics of what we generally can and cannot do in certain situations with our projects.
> 
> If this sounds like a ton of new information for an already overwhelmed, new user (because it does seem like it), then I propose that we provide generic/blank/custom-made images for the first few projects so the user does not have to worry about any of this in the beginning.
> 
> We could even use blank boxes made in CSS, since the pictures aren't really important at all for the first few projects. However, it would be really nice to use pictures that are free-to-use.

@Eduardo06sp 
We cannot provide images to users because its a recipes project and we wont know which recipe they will choose. and personally i think having/seeing images of your favourite dish is always fun in odin-recipes website. : ) 
Also i was aware of too much info that's why i tried to keep my note as concise and short as possible below 180 words i believe.

abhilash15500:
> @ManonLef Honestly, I wholeheartedly believe that we need a lesson dedicated to just copyright & how it relates to what we do in TOP.
> 
> It doesn't have to be super detailed and long, but it does need to cover the basics of what we generally can and cannot do in certain situations with our projects.
> 
> If this sounds like a ton of new information for an already overwhelmed, new user (because it does seem like it), then I propose that we provide generic/blank/custom-made images for the first few projects so the user does not have to worry about any of this in the beginning.
> 
> We could even use blank boxes made in CSS, since the pictures aren't really important at all for the first few projects. However, it would be really nice to use pictures that are free-to-use.

I completely agree with @Eduardo06sp. I've noticed many learners using copyrighted images in their projects. We definitely need a dedicated lesson on this. The placement of the lesson can be decided by the top core members based on what they think is best.

KevinMulhern (changes requested):
Thanks for doing this, and sorry for the delay @XAJX179 

Would you mind moving it to the links and images lesson please? I think it contextually makes more sense in that lesson, and we're already using way too many notes in this project.

KevinMulhern (approved):
Great stuff, thanks @XAJX179!

PR #28731, opened 2024-09-01
Function Basics: Add question to Knowledge Check

youssef-el-atmani:
## Because

- To point learners to think about the difference between function declaration & expression
- Helps reinforce their understanding and retention of those two concepts.

## This PR

- Add a question about **function expression**
- Add a question about the difference between **function expressions** and **function declarations**

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/foundations/javascript_basics/function_basics.md
+++ b/foundations/javascript_basics/function_basics.md
@@ -80,6 +80,8 @@ The following questions are an opportunity to reflect on key topics in this less
 - [What is function scope?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions#function_scope_and_conflicts)
 - [What are return values?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Return_values#what_are_return_values)
 - [What are arrow functions?](https://javascript.info/arrow-functions-basics)
+- [What is a function expression?](https://javascript.info/function-expressions)
+- [What is the difference between a function declaration and a function expression?](https://javascript.info/function-expressions#function-expression-vs-function-declaration)
 
 ### Additional resources
 
--- a/foundations/javascript_basics/function_basics.md
+++ b/foundations/javascript_basics/function_basics.md
@@ -80,7 +80,6 @@ The following questions are an opportunity to reflect on key topics in this less
 - [What is function scope?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions#function_scope_and_conflicts)
 - [What are return values?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Return_values#what_are_return_values)
 - [What are arrow functions?](https://javascript.info/arrow-functions-basics)
-- [What is a function expression?](https://javascript.info/function-expressions)
 - [What is the difference between a function declaration and a function expression?](https://javascript.info/function-expressions#function-expression-vs-function-declaration)
 
 ### Additional resources


foundations/javascript_basics/function_basics.md
```diff
@@ -80,6 +80,8 @@ The following questions are an opportunity to reflect on key topics in this less
 - [What is function scope?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions#function_scope_and_conflicts)
 - [What are return values?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Return_values#what_are_return_values)
 - [What are arrow functions?](https://javascript.info/arrow-functions-basics)
+- [What is a function expression?](https://javascript.info/function-expressions)
```
MaoShizhong, line 83: Arrow functions are a type of function expression, just a different syntax to the traditional syntax (plus some other small caveats). Given the next proposed question which covers declarations vs expressions, I think this particular one is a little redundant.
```suggestion
```

zachmmeyer (changes requested):
Your `Because` is blank on this PR. Why do you want to add these to an already overflowing "Additional Resources" section on this lesson. Does The Odin Project lack these explanations during it's lessons? Does this lesson need more clarification? 
What is the reason beyond just wanting to add more resources?

youssef-el-atmani:
> Your `Because` is blank on this PR. Why do you want to add these to an already overflowing "Additional Resources" section on this lesson. Does The Odin Project lack these explanations during it's lessons? Does this lesson need more clarification? What is the reason beyond just wanting to add more resources?

Hi @zachmmeyer, thank you for reviewing my contribution!

I'd like to clarify that my addition is intended for the "Knowledge Check" section rather than the "Additional Resources" section. The motivation for this addition is to encourage learners to actively engage with  the concepts they've just learned, rather than simply consuming information: which is the same case for all the "Knowledge Check" questions, and that's why I didn't fill up the `Because` part.

Since the questions in the "Knowledge Check" section are meant to prompt reflection and reinforce understanding, I believe my *little contribution* aligns with this purpose. It is a light addition, similar to the existing questions such as:

[What are functions useful for?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions)
[How do you invoke a function?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions#invoking_functions)
[What are anonymous functions?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions#anonymous_functions_and_arrow_functions)
[What is function scope?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Functions#function_scope_and_conflicts)
[What are return values?](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Return_values#what_are_return_values)
[What are arrow functions?](https://javascript.info/arrow-functions-basics)
While these questions may seem straightforward, They help make learning more interactive by getting learners "us" to explain what they "we" know. This can help them "us" understand and remember the material better.

I hope this clears up any confusion. 

Thank you again for your time!

foundations/javascript_basics/function_basics.md
youssef-el-atmani: "Arrow functions are a type of function expression"
I thought they were different, my bad!

I’m not sure why you have found the second question about `the difference between a function declaration and a function expression` redundant. There is one big difference that led me to add this question: a function declaration gets hoisted and can be accessed before its declaration line. On the other hand, if you try to access a function expression before its declaration, you will get an error. I thought adding a question about this would help it stick in learners' minds.

I totally agree that it might seem trivial, but it’s just a question about something we’ve just learned. For example, look at the other existing questions in [knowledge-check](https://www.theodinproject.com/lessons/foundations-function-basics#knowledge-check):

- What are functions useful for?
- How do you invoke a function?

These are straightforward and simple questions, but I think they are important and serve their purpose.

MaoShizhong: I think you've misunderstood which question I referred to as redundant. The one I highlighted for deletion ("What is a function expression?") is somewhat redundant since arrow functions are also a type of function expression, and I'm not sure the question adds a great deal. The second question you added is perfectly fine IMO.

On top of this, what Zach pointed out about the PR form is correct - you need to explain why your made the proposed changes as opposed to just making them and saying what they are.

youssef-el-atmani: Thanks @MaoShizhong, next time I'll be sure to fill all the sections of the form.

MaoShizhong:
@zachmmeyer I'd personally be happy with approving the PR once the first KC question I highlighted above is removed, unless you feel differently? I'll defer to your judgement on that

zachmmeyer:
I do apologize for misunderstanding the original intent of this PR. Thank you for filling out the required sections and elaborated further.
I'm going to defer to @MaoShizhong. Thank you!

foundations/javascript_basics/function_basics.md
MaoShizhong: @youssef-el-atmani To clarify in case you misunderstood - I'm requesting that the first knowledge check question you added is removed. We only need the second question you added. Once the first one is removed, I'm happy to approve and merge this PR.

youssef-el-atmani: Sorry @MaoShizhong , I did get that the first  question should be removed, but I didn't get that I am the one who should remove it!
**Done**: First question has been removed

Thank you a lot for the clarification, and sorry because this PR toked way more time than it needed!

MaoShizhong (approved):
Thank you for contributing!

PR #28733, opened 2024-09-02
Working with APIs: Clarify default 'cors' mode behavior in CORS section

Maddily:
## Because

I wanted to mention the default behavior when sending cross-origin requests using `fetch`.

## This PR

- Mentions that, in case the request being sent is cross-origin, `mode`: 'cors'` is set by default.
- Adds that explicitly specifying `mode: 'cors'` in the options is good for clarity.

## Issue

Closes #XXXXX

## Additional Information

N/A

## Pull Request Requirements

### DIFF
--- a/javascript/asynchronous_javascript_and_apis/working_with_apis.md
+++ b/javascript/asynchronous_javascript_and_apis/working_with_apis.md
@@ -132,7 +132,7 @@ fetch('url.url.com/api', {
 });
 ```
 
-Adding the `{mode: 'cors'}` after the URL, as shown above, will solve our problems for now. In the future, however, you may want to look further into the implications of this restriction.
+Adding the `{mode: 'cors'}` after the URL, as shown above, will solve our problems for now. Note that `mode` is set to `'cors'` by default if the request is cross-origin. Explicitly specifying `mode: 'cors'` can still be useful for clarity and ensuring that the intended behavior is explicitly defined. In the future, however, you may want to look further into the implications of this restriction.
 
 ### Let's do this
 


MaoShizhong:
We can actually probably just remove the `{ mode: 'cors' }` bits entirely (from this lesson and a few other lessons in the curriculum) as that's the default mode for `fetch`. So the CORS section introducing that to "solve a problem" is redundant because there is no problem when omitted.

If you can put this on hold for a bit, I'm discussing with the team regarding some other CORS material because there might be an additional thing we need to do on top of removing the mode options references. I'll get back to you once we've discussed that through.

Maddily:
That makes sense. I'll put this on hold for now and wait for your update after the team discussion.

Maddily:
That's strange because, in the lesson, I tried removing the `mode: 'cors'` option and the request was still a success.

MaoShizhong:
@wise-king-sullyman Did you ever get round to confirming this in Safari on MacOS? And discussing how CORS can be introduced in the Rails pathway?

wise-king-sullyman:
My apologies, this slipped through the cracks on me. I just tested in Safari and confirmed that fetch is defaulting to cors mode so I think we should be ok to remove it as @mao-sz suggested.

Would you still be interested in making this change @Maddily? I know it has been quite a while since you opened this PR.

Maddily:
@wise-king-sullyman Absolutely. I'm on it.

Maddily:
@mao-sz I've removed the cors references from this lesson as well as _Fetching Data In React_ and _Async and Await_. Could you confirm if these were all the lessons that mentioned `mode: 'cors'`, or if I missed any?

PR #28746, opened 2024-09-05
ES6 Modules: Add warning about opening directly in browser

MaoShizhong:
## Because

ES6 modules require a dev server to be opened (as opposed to opening directly in the browser under `file://`) due to CORS. I had assumed we directly assigned people to use the Live Preview extension in VSC in an earlier lesson but they're actually just tips, so not everyone would have been using it ([user on Discord confused by CORS error due to opening via terminal](https://discord.com/channels/505093832157691914/513125308757442562/1281357219656630303)).

A quick note can be added that live server is necessary for now. Later, webpack-dev-server will be introduced.

## This PR

- Adds warning box for opening ES6 modules, advising use of Live Preview instead

## Issue

N/A

## Additional Information

Changed from Live Server to Live Preview due to #28801 

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/es6_modules.md
+++ b/javascript/organizing_your_javascript_code/es6_modules.md
@@ -214,7 +214,7 @@ two.js <-------------- one.js <-------------- three.js
 
 Note that we only needed the one script tag, as the browser will handle the additional file dependencies for us. We also did not need to add the `defer` attribute, as `type="module"` will automatically defer script execution for us.
 
-If you had coded along with the IIFE example at the start of the lesson, try rewriting the JavaScript to use `import` and `export`, and link only the entry point as a module script. Due to browser security reasons, ES6 modules cannot be loaded if you open the HTML file directly in the browser, so make sure you use Visual Studio Code's [Live Server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) if you aren't already.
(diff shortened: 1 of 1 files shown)


damon314159:
I did a quick site wide Google for cors and didn't see any mentions of cors this early in the course. If that is the case, is it wise to introduce it in a note box with no further explanation? 

That feels like it would cause a lot of questions in discord, or just unnecessary rabbit holes into cors and headers and such.

MaoShizhong:
> I did a quick site wide Google for cors and didn't see any mentions of cors this early in the course. If that is the case, is it wise to introduce it in a note box with no further explanation?
> 
> That feels like it would cause a lot of questions in discord, or just unnecessary rabbit holes into cors and headers and such.

The note is less about CORS itself and more that to run the ESM code you actually need a server, because if you try to open the HTML file in the browser directly (as some might still be doing), they will actually get a CORS error.

That being said, looking at this again, it probably doesn't need a note box but rather just adding to the end of the previous sentence:

> If you had coded along with the IIFE example at the start of the lesson, try rewriting the JavaScript to use `import` and `export`, and link only the entry point as a module script. Due to browser security reasons, ES6 modules cannot be loaded if you open the HTML file directly in the browser, so make sure you use Visual Studio Code's [Live Server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer) if you aren't already.

What do you think?

CouchofTomato (approved):

PR #28782, opened 2024-09-11
feat(mdlint): improve error message for TOP005

nikitarevenco:
Note: Making this PR to test if this will even work, only review if it's *not* marked as draft

image

### DIFF
--- a/markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
+++ b/markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
@@ -1,10 +1,11 @@
 module.exports = {
   names: ["TOP005", "blanks-around-multiline-html-tags"],
-  description: "Multiline HTML tags should be surrounded by blank lines or code block delimiters",
+  description:
+    "Multiline HTML tags should be surrounded by blank lines or code block delimiters",
   tags: ["html", "blank_lines"],
   parser: "markdownit",
   information: new URL(
-    "https://github.com/TheOdinProject/curriculum/blob/main/markdownlint/docs/TOP005.md"
+    "https://github.com/TheOdinProject/curriculum/blob/main/markdownlint/docs/TOP005.md",
   ),
   function: function TOP005(params, onError) {
     /**
@@ -12,16 +13,25 @@ module.exports = {
      * We only want to flag HTML tags we use for actual markup,
      * or md code block examples of such.
      */
-    const IGNORED_FENCE_TYPES = ["html", "jsx", "erb", "ejs", "ruby", "javascript"];
+    const IGNORED_FENCE_TYPES = [
+      "html",
+      "jsx",
+      "erb",
+      "ejs",
+      "ruby",
+      "javascript",
+    ];
     const ignoredFencesLineRanges = params.parsers.markdownit.tokens
       .filter((token) => {
-        return token.type === "fence" && IGNORED_FENCE_TYPES.includes(token.info);
+        return (
+          token.type === "fence" && IGNORED_FENCE_TYPES.includes(token.info)
+        );
       })
       .map((token) => token.map);
 
     const isWithinIgnoredFence = (lineNumber) => {
       return ignoredFencesLineRanges.some(
-        (range) => range[0] < lineNumber && lineNumber < range[1]
+        (range) => range[0] < lineNumber && lineNumber < range[1],
       );
     };
 
@@ -33,7 +43,7 @@ module.exports = {
         }
         return lineNumbers;
       },
-      []
+      [],
     );
 
     isolatedHtmlTagsLineNumbers.forEach((lineNumber, i) => {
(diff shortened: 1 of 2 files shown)


MaoShizhong:
A heads up if you just want to test something like this first, you can PR in your own fork, as opposed to testing out in the upstream repo. You can then open a new PR from that branch to upstream main if you decide it's appropriate.

nikitarevenco:
> A heads up if you just want to test or experiment with something like this first, please PR in your own fork, as opposed to testing out in the upstream repo. You can then open a new PR from that branch to upstream main if you decide it's appropriate.

thanks for the tip!

It's ready for review now

nikitarevenco:
image

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
```diff
@@ -63,15 +63,20 @@ module.exports = {
         replacementText = `${replacementText}\n`;
       }
 
+      const ERROR_MESSAGE_BASE = "Expected a newline (\\n) or a code block delimiter (```)";
+
+      const detail = ERROR_MESSAGE_BASE
+        .concat(lineBeforeIsValid ? "" : " before the tag")
+        .concat(lineBeforeIsValid || lineAfterIsValid ? "" : " and")
+        .concat(lineAfterIsValid ? "" : " after the tag");
+
```
MaoShizhong, line 72: Nit - given that this bit follows the same general logic as the bit above in terms of conditionally altering the replacementText, for constistency, it should be written in the same sort of format.

i.e. instead of having two very similar bits of logic handled via completely different means, they ought to be handled the same way, in this case, `let` variable and conditionals with template strings, rather than suddenly using concat methods and ternaries.

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
```diff
@@ -63,15 +63,20 @@ module.exports = {
...
        * lineNumber is params.lines index (0-indexed).
        * +1 required as file line numbers are 1-indexed.
        */
       onError({
-        lineNumber: lineNumber + 1,
-        detail: `\n  Expected blank lines/code block delimiters: { Before: 1, After: 1 }\n  Actual blank lines/code block delimiters: { Before: ${
-          lineBeforeIsValid ? 1 : 0
-        }, After: ${lineAfterIsValid ? 1 : 0} }\n`,
+        lineNumber: lineNumber + 1,      
+        detail,
```
MaoShizhong, line 79: Shorthand is nice but again for consistency, ideally use all shorthand if possible, or none at all. Avoid mixing.
```suggestion
        detail: detail,
```

MaoShizhong (changes requested):

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
```diff
@@ -63,15 +63,20 @@ module.exports = {
         replacementText = `${replacementText}\n`;
       }
 
+      const ERROR_MESSAGE_BASE = "Expected a newline (\\n) or a code block delimiter (```)";
```
MaoShizhong, line 66: "Blank line" is clearer than "newline (\n)" since people might actually try to put "\n" on the line before/after.
```suggestion
      const ERROR_MESSAGE_BASE = "Expected a blank line or a code block delimiter (```)";
```

nikitarevenco: Ok, that's true!

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
nikitarevenco: I personally prefer no-mutation when I can, but I agree with you that consistency is definitely important! I have refactored it

MaoShizhong: The specific approach actually used doesn't matter so much so long as it's within reason. More importantly it's consistency else anyone else working on this file can get more confused when two things that do essentially the exact same thing are written in two completely different ways.

You might want to test the latest code though :P The `const` looks suspicious.

nikitarevenco:
Tests aren't passing so I'll let you know when I have fixed it

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
nikitarevenco: Oh yea oopsie! I fixed it!

nikitarevenco:
Ok so I'll use my computer to edit the code and test it instead of relying on github's code editor, this approach only works well for documents

nikitarevenco:
I fixed everything

MaoShizhong (changes requested):
Nit again sorry - can the strings be handled via template literals instead of concat/increment? Same reasoning as before - consistency of approach (in file and between files).

nikitarevenco:
> Nit again sorry - can the strings be handled via template literals instead of concat/increment? Same reasoning as before - consistency of approach (in file and between files).

Of course!

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
```diff
@@ -68,10 +68,8 @@ module.exports = {
        * +1 required as file line numbers are 1-indexed.
        */
       onError({
-        lineNumber: lineNumber + 1,
-        detail: `\n  Expected blank lines/code block delimiters: { Before: 1, After: 1 }\n  Actual blank lines/code block delimiters: { Before: ${
-          lineBeforeIsValid ? 1 : 0
-        }, After: ${lineAfterIsValid ? 1 : 0} }\n`,
+        lineNumber: lineNumber + 1,      
+        detail: `Expected a blank line or a code block delimiter (\`\`\`)${lineBeforeIsValid ? "" : " before the tag"}${lineBeforeIsValid || lineAfterIsValid ? "" : " and"}${lineAfterIsValid ? "" : " after the tag"}`,
```
MaoShizhong, line 72: This isn't quite what I meant. What you had before with the if conditionals was fine, it was just that the string formation was handled via concatenation and not template literals, which is inconsistent with how the same logic right before it handled things.

What I meant was:
```js
      if (!lineBeforeIsValid) {
        errorMessage = `${errorMessage} before the tag`;
      }
      if (!lineBeforeIsValid && !lineAfterIsValid) {
        errorMessage = `${errorMessage} and`;
      }
      if (!lineAfterIsValid) {
        errorMessage = `${errorMessage} after the tag`;
      }
```

Which follows the exact same approach as the replacementText bit before. Else you'd have the same logic using template literals to form a string then suddenly concatenation, when they should be the same.

MaoShizhong (changes requested):

markdownlint/TOP005_blanksAroundMultilineHtmlTags/TOP005_blanksAroundMultilineHtmlTags.js
nikitarevenco: > This isn't quite what I meant. What you had before with the if conditionals was fine, it was just that the string formation was handled via concatenation and not template literals, which is inconsistent with how the same logic right before it handled things.
> 
> What I meant was:
> 
> ```js
>       if (!lineBeforeIsValid) {
>         errorMessage = `${errorMessage} before the tag`;
>       }
>       if (!lineBeforeIsValid && !lineAfterIsValid) {
>         errorMessage = `${errorMessage} and`;
>       }
>       if (!lineAfterIsValid) {
>         errorMessage = `${errorMessage} after the tag`;
>       }
> ```
> 
> Which follows the exact same approach as the replacementText bit before. Else you'd have the same logic using template literals to form a string then suddenly concatenation, when they should be the same.

Ah :laughing: , this makes more sense now. Thank you for clarifying!

MaoShizhong (approved):

PR #28818, opened 2024-09-20
Ruby: Project Management: Change method name in description

ruben-vl:
## Because
The description refers to the above-mentioned method as 'foo', while the example uses a method called 'food_opinion'. Probably wasn't changed when changing the example to something more descriptive.

## This PR
- Fixes the method name in the description

## Pull Request Requirements
-   [X] If this PR addresses an open issue, it is linked in the `Issue` section


CouchofTomato (approved):

MaoShizhong:
> The MarkDownLint check fails, but I cannot figure out what causes it. I checked the preview tool and the style guide, but cannot find issues there. My assumption is that the underscore creates a problem, but I don't see any details on why the check failed.

Looks like there's an error in how the lint actions were filtering paths (it used the wrong workflow which then led to it running the linter with no file path, hence the output showing the `--help` display instead. I'll see if I can get a fix in soon.

PR #28828, opened 2024-09-21
Ruby: Linting and RuboCop: Add minor clarifications and improvements

ruben-vl:
## Because
Just some minor changes to improve consistency (e.g. `touch` or `nano`  while the `rubocop --init` was introduced in a previous paragraph).

Also some punctuation and spacing to improve readability and limit confusion.

## This PR
- adds consistency with earlier paragraph by adding the `rubocop --init` command
- rewrites a few sentences in a minor way to avoid confusion and improve readability

## Pull Request Requirements
-   [X] If this PR addresses an open issue, it is linked in the `Issue` section

### DIFF
--- a/ruby/object_oriented_programming_basics/linting_and_rubocop.md
+++ b/ruby/object_oriented_programming_basics/linting_and_rubocop.md
@@ -226,7 +226,7 @@ This will disable the `AbcSize` Cop from `Metrics` department between those comm
 
 Some rules are a lot more arbitrary - the Style department is going to be the prime ground for strong arguments about things that don't really matter - like double-quoting all strings vs making a distinction between plain strings and string interpolation. Perhaps you have strong feelings about quotes, so let's help you out by showing you how to show them to RuboCop.
 
-Start by creating a `.rubocop.yml` file using the command touch or nano(nano will open the text editor right away). Don't forget that it must be a dotfile, meaning it needs to have a dot before its name. Now, you need to find out what rule you want to change or disable. For the possible options always consult the documentation - not every Cop is just a simple on/off, there might be more options. As an example, we'll be changing the rules regarding strings, frozen string literals and we'll enable NewCops.
+Start by creating a `.rubocop.yml` file using the command `bundle exec rubocop --init` or manually using touch or nano (nano will open the text editor right away). Don't forget that it must be a dotfile, meaning it needs to have a dot before its name. Now, you need to find out what rule you want to change or disable. For the possible options always consult the documentation - not every Cop is just a simple on/off, there might be more options. As an example, we'll be changing the rules regarding strings, frozen string literals and we'll enable NewCops.
 
 ```yaml
 # This is .rubocop.yml in ~/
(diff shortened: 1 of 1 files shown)


scheals:
Originally I absent-mindedly told people to run `bundle exec` in `~` which should not work, as no Gemfile should be present in the home directory. What probably would be the simplest change is dropping `bundle exec` and running `rubocop --init` instead but a different one was made: https://github.com/TheOdinProject/curriculum/commit/d7cc9578668656b2ec3dee7514dcc66c5cf788d8 Because of that, the reference to running it in the home directory was lost, together with a mention of there being an auto-generated comment. It seems that the home directory being referenced heavily in the section made sure folks created the config file there but it does feel confusing at the moment.

Wonder if the comment that gets generated automatically when running the command needs to be added to the code block or not:
```rb
# The behavior of RuboCop can be controlled via the .rubocop.yml
# configuration file. It makes it possible to enable/disable
# certain cops (checks) and to alter their behavior if they accept
# any parameters. The file can be placed either in your home
# directory or in some project directory.
#
# RuboCop will start looking for the configuration file in the directory
# where the inspected file is and continue its way up to the root directory.
#
# See https://docs.rubocop.org/rubocop/configuration
```

ruben-vl:
> Originally I absent-mindedly told people to run `bundle exec` in `~` which should not work, as no Gemfile should be present in the home directory. What probably would be the simplest change is dropping `bundle exec` and running `rubocop --init` instead but a different one was made: [d7cc957](https://github.com/TheOdinProject/curriculum/commit/d7cc9578668656b2ec3dee7514dcc66c5cf788d8) Because of that, the reference to running it in the home directory was lost, together with a mention of there being an auto-generated comment. It seems that the home directory being referenced heavily in the section made sure folks created the config file there but it does feel confusing at the moment.

Interesting! Thanks for linking the original change. I see the confusing thing about the home folder `~` in the following paragraph, since that kind of comes out of nowhere now in the example and the paragraph after the code snippet. Maybe we should rethink that whole section, to make the context in which the different commands need to be run, clearer.

> Wonder if the comment that gets generated automatically when running the command needs to be added to the code block or not:
> 
> ```ruby
> # The behavior of RuboCop can be controlled via the .rubocop.yml
> # configuration file. It makes it possible to enable/disable
> # certain cops (checks) and to alter their behavior if they accept
> # any parameters. The file can be placed either in your home
> # directory or in some project directory.
> #
> # RuboCop will start looking for the configuration file in the directory
> # where the inspected file is and continue its way up to the root directory.
> #
> # See https://docs.rubocop.org/rubocop/configuration
> ```

It is mentioned shortly in the [section before it](https://www.theodinproject.com/lessons/ruby-linting-and-rubocop#you-are-the-code-dictator).

> To create the .rubocop.yml file, run this command from your terminal in your project’s directory:
> `bundle exec rubocop --init`
> It won’t have anything in it besides `a comment describing what it is for`.

It was clear to me when I was following the section that it should be run in the project folder instead of the home folder when using it with bundle. And I don't think adding the whole auto-generated text to the lesson explicitly adds much.

CouchofTomato:
If a rethink of that section is necessary, do you want to consider it in the context or this PR, or do you want this to be assessed as is and another issue can be made for that?

ruben-vl:
> If a rethink of that section is necessary, do you want to consider it in the context or this PR, or do you want this to be assessed as is and another issue can be made for that?

I think the small adjustments made in this PR are still relevant

JoshDevHub (approved):
This stuff seems good to me as is. If there's more changes anyone wants to recommend around command context in the home directory, feel free to open another issue or PR about it.

PR #28854, opened 2024-09-24
HashMap Data Structure lesson: fixed explanation on sets

### LINKED ISSUE #28853: HashMap Data Structure: Set contains only values, not keys
### Checks

### Describe your suggestion

There's a sentence in the **Buckets** section that says: `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only keys and no values.`

The correct statement should be: `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only values and no keys.`

Basically, a set contains only **values**, not keys.

References:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
- https://javascript.info/map-set

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-hashmap-data-structure

### (Optional) Discord Name

_No response_

_No response_

Prin-S:
Sets contain only values and not keys.

## Because
In the lesson on HashMap Data Structure for both JS and Ruby paths, the following is stated: `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only keys and no values.`

This is not correct because a set contains only values, not keys.

## This PR
- In `curriculum/javascript/computer_science/hash_map_data_structure.md`, `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only keys and no values.` is changed to `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only values and no keys.`.
- In `curriculum/ruby/computer_science/hash_map_data_structure.md`, `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only keys and no values.` is changed to `A Set is similar to a hash map but the key difference (pun intended) is that a Set will have nodes with only values and no keys.`.

## Issue
Closes #28853

## Additional Information
N/A

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/hash_map_data_structure.md
+++ b/javascript/computer_science/hash_map_data_structure.md
@@ -92,7 +92,7 @@ Buckets are storage that we need to store our elements. Simply, it's an array. F
 1. Find the bucket at index `385`.
 1. Store the key value pair in that bucket. In this case, the key would be "Fred" and the value would be "Smith".
 
-What if the bucket at index `385` already contains an item with the same key "Fred"? We check if it's the same item by comparing the keys, then we overwrite the value with our new value. This is how we can only have unique values inside a `Set`. A `Set` is similar to a hash map but the key difference (pun intended) is that a `Set` will have nodes with only keys and no values.
+What if the bucket at index `385` already contains an item with the same key "Fred"? We check if it's the same item by comparing the keys, then we overwrite the value with our new value. This is how we can only have unique values inside a `Set`. A `Set` is similar to a hash map but the key difference (pun intended) is that a `Set` will have nodes with only values and no keys.
 
 This is an oversimplified explanation; we'll discuss more internal mechanics later in the lesson.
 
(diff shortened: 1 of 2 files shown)


damon314159:
It is more correct to say that a hashset contains keys, than to say it contains values, because the things you store in a hashset get hashed, and lookups are performed against them. That is what keys are. Values, by contrast, are not hashed and lookups are not performed against them

JoshDevHub:
I agree with what @damon314159 has said here. From the perspective of creating a `HashSet`, you want to think of the things you're inserting as "keys." This is because they occupy a role similar to the `HashMap` keys -- they are what's getting hashed and duplicate values cannot exist.

If we just change the wording to say it contains only values, then learners may miss what exactly is happening.

As an aside, in the future you should wait to be assigned an issue before working on it. This gives the team the ability to talk over the changes before you spend any extra time working on it.

PR #28886, opened 2024-10-01
Added additional resources for Webpack

WadieZN:
There were no videos in the Webpack course, so I added a playlist that helped me personally to understand the concepts.

### DIFF
--- a/javascript/organizing_your_javascript_code/webpack.md
+++ b/javascript/organizing_your_javascript_code/webpack.md
@@ -446,4 +446,4 @@ The following questions are an opportunity to reflect on key topics in this less
 
 This section contains helpful links to related content. It isn't required, so consider it supplemental.
 
-- It looks like this lesson doesn't have any additional resources yet. Help us expand this section by contributing to our curriculum.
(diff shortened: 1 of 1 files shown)


MaoShizhong:
Thanks for the PR @WadieZN.

I'm going to close this PR without merging for the following reasons:

- The content in the proposed videos does not entirely match the content from the lesson. It includes more than the lesson intends to expose, and also includes things that are no longer relevant for the current Webpack version (and so teaches things differently from our lesson).
- One of the biggest past problems with this lesson was that the assignments were handed off to multiple external guides which differed from each other and also introduced things that were out of scope of the lesson without sufficient explanation, and/or did not include things we actually wanted explaining. That's why the whole lesson was rewritten with full in-house content so we only teach what we feel is in scope for the curriculum and nothing more. Linking to another set of lessons that teaches essentially very different things in a very different way, that'll be more likely to cause confusion than help resolve it.
- We don't *need* to link to alternative forms of learning material just for the sake of variety. We also can't link to every alternative resource out there. People will naturally need to do their own research and if that includes finding other resources that help them understand things, then that's awesome. But everyone will find different things more helpful or less helpful.

Hope that makes sense as to why the resource has been declined.

PR #28915, opened 2024-10-04
Time Complexity: Update the worst-case scenario example

### LINKED ISSUE #28903: JavaScript: Clarify the worst-case scenario example in Big Ω (Omega notation) section
### Checks

### Describe your suggestion

I think that the explanation of the worst-case scenario under the **Big Ω (Omega notation)** section in the **Time Complexity** lesson is inaccurate/incomplete.

The lesson describes it as follows:
> In the worst case (Big O), which would happen if the item is not in the array, we would say it had linear complexity O(N). This is because the item we are looking for is not in the array, so our code must iterate on every value.

However, this explanation overlooks another important worst-case scenario: what if the element we're looking for is the last one in the array?

I believe the worst-case scenario can occur in two situations:
- **Item not found**: Same as the lesson explains, the algorithm must check each element, resulting in O(n) complexity.
- **Item found at the last position:** In this case, we would also check each element, including the last element, which again results in a time complexity of O(n).

Clarifying both scenarios is essential for a complete understanding of time complexity. Let me know what you think.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-time-complexity

### (Optional) Discord Name

Maddie

_No response_

Maddily:
## Because

The previous explanation overlooked the worst-case scenario when the element we're looking for is the last one in the array.

## This PR

- Updates the worst-case explanation to make sure it covers both possible cases: when the item isn't found and when the item is the last one in the array.

## Issue

Closes #28903

## Additional Information

N/A

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/time_complexity.md
+++ b/javascript/computer_science/time_complexity.md
@@ -246,7 +246,7 @@ function findValue(arr) {
 }
 ```
 
-In the worst case (Big O), which would happen if the item is not in the array, we would say it had linear complexity `O(N)`. This is because the `item` we are looking for is not in the array, so our code must iterate on every value. If the array input doubles in size then the worst case also means our method must double the number of iterations looking for the `item`.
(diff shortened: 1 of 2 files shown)


JoshDevHub (approved):
Thank you for contributing.

Issue #27682, opened 2024-03-25
HashMap Data Structure: more clarity over need to rehash existing keys when growing a hash map


gtpWebDev:
### Checks

- [X] This is not a duplicate of an existing issue (please have a look through our [open issues list](https://github.com/TheOdinProject/curriculum/issues) to make sure)
- [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [ ] Would you like to work on this issue?

### Describe your suggestion

As I understand it, when growing a hash map, it is necessary to rehash all existing keys. The course currently says "To grow our buckets, we create a new buckets list that is double the size of the old buckets list, then we copy all nodes over to the new buckets." That wasn't very clear to me, and skimming solutions from students on the hashmap project, 2 out of 3 just mapped the existing keys directly to the larger hash map without rehashing them. It would certainly benefit from more explanation.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/javascript-hashmap-data-structure

### (Optional) Discord Name

glenoftzintel

### (Optional) Additional Comments

Enjoying the course, intending to start to provide a bit more feedback going forward!

JoshDevHub:
I agree this is a good idea, and thank you for making this issue.

Might be something we handle as part of #27174, but I'll keep this open in case we don't.

goran1010:
Since the other issue https://github.com/TheOdinProject/curriculum/issues/27174 is on hold, I'll assign myself to do the small change needed here, to make it clear what needs to happen to keys when we grow the hash map.

goran1010:
Re-reading the lesson again, I realize that the work the issue brought up originally is no longer needed, and I'm going to close this issue, since PR https://github.com/TheOdinProject/curriculum/pull/29008 has already addressed this by changing the sentence in question to:

> To grow our array, we create a new one that is double its size and then copy all existing nodes over to the buckets of this new array, _hashing their keys again_.

PR #29013, opened 2024-10-28
Project HashMap: Improve requirement clarity

cyphersept:
## Because

- This PR aims to improve clarity of the assignment instructions for Project: HashMap by decreasing wordiness and making implicit requirements explicit, so that students can more easily understand the complex project standards

## This PR

- Added variables for load factor and capacity to explicit requirements
- Clarified expected behaviour for bucket growth
- Added link to previous lesson's section on collision handling referenced in the assignment
- Clarified expected outcomes in testing section
- Made terminology more consistent with itself and with previous lesson: "capacity" is now used only to describe the number of buckets in the hash map, "load levels" and "fullness" used only to describe what fraction of the HashMap's capacity is actually filled, and "load factor" is used for the maximum allowed load level before the buckets are grown.
- Some verbose sentences rewritten 
- Some vague descriptions expanded
- Verbal "fluff" removed in lengthy sections
- Minor grammatical fixes

## Issue

Related to https://github.com/TheOdinProject/curriculum/issues/27682

## Additional Information

Discord username: @phosphorflux

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_hash_map.md
+++ b/javascript/computer_science/project_hash_map.md
@@ -51,7 +51,7 @@ if (index < 0 || index >= buckets.length) {
 
       Recall that collisions occur when *TWO DIFFERENT* keys generate the same hash code and get assigned to the same bucket. (e.g. `Carlos` and `Carla` are both hashed to `3`, so `3` becomes a location for `Carlos` AND `Carla`. However, we know that this is not an update because the keys are different). Review the [dealing with collisions](https://www.theodinproject.com/lessons/javascript-hashmap-data-structure#collisions) section of the previous lesson to find a way to handle our collisions.
 
-      - Remember to grow your buckets when your hash map reaches the `load factor` by copying its contents to a new hash map with double the previous capacity. The methods mentioned later in this assignment can help you handle the growth logic, so you may want to implement this feature near the end. However, we mention this with `set()` because it's important to grow buckets exactly as they are being expanded.
+      - Remember to grow your buckets to double their capacity when your hash map reaches the `load factor`. The methods mentioned later in this assignment can help you handle the growth logic, so you may want to implement this feature near the end. However, we mention this with `set()` because it's important to grow buckets exactly as they are being expanded.
 
   1. `get(key)` takes one argument as a key and returns the value that is assigned to this key. If a key is not found, return `null`.
 
--- a/javascript/computer_science/project_hash_map.md
+++ b/javascript/computer_science/project_hash_map.md
@@ -1,6 +1,6 @@
 ### Introduction
 
-You already know the magic behind hash maps; now it's time to write your own implementation!
+You already know the magic behind hash maps. Now it's time to write your own implementation!
 
 #### Limitation
 
--- a/javascript/computer_science/project_hash_map.md
+++ b/javascript/computer_science/project_hash_map.md
@@ -1,16 +1,16 @@
 ### Introduction
 
-You already know the magic behind hash maps, now it's time to write your own implementation!
+You already know the magic behind hash maps; now it's time to write your own implementation!
 
 #### Limitation
 
-  Before we get started, we need to lay down some ground rules. JavaScript's dynamic nature of array allows us to insert and retrieve indexes that are outside our array size range. Example: if we create an array of size `16` to be our buckets size, nothing stopping us from storing items at index `500`. This defeats the purpose we are trying to demonstrate, so we need to put some self restriction to work around this.
+  Before we get started, we need to lay down some ground rules. JavaScript's dynamic nature of arrays allows us to insert and retrieve indexes that are outside our array size range. Example: if we create an array of size `16` to represent our buckets, nothing stops us from storing items at index `500`. This defeats the purpose of limiting storage size in hash maps, so we need to enforce some restrictions.
 
-  Use the following snippet whenever you access a bucket through an index. We want to throw an error if we try to access an out of bound index:
+  Use the following snippet whenever you access a bucket through an index. We want to throw an error if we try to access an out-of-bounds index:
 
 ```javascript
 if (index < 0 || index >= buckets.length) {
-  throw new Error("Trying to access index out of bound");
+  throw new Error("Trying to access index out of bounds");
 }
 ```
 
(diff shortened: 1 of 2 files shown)


javascript/computer_science/project_hash_map.md
```diff
@@ -1,26 +1,26 @@
 ### Introduction
 
-You already know the magic behind hash maps, now it's time to write your own implementation!
+You already know the magic behind hash maps; now it's time to write your own implementation!
```
JoshDevHub, line 3:
```suggestion
You already know the magic behind hash maps. Now it's time to write your own implementation!
```
Would rather these just be two separate sentences I think.

javascript/computer_science/project_hash_map.md
```diff
@@ -35,23 +35,23 @@ if (index < 0 || index >= buckets.length) {
...
 
-      - Remember to grow your buckets size when it needs to, by calculating if your bucket has reached the `load factor`. Some of the methods in this assignment that are mentioned later could be reused to help you handle that growth logic more easily. So you may want to hold onto implementing your growing functionality just for now. However, the reason why we mention it with `set()` is because it's important to grow buckets exactly when they are being expanded.
+      - Remember to grow your buckets when your hash map reaches the `load factor` by copying its contents to a new hash map with double the previous capacity. The methods mentioned later in this assignment can help you handle the growth logic, so you may want to implement this feature near the end. However, we mention this with `set()` because it's important to grow buckets exactly as they are being expanded.
```
JoshDevHub, line 54: > by copying its contents to a new hash map

A bit curious what you mean with this exactly? To me, this wording could be misinterpreted as requiring a new hash map instance, because `HashMap` is the name of the class/factory function being used.

JoshDevHub (changes requested):
Nice work. Looks mostly good, just have one small nit and one small thing to discuss before I approve/merge. Let me know what you think.

javascript/computer_science/project_hash_map.md
cyphersept: You're right, the wording's confusing. I was referencing instructions from the previous lesson because "doubling the bucket capacity" being the expected outcome after implementing growth logic should be mentioned explicitly as a requirement, instead of implicitly being found on another page. Maybe something like:
> Remember to grow your buckets to double capacity when your hash map reaches the `load factor`.
?

javascript/computer_science/project_hash_map.md
```diff
@@ -35,23 +35,23 @@ if (index < 0 || index >= buckets.length) {
...
+      You are free to use that, or you can conduct your own research on hashing algorithms. Beware, this is a deep, deep rabbit hole.
 
-      Also, there is one edge case with long keys that was not taken into consideration in the function or rather how we applied a modulo `%` operator. JavaScript is unable to hold large numbers precisely. At some point, calculations are going to be inaccurate, which significantly increases the chances of collisions. There are a few ways how we could handle it but we recommend that you apply the modulo operator on *each iteration* instead of outside the loop at the end. In that case, we prevent the output from becoming larger than our bucket's length.
+      However, there is one edge case our `hash` function still needs to address. For very long keys, our hash code will exceed the maximum integer value allowed by Javascript. Once that happens, calculations become inaccurate, and the chance of collisions significantly increases. One way to avoid this issue is to apply the modulo `%` operator on *each iteration* instead of outside the loop at the end. This ensures the output never becomes larger than our bucket's length.
```
JoshDevHub, line 40:
```suggestion
      However, there is one edge case our `hash` function still needs to address. For very long keys, our hash code will exceed the maximum integer value allowed by JavaScript. Once that happens, calculations become inaccurate, and the chance of collisions significantly increases. One way to avoid this issue is to apply the modulo `%` operator on *each iteration* instead of outside the loop at the end. This ensures the output never becomes larger than our bucket's length.
```

JoshDevHub (changes requested):
Nice work. Just one more small change to get the Markdown linter to pass, and then I'll be good to merge.

Another thing: the Ruby course has this exact same project in it just with example code written in the Ruby language. Would you be comfortable making a new PR that makes these updates there as well? You shouldn't have to know any Ruby to do this, and if you had any questions, I'd be glad to help.

cyphersept:
@JoshDevHub thanks for the catch, I fixed the linter issue. I went ahead and made the changes to the Ruby lesson on the newest commit after the lint fix because I think it makes more sense to keep the changes synced in the same PR. If you'd rather I open a new PR, please let me know.

ruby/computer_science/project_hash_map.md
```diff
@@ -1,12 +1,12 @@
...
 
 #### Limitation
 
-  Before we get started, we need to lay down some ground rules. Ruby's dynamic nature of array allows us to insert and retrieve indexes that are outside our array size range. Example: if we create an array of size `16` to be our buckets size, nothing is stopping us from storing items at index `500`. This defeats the purpose we are trying to demonstrate, so we need to put some self restriction to work around this.
+  Before we get started, we need to lay down some ground rules. Ruby's dynamic nature of arrays allows us to insert and retrieve indexes that are outside our array size range. Example: if we create an array of size `16` to represent our buckets, nothing stops us from storing items at index `500`. This defeats the purpose of limiting storage size in hash maps, so we need to enforce some restrictions.
 
-  Use the following snippet whenever you access a bucket through an index. We want to raise an error if we try to access an out of bound index:
+  Use the following snippet whenever you access a bucket through an index. We want to throw an error if we try to access an out-of-bounds index:
```
JoshDevHub, line 9: > We want to throw an error

Needs to go back to being "raise" here. It's the Ruby way to talk about errors.

JoshDevHub, line 9:
```suggestion
  Use the following snippet whenever you access a bucket through an index. We want to raise an error if we try to access an out-of-bounds index:
```
In Ruby we use this "raise" language when talking about having an error.

ruby/computer_science/project_hash_map.md
```diff
@@ -92,21 +92,21 @@ You already know the magic behind hash maps, now it's time to write your own imp
     test.set('lion', 'golden')
     ```
 
-1. After populating your hash map with the data above, your hash map's actual capacity should now be at `0.75` (full capacity).
+1. After populating your hash map with the data above, your hash map's current load levels should now be at `0.75` (full capacity).
 
-1. Now with a full hash map, try overwriting a few nodes using `#set(key, value)`. By right, this should only over-write the existing `values` of your nodes and not add new ones.
+1. Now with a full hash map, try overwriting a few nodes using `#set(key, value)`. This should only overwrite the existing `values` of your nodes and not add new ones, so `length()` should still return the same value and `capacity` should remain the same.
```
JoshDevHub, line 97:
```suggestion
1. Now with a full hash map, try overwriting a few nodes using `#set(key, value)`. This should only overwrite the existing `values` of your nodes and not add new ones, so `#length` should still return the same value and `capacity` should remain the same.
```
Small nit here to change the method reference styling to fit Ruby conventions.

JoshDevHub (changes requested):
Alright, looking good. And it's fine to have the Ruby stuff in this PR as well.

Just a couple small changes related to differences in Ruby and JS, but after those are good to go, I'll be glad to merge.

Thank you for making these changes. I think they make the project a lot better :1st_place_medal:

JoshDevHub (approved):
Looks great. Thank you so much for putting in the work on this.

PR #29021, opened 2024-10-29
Conditional Logic: Correct ternary operator syntax

youssef-el-atmani:
Clarified ternary operator syntax by replacing "conditional" with "logical" for accuracy.

## Because

To make the ruby ternary operator syntax more accurate

## This PR

- Replaced "conditional" with "logical" in the ternary operator syntax

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/ruby/basic_ruby/conditional_logic.md
+++ b/ruby/basic_ruby/conditional_logic.md
@@ -288,7 +288,7 @@ You should use an `unless` statement when you want to **not** do something if a 
 
 The ternary operator is a one-line `if...else` statement that can make your code much more concise.
 
-Its syntax is `conditional statement ? <execute if true> : <execute if false>`. You can assign the return value of the expression to a variable.
+Its syntax is `logical statement ? <execute if true> : <execute if false>`. You can assign the return value of the expression to a variable.
 
 ```ruby
 age = 19
--- a/ruby/basic_ruby/conditional_logic.md
+++ b/ruby/basic_ruby/conditional_logic.md
@@ -288,7 +288,7 @@ You should use an `unless` statement when you want to **not** do something if a 
 
 The ternary operator is a one-line `if...else` statement that can make your code much more concise.
 
-Its syntax is `logical statement ? <execute if true> : <execute if false>`. You can assign the return value of the expression to a variable.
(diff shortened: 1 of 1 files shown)


CouchofTomato:
Why do you feel conditional statement is not correct?

youssef-el-atmani:
Thank you @CouchofTomato for give me a chance to clarify my opinion
> Why do you feel conditional statement is not correct?

Because the right name for something that could evaluates to `true` or `false` is a `logical statement` not a `conditional statement`,
for example :
```ruby
#the followings are logical statements
5 == 7                #=> evaluates to `false`
6 >= 10               #=> evaluates to `false`
"hi" != "hello"       #=> evaluates to `true`


# I think the following is a conditional statement, which is composed of : `if ... end`
if logical_statement
  #do something 
else
  # do that another thing
end
```
I think `logical statement ? <execute if true> : <execute if false>` is as whole a conditional statement.

youssef-el-atmani:
@CouchofTomato May I ask you about what do you think?

CouchofTomato:
Apologies for the delay.

I don't disagree with you that it's a logical expression. However you'll often see it referred to as the condition in a lot of places because it decides the executed part of a conditional operator. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/conditional-operator
etc etc

So in isolation it definitely is a logical expression, but as part of a conditional operator it is often referred to as the condition. That's why I'm not sure it's wrong to leave it as it is.

What are your thoughts?

youssef-el-atmani:
Thanks a lot @CouchofTomato, I guess it should stay as it is, `conditional statement ? <execute if true> : <execute if false>`

There is a little change I recommend based on the resources you provided: `MDN and Microsoft`,
**MDN**: `condition ? exprIfTrue : exprIfFalse`
**Microsoft**: `condition ? consequent : alternative`

So I recommend replacing `conditional statement` by `condition`, because I guess it fits better
And It is up to you to decide whether that change is worth it or not

CouchofTomato (approved):

CouchofTomato:
Yeah I'm happy to accept that. Thanks for working through this with me and I'm glad we arrived at a satisfactory outcome.

PR #29049, opened 2024-11-04
JS Basics- Problem Solving: Fix incorrect interpretation

bhagyeshsp:
The current explanation tells that the loop continues until the value of `i` is greater than the value of `answer`. But in fact, the parameter condition is `i <= answer`. Hence, the correction.

## Because

The incorrect interpretation of `i <= answer` condition is fixed.

## This PR

- Replaced only a few words in the same line

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/foundations/javascript_basics/problem_solving.md
+++ b/foundations/javascript_basics/problem_solving.md
@@ -134,7 +134,7 @@ for (let i = 1; i <= answer; i++) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


MaoShizhong:
The current wording is correct. The condition part of a loop tells us the condition that the loop keeps going. So `i <= answer` means "keep looping while i is less than or equal to answer" which is the same as "loop until i is greater than answer".

bhagyeshsp:
Oops! Apologies for raising the false flag.
I will have to double-triple check before raising the PR in the future.
Thanks for the explanation.

PR #28974, opened 2024-10-19
Node revamp cookies

MaoShizhong:
WIP

## Because

Part of the Node Revamp milestone 2 - new lesson on Cookies.

## This PR

- Adds a new Cookies lesson to the Node.js course - Authentication section.

## Issue

Closes #28848

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project curriculum contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x] If this PR addresses an open issue, it is linked in the `Issue` section
-   [x] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [x] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

PR #29082, opened 2024-11-13
New lesson: Session-based authentication

### LINKED ISSUE #28847: New Lesson: Sessions
## Lesson overview

Lesson about sessions and implementing session-based auth.

## Lesson outline

The following items define the scope of the lesson. The lesson outline is flexible; you can improve, expand, and omit items whilst writing lesson content (make sure the pull request description mentions that).

- High level overview of sessions and session-based auth.
- Touch on the use of cookies in session-based auth but don't dive into details (separate lesson).
- Give simple example of manual implementation of session-based auth in a same-site context using `express-session`.
  - Also demonstrate using a DB session store e.g. (pg-connect-simple) for persistence.
- High level overview of password storage security and password hashing with argon2id.

## Acceptance criteria

If the requirements here are not met, the work effort is not complete.

- [ ] Lesson conforms to the outline above (unless outline has been modified).
- [ ] Lesson follows our [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md).

### Get the lesson template

Download the lesson template using the following command (replace `lesson_name` with the actual lesson name):

```bash
curl -o <lesson_name>.md https://raw.githubusercontent.com/TheOdinProject/curriculum/main/templates/lesson-template.md 
```

MaoShizhong:
## Because

As part of the Node revamp's 2nd milestone

## This PR

- Adds a new lesson for session-based auth to come after the cookies lesson (https://github.com/TheOdinProject/curriculum/pull/28974)

## Issue

Closes #28847

## Pull Request Requirements

### DIFF
--- a/nodeJS/authentication/sessions.md
+++ b/nodeJS/authentication/sessions.md
@@ -197,7 +197,7 @@ We only need to serialize the user ID as that will never change for a user (unli
 
 #### Warning: Use generic login error messages
 
-While it may seem like better UX to inform which field is incorrect when logging in with incorrect credentials, it is better to be vague. If you misspell the username but it happens to match a different user's username, you may not realize if the error message implies only the password is incorrect. It also makes it less likely that malicious people realise a username exists and they only have to guess the password.
+Don't specify which form fields are incorrect when providing validation feedback. Providing specific feedback can allow attackers to target accounts if they know a specific username exists, for example. It also means if you misspell your username but it happens to match someone else's username, you're less likely to be misled into thinking the you entered your username correctly.
 
 </div>
 
--- a/nodeJS/authentication/sessions.md
+++ b/nodeJS/authentication/sessions.md
@@ -43,45 +43,10 @@ CREATE TABLE users (
 Now we'll set up a minimal express app. We'll need the following dependencies installed first:
 
 ```bash
-npm install express express-session pg ejs
+npm install express express-session ejs dotenv pg connect-pg-simple
 ```
 
-Now our `app.js`:
-
-```javascript
-// app.js
-const path = require("node:path");
-const { Pool } = require("pg");
-const express = require("express");
-const session = require("express-session");
-
-const pool = new Pool({
-  // add your db configuration
-});
-
-const app = express();
-app.set("views", path.join(__dirname, "views"));
-app.set("view engine", "ejs");
-
-app.use(session({
-  secret: "we will explain and change this secret later",
-  resave: false,
-  saveUninitialized: false,
-  cookie: {
-    httpOnly: true,
-    maxAge: 2 * 24 * 60 * 60 * 1000, // 2 days
-  },
-}));
-app.use(express.urlencoded({ extended: false }));
-
-app.get("/", (req, res) => res.render("index"));
-
-app.listen(3000, () => {
-  console.log("App listening on port 3000!");
-});
-```
-
-And our initial `index.ejs` view:
+Now our initial `index.ejs` view:
 
 ```ejs
 <!-- views/index.ejs -->
@@ -96,6 +61,53 @@ And our initial `index.ejs` view:
 </body>
 </html>
 ```
+
+And our `app.js`:
+
+```javascript
+// app.js
+require("dotenv").config();
+const path = require("node:path");
+const { Pool } = require("pg");
+const express = require("express");
+const session = require("express-session");
+const pgSession = require("connect-pg-simple")(session);
+
+const pool = new Pool({
+  // add your db configuration
+});
+
+const app = express();
+app.set("views", path.join(__dirname, "views"));
+app.set("view engine", "ejs");
+
+app.use(session({
+  secret: process.env.SESSION_SECRET,
+  resave: false,
+  saveUninitialized: false,
+  cookie: {
+    httpOnly: true,
+    maxAge: 2 * 24 * 60 * 60 * 1000, // 2 days
+  },
+  store: new pgSession({ pool: pool }),
+}));
+app.use(express.urlencoded({ extended: false }));
+
+app.get("/", (req, res) => res.render("index"));
+
+const PORT = process.env.PORT || 3000;
+app.listen(PORT, () => {
+  console.log(`App listening on port ${PORT}!`);
+});
+```
+
+Let's talk about our `session` config, which we set for our whole app to use on every request. First, we set a **session secret**, which we set in our `.env` file since it's, well, a secret. This should be some string that's hard to guess - you can use a random key generator online for this - and is used by express-session alongside the session ID to generate a hash and sign the resulting session cookie. Then if an attacker tries to tamper with a session cookie, the signature would no longer match and the server can invalidate it.
+
+We also turn off `resave` and `saveUninitialized`, which means express-session will only save a session to our database or overwrite an existing one if it gets modified as part of the request. No need to save anything that hasn't been changed.
+
+We then pass in options for the session cookies that will be created. In the example, we make it inaccessible to JavaScript on the front-end and set a 2-day expiry. You can always use environment variables to set or conditionally set values (e.g. `httpOnly: process.env.NODE_ENV === "prod"` can allow you to access session cookies via front-end JavaScript in development but prevent it in production).
+
+Lastly, we use the [connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple) library to make express-session store session data in our database inside a new table. Without this, sessions would be stored in memory by default which would not persist through any server restarts!
 
 ### Creating users
 
--- a/nodeJS/authentication/sessions.md
+++ b/nodeJS/authentication/sessions.md
@@ -133,7 +133,7 @@ We will need a way for a user to create an account, so let's create a sign up fo
   <h1>Sign up</h1>
   <form action="/sign-up" method="POST">
     <label for="username">Username</label>
-    <input id="username" name="username" placeholder="username" type="text" />
+    <input id="username" name="username" type="text" />
     <label for="password">Password</label>
     <input id="password" name="password" type="password" />
     <button>Sign Up</button>
@@ -193,7 +193,9 @@ app.post("/login", async (req, res, next) => {
       req.session.userId = user.id;
       res.redirect("/");
     } else {
-      res.render("login", { error: "Incorrect username or password" });
+      res.render("login", {
+        error: "Incorrect username or password",
+      });
     }
   } catch(err) {
     next(err);
@@ -227,7 +229,7 @@ Let's create the login page:
   <h1>Please log in</h1>
   <form action="/login" method="POST">
     <label for="username">Username</label>
-    <input id="username" name="username" placeholder="username" type="text">
+    <input id="username" name="username" type="text">
     <label for="password">Password</label>
     <input id="password" name="password" type="password">
     <button>Log in</button>
@@ -274,7 +276,8 @@ async function checkAuthenticated(req, res, next) {
         "SELECT * FROM users WHERE id = $1",
         [req.session.userId],
       );
-      // add the user details we need to req so we can access it in the next middleware
+      // add the user details we need to req
+      // so we can access it in the next middleware
       req.user = {
         id: rows[0].id,
         username: rows[0].username,
--- a/nodeJS/authentication/sessions.md
+++ b/nodeJS/authentication/sessions.md
@@ -1,8 +1,8 @@
 ### Introduction
 
-Now that we've been introduced to cookies and their various uses and properties, let's use them to help us implement something. We want to allow someone to log in once and let the server "remember" them, automatically recognising any future requests from them.
+Now that we've been introduced to cookies and their various uses and properties, let's use them to help us implement something. We want to allow someone to log in once and let the server "remember" them, automatically recognizing any future requests from them.
 
-The basic login process (with a username and password) is rather straightforward. The user submits a form with their username and password, then the server checks the database if that username/password combo exists. If it does, great - it knows who the requester is and can continue with the rest of the request. If it does not exist, it does not know who the requester is so it can end the request there and then.
+The basic login process (with a username and password) is rather straightforward. The user submits a form with their username and password, then the server checks the database to see if that username/password combo exists. If it does, great - it knows who the requester is and can continue with the rest of the request. If it does not exist, it does not know who the requester is, so it can end the request there and then.
 
 So if someone does successfully "log in", how does the server recognize that the next request that user sends is coming from them? Without a system to persist the login, it'd just be a plain ol' request like any other and could have come from anyone. To handle this, we will use **sessions**.
 
@@ -20,13 +20,13 @@ This section contains a general overview of topics that you will learn in this l
 
 A session is just information about a user's interaction with the site in a given time period and can be used to store a whole variety of data. For persisting logins, we can store (serialize) some information about that user, such as their user ID, in a database table. That data will have its own ID and may also have an expiry time. We can then store that session's ID in a cookie (it doesn't need anything else stored in it) and send it back to the user in the server response.
 
-The client now has that cookie with the session ID and can then attach it to any future requests. The server can then check the database for a valid session with the same ID it found in the cookie. If there is a matching session, great - it can extract the serialized user information (deserialize) and continue with the request now it knows who made it. If there is no matching or valid session, like with logging in, we don't know who the user is so we can end the request there.
+The client now has that cookie with the session ID and can then attach it to any future requests. The server can then check the database for a valid session with the same ID it found in the cookie. If there is a matching session, great - it can extract the serialized user information (deserialize) and continue with the request now it knows who made it. If there is no matching or valid session, like with logging in, we don't know who the user is, so we can end the request there.
 
 This is exactly like having a name badge or access pass at work or some event. The cookie is like an access pass which you (the client) give to a machine or security (the server) and it checks who you are and if you're allowed in or not. You can go home and come back the next day, reusing that pass as many times as you need so long as you still have it and your details are still in the system.
 
 ### Implementing sessions
 
-Let's use [express-session](https://expressjs.com/en/resources/middleware/session.html) to implement a basic session authentication system - code along! For the purpose of streamlining our example, we'll put all of the JavaScript in `app.js` and start with hardcoding a few things. When it comes to your own projects, separate different parts like routes, controllers etc. to their own files and folders as you'll have done before. The same can be done for any config for auth features.
+Let's use [express-session](https://expressjs.com/en/resources/middleware/session.html) to implement a basic session authentication system - code along! For the purpose of streamlining our example, we'll put all of the JavaScript in `app.js` and start with hardcoding a few things. When it comes to your own projects, separate different parts like routes, controllers etc. to their own files and folders as you'll have done before. The same can be done for any configuration for authentication features.
 
 #### Setup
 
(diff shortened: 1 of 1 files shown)


nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,437 @@
...
+
+const PORT = process.env.PORT || 3000;
+app.listen(PORT, () => {
+  console.log(`App listening on port ${PORT}!`);
+});
+```
+
+#### Session store
+
+Let's talk about our session config which we apply to every incoming request (by mounting it on `app`). Firstly, we use the [connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple) library to make express-session store session data in our database (creating a "session" table if it does not already exist). Without this, sessions would be stored in memory by default which would not persist through any server restarts!
```
damon314159, line 109: Is there pedagogical benefit to having the learner make their own `sessions` table and writing their own helper functions to do CRUD ops on them? If the aim of the lesson is to get to grips with sessions, I'm not sure it's a good idea to abstract away the actual persistent session info

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,437 @@
...
+
+By far the worst way we can store passwords is to just store them in plaintext like we've done in our example app earlier. Even if we encrypted the passwords, all an attacker would need is the key to decrypt all the passwords. Let's face it, if someone managed to gain access to your database, it probably wouldn't be very hard for them to get the encryption key (assuming they don't already have it).
+
+Remember [hash functions](https://www.theodinproject.com/lessons/javascript-hashmap-data-structure#what-is-a-hash-code) from the Hashmap lesson? We want to hash our passwords, then store the hash since hashes are one-way functions. We also want to [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the password when hashing to prevent identical passwords from being stored with identical hashes. On top of all that, we also want the hash function to be purposely slow - not so slow that a normal user will be waiting ages just to log in but certainly slow enough to minimize the number of attempts an attacker might be able to make in a given amount of time.
```
damon314159, line 344: I think "prevent identical passwords from being stored with identical hashes" doesn't do the best job at making the point. I can see that we don't want to rabbit hole into rainbow tables here, but I wonder if "prevent attackers being able to compare hashes against precomputed hashes of common passwords" is just enough to get the idea across

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,437 @@
...
+app.post("/login", async (req, res, next) => {
+  try {
+    const { rows } = await pool.query(
+      "SELECT * FROM users WHERE username = $1",
+      [req.body.username],
+    );
+    const user = rows[0];
+
+    // argon2.verify requires an argon2 hash as the first argument
+    // so we must early return if there is no matching user
```
damon314159, line 383: There's a timing attack here, which leaks usernames. Do we think a note about this? Or just ignore it

MaoShizhong: I'm surprised I glossed over this, considering I mention generic validation feedback earlier in the lesson. Will address this - thinking of using a predetermined argon2id hash as a fallback if `user?.password` short circuits so that `argon2.verify()` can still run fine. Then only go down the happy path if both user exists and password matches.

Any alternative suggestions?

nodeJS/authentication/session_based_authentication.md
MaoShizhong: Possibly. My understanding from prior discussion (which may be flawed) was to stay with `express-session` but handle it manually rather than as part of `passport` and its strategies. But if @01zulfi thinks it'll be even better to do this lesson with properly manual session management then the lesson can be amended accordingly.

nodeJS/authentication/session_based_authentication.md
damon314159: That's probably the simplest approach. You probably want to pick a huge random string to hash, since if you use something that someone actually tries as a password, you might accidentally grant a session to a user that doesn't exist. Maybe hash a uuid.
Obligatory I am not a cybersec guy message

MaoShizhong: Yeah, the random argon2 hash would just be to force the hash verification. Then change the following conditional to `if (user && isMatchingPassword)`

nodeJS/authentication/session_based_authentication.md
MaoShizhong: What do you think about the following:

> We also want to [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the password when hashing so that the identical passwords will produce a different hash each time, preventing attackers from comparing hashes against precomputed hashes of common passwords (often referred to as "rainbow tables").

nodeJS/authentication/session_based_authentication.md
01zulfi: We want our learners to experience setting up the session table from scratch, insert and read from it, set up cookies and set up middlewares for auth. That's it, no other fancy stuff. That's where we introduce these libraries. 

But... now that I've skimmed through the lesson, it does a great job teaching session auth. I'll need to come back to this comment with concrete thoughts.

MaoShizhong: No problem at all. Let me know which direction you'd prefer the lesson to go in and I can adjust if necessary. Going a full manual approach would mean implementation would likely be a lot more simple (no secrets and stuff - I feel manually implementing that would be out of the scope of such a lesson, but also I feel the topic mention is key enough to sessions in general.

columk1: I think there's clarity to be gained by implementing the core functionality without express-session first, leaving out nuances like signing the sid. 

Abstracting simple functions away can make the middleware seem more complicated than it is, even though it's explained well. This first version could be swapped out for express-session afterwards. Although I also think it's a great lesson as is.

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
+### Introduction
+
+Now that we've been introduced to cookies and their various uses and properties, let's use them to help us implement something. We want to allow someone to log in once and let the server "remember" them, automatically recognizing any future requests from them.
+
+The basic login process (with a username and password) is rather straightforward. The user submits a form with their username and password, then the server checks the database to see if that username/password combo exists. If it does, great - it knows who the requester is and can continue with the rest of the request. If it does not exist, it does not know who the requester is, so it can end the request there and then.
+
+So if someone does successfully "log in", how does the server recognize that the next request that user sends is coming from them? Without a system to persist the login, it'd just be a plain ol' request like any other and could have come from anyone. To handle this, we will use **sessions**.
```
01zulfi, line 7: We can say that one way of doing this would be sending credentails in every request, and authenticating them. But with this suggestion change, the paragraph doesn't flow very well. I'll defer it to you whether you want to add something like this

```suggestion
So if someone does successfully "log in", how does the server recognize that the next request that user sends is coming from them? One way would be to send the username and password for every request, and then perform the necessary action after authentcation. As you can imagine, this can be cumbersome..  Without a system to persist the login, it'd just be a plain ol' request like any other and could have come from anyone. To handle this, we will use **sessions**.
```

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
...
+- Implement authentication with sessions.
+- Use a database as a session store.
+- Explain how and why passwords are hashed before being stored.
+
+### Sessions
+
+A session is just information about a user's interaction with the site in a given time period and can be used to store a whole variety of data. For persisting logins, we can store (serialize) some information about that user, such as their user ID, in a database table. That data will have its own ID and may also have an expiry time. We can then store that session's ID in a cookie (it doesn't need anything else stored in it) and send it back to the user in the server response.
+
+The client now has that cookie with the session ID and can then attach it to any future requests. The server can then check the database for a valid session with the same ID it found in the cookie. If there is a matching session, great - it can extract the serialized user information (deserialize) and continue with the request now it knows who made it. If there is no matching or valid session, like with logging in, we don't know who the user is, so we can end the request there.
```
01zulfi, line 23: See comment on "end request" above

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
...
+  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+  username VARCHAR ( 255 ),
+  password VARCHAR ( 255 )
+);
+```
+
+Now we'll set up a minimal Express app. We'll need the following dependencies installed first:
+
+```bash
+npm install express express-session ejs dotenv pg connect-pg-simple
```
01zulfi, line 46: We need to include the step where we set up the session table for `connect-pg-simple`https://www.npmjs.com/package/connect-pg-simple

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
...
+
+const PORT = process.env.PORT || 3000;
+app.listen(PORT, () => {
+  console.log(`App listening on port ${PORT}!`);
+});
+```
+
+#### Session store
+
+Let's talk about our session config which we apply to every incoming request (by mounting it on `app`). Firstly, we use the [connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple) library to make express-session store session data in our database (creating a "session" table if it does not already exist). Without this, sessions would be stored in memory by default which would not persist through any server restarts!
```
01zulfi, line 109: I don't think we need to mention this

```suggestion
Let's talk about our session config which we apply to every incoming request (by mounting it on `app`). Firstly, we use the [connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple) library to make express-session store session data in our database (creating a "session" table if it does not already exist).
```

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
...
+app.post("/login", async (req, res, next) => {
+  try {
+    const { rows } = await pool.query(
+      "SELECT * FROM users WHERE username = $1",
+      [req.body.username],
+    );
+    const user = rows[0];
+
+    if (user?.password === req.body.password) {
+      req.session.userId = user.id;
```
01zulfi, line 205: We can add a sentence to explain why we're doing this

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
...
+
+Remember [hash functions](https://www.theodinproject.com/lessons/javascript-hashmap-data-structure#what-is-a-hash-code) from the Hashmap lesson? We want to hash our passwords, then store the hash since hashes are one-way functions. We also want to [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)) the password when hashing so that the identical passwords will produce a different hash each time, preventing attackers from comparing hashes against precomputed hashes of common passwords (often referred to as "rainbow tables"). On top of all that, we also want the hash function to be purposely slow - not so slow that a normal user will be waiting ages just to log in but certainly slow enough to minimize the number of attempts an attacker might be able to make in a given amount of time.
```
01zulfi, line 344: > On top of all that, we also want the hash function to be purposely slow - not so slow that a normal user will be waiting ages just to log in but certainly slow enough to minimize the number of attempts an attacker might be able to make in a given amount of time.

Sounds interesting, why would we want hashing to be purposefully slow, can't we rate limit login?

nodeJS/authentication/session_based_authentication.md
```diff
@@ -0,0 +1,445 @@
...
+
+You don't need to know all the details of specific attack techniques but in this case, it doesn't take much to ensure that the same process always runs regardless of whether a user exists or not.
+
+</div>
+
+### Assignment
+
+<div class="lesson-content__panel" markdown="1">
+
+1. Bookmark the [express-session] docs if you haven't yet.
```
01zulfi, line 422: We can add link to the docs

01zulfi (changes requested):
Great stuff, left a few comments below

nodeJS/authentication/session_based_authentication.md
MaoShizhong: How much into rate limiting vs slow hashing do you feel it'd be appropriate to go into? My initial thoughts are separating to a new paragraph then briefly mentioning slow hashing in case the database is breached/leaked and people try to break through offline. But then if I talk about rate-limiting, how much of that would you want in this lesson? Would you want the lesson example app to demonstrate installing `express-rate-limit` and setting a limit? Or an alternative?

nodeJS/authentication/session_based_authentication.md
MaoShizhong: .......oops

nodeJS/authentication/session_based_authentication.md
01zulfi: I think we should cover neither. This could be explored in a future lesson perhaps

MaoShizhong: Fair enough. Perhaps then that part can be replaced with something like

> There are even more things to account for and that's why we have purpose-built hashing algorithms like argon2.

Then that should lead nicely into the next section.

nodeJS/authentication/session_based_authentication.md
MaoShizhong: ....oops

nodeJS/authentication/session_based_authentication.md
MaoShizhong: Let me know how the [new paragraph wording](https://github.com/TheOdinProject/curriculum/pull/29082/files#diff-20aacaf508a0cee5280bfca178274458b785503c3ec8e40766bb56f4dd0c34ffR7) reads to you.

nodeJS/authentication/session_based_authentication.md
MaoShizhong: https://github.com/TheOdinProject/curriculum/pull/29082/commits/f93b3154c6bf563a3de6e3226366dea688cd758a

Let me know how you feel about this solution. Looking at express session docs further, annoyingly we need to manually save before a redirect else the redirect *could* trigger a page load before the session writing. On the one hand, it's annoying and also doesn't have Promise syntax available - callback only for now. On the other hand, I actually think it makes the process in the function a little clearer as to what's actually going on.

nodeJS/authentication/session_based_authentication.md
01zulfi: sounds good to me

nodeJS/authentication/session_based_authentication.md
MaoShizhong: Ah I see, it's not actually that clear from the paragraph after the code block.

nodeJS/authentication/sessions.md
```diff
@@ -0,0 +1,463 @@
...
+So if someone does successfully "log in", how does the server recognize that the next request that user sends is coming from them? Without a system to persist the login, it'd just be a plain ol' request like any other and could have come from anyone. We *could* send the username and password with every request and authenticate that every time, but as you can imagine, this can be quite cumbersome. To handle this better, we will use **sessions**.
+
+### Lesson overview
+
+This section contains a general overview of topics that you will learn in this lesson.
+
+- Describe what sessions are.
+- Explain how sessions and cookies can be used together to persist logins.
+- Implement authentication with sessions.
+- Use a database as a session store.
```
01zulfi, line 16: Not required
```suggestion
```

nodeJS/authentication/sessions.md
```diff
@@ -0,0 +1,463 @@
...
+app.use(session({
+  store: new pgSession({
+    pool: pool,
+    createTableIfMissing: true,
+  }),
+  resave: false,
+  saveUninitialized: false,
+  secret: process.env.SESSION_SECRET,
+  cookie: {
+    httpOnly: process.env.NODE_ENV === 'prod',
```
01zulfi, line 93: lets use full word here
```suggestion
    httpOnly: process.env.NODE_ENV === 'production',
```

nodeJS/authentication/sessions.md
```diff
@@ -0,0 +1,463 @@
...
+    secure: process.env.NODE_ENV === 'prod',
```
01zulfi, line 94: same here
```suggestion
    secure: process.env.NODE_ENV === 'production',
```

nodeJS/authentication/sessions.md
```diff
@@ -0,0 +1,463 @@
...
+
+const PORT = process.env.PORT || 3000;
+app.listen(PORT, () => {
+  console.log(`App listening on port ${PORT}!`);
+});
+```
+
+#### Session store
+
+Let's talk about our session config which we apply to every incoming request (by mounting it on `app`). Firstly, we use the [connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple) library to make express-session store session data in a "session" table in our database, creating the table if it does not already exist. If you look inside your database in psql, you'll be able to see what the session table looks like. You can also have a look at the [SQL queries for creating the session table](https://github.com/voxpelli/node-connect-pg-simple/blob/HEAD/table.sql).
```
01zulfi, line 112: I think we should ask learners to look at the table.

```suggestion
Let's talk about our session config which we apply to every incoming request (by mounting it on `app`). Firstly, we use the [connect-pg-simple](https://www.npmjs.com/package/connect-pg-simple) library to make express-session store session data in a "session" table in our database, creating the table if it does not already exist. Go inside your database in psql, you'll be able to see what the session table looks like. You can also have a look at the [SQL queries for creating the session table](https://github.com/voxpelli/node-connect-pg-simple/blob/HEAD/table.sql).
```

I think we should explain the session table ourselves as well. Just some text on each column's purpose should do

nodeJS/authentication/sessions.md
```diff
@@ -0,0 +1,463 @@
...
+    if (user?.password === req.body.password) {
+      // serialize the user ID in the session object so it can be retrieved later
+      req.session.userId = user.id;
+      req.session.save((err) => {
+        if (err) {
+          next(err);
+        } else {
+          res.redirect("/");
+        }
+      });
```
01zulfi, line 218: We should add some description to this. Something like

> the userId gets saved against this session in the database. for subsequent requests, this session (which will be retrieved from the cookie) will contain the userId, which we can use to authenticate

nodeJS/authentication/sessions.md
```diff
@@ -0,0 +1,463 @@
...
+
+```javascript
+async function checkAuthenticated(req, res, next) {
+  try {
+    if (!req.session.userId) {
+      res.redirect("/login");
+    } else {
+      const { rows } = await pool.query(
+        "SELECT * FROM users WHERE id = $1",
+        [req.session.userId],
```
01zulfi, line 302: I think this might feel a bit _magical_  to the learner. Hopefully some changes above might help

01zulfi (changes requested):

nodeJS/authentication/sessions.md
MaoShizhong: Let me know how the changes feel in regards to this bit.

01zulfi (approved):
@MaoShizhong this is lookin great, can't wait for it to be released 🔥

mao-sz:
Added a little terminology clarification note box now that later lessons have been written. The term "session management" as a more general concept (not specifically stateful sessions) is used in later lessons as well as linked resources, so it'd probably be sensible to clarify that terminology at the start to reduce confusion later on.

PR #29190, opened 2024-12-13
Update transforms.md

alexanderjuxoncobb:
The description of order of transformations was incorrect. MDN says that composition is right to left, which is talking about the inner function g of f(g(x)) being done first. This is equivalent to transform: g() f(), so when laid out the transform actually goes from left to right when reading it. I have changed the description to describe this change too.

## Because

The description of order of transformations was incorrect. MDN says that composition is right to left, which is talking about the inner function g of f(g(x)) being done first. This is equivalent to transform: g() f(), so when laid out the transform actually goes from left to right when reading it. I have changed the description to describe this change too.

## This PR

1.  description for blue and red squares

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/advanced_html_css/animation/transforms.md
+++ b/advanced_html_css/animation/transforms.md
@@ -157,9 +157,9 @@ on <a href="https://codepen.io">CodePen</a>.</span>
 
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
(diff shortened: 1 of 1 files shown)


JackBogart (commented):
Transforms aren't left to right, the lesson and MDN are correct.

PR #29343, opened 2025-01-22
fix: Add generator to export image assets with their original names

oldraphash:
I added a snippet of generating code to export the image assets with their original names and extensions, I cleaned up the part that says you don't need to understand how to change the names automatically generated by webpack. At the end, **it includes a webpack article explaining the generator code snippet**.

## Because
I think the changes in this commit make it easier for students to understand, it's a **small** and **significant change**.

## Changes

- Added generator to export assets with their original names.
- Cleanup  the part that says "you don't need to understand how to change the names automatically generated by webpack"
- Added a webpack guide explaining the generator, and how u can use custom filenames when bundle.

### DIFF
--- a/javascript/organizing_your_javascript_code/webpack.md
+++ b/javascript/organizing_your_javascript_code/webpack.md
@@ -280,6 +280,10 @@ There are three different ways you could be dealing with local image files:
    {
      test: /\.(png|svg|jpg|jpeg|gif)$/i,
      type: "asset/resource",
+     generator: {
+        // Exports (png, svg, etc.) with their original names
+        filename: "[name][ext]",
+      },
    }
    ```
 
@@ -331,13 +335,16 @@ module.exports = {
       {
         test: /\.(png|svg|jpg|jpeg|gif)$/i,
         type: "asset/resource",
+        generator: {
+          filename: "[name][ext]",
+        },
       },
     ],
   },
 };
 ```
 
-You might notice that when images are included when bundling, the output image file in `dist` has a different file name (it will likely be some jumble of numbers and letters). By default, Webpack gives your bundled image files a new name by hashing their contents. You do not need to know how this works, nor do you need to dig into the details of why, nor how to change it. You just need to be aware that this is expected behavior (it's to do with preventing issues with the browser cache and matching file names).
+You might notice that when images are included when bundling, the output image file in `dist` has a different file name (it will likely be some jumble of numbers and letters). By default, Webpack gives your bundled image files a new name by hashing their contents. You just need to be aware that this is expected behavior (it's to do with preventing issues with the browser cache and matching file names).
 
 <div class="lesson-note lesson-note--tip" markdown="1">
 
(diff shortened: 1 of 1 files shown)


javascript/organizing_your_javascript_code/webpack.md
```diff
@@ -331,13 +335,16 @@ module.exports = {
...
+          filename: "[name][ext]",
+        },
       },
     ],
   },
 };
 ```
 
-You might notice that when images are included when bundling, the output image file in `dist` has a different file name (it will likely be some jumble of numbers and letters). By default, Webpack gives your bundled image files a new name by hashing their contents. You do not need to know how this works, nor do you need to dig into the details of why, nor how to change it. You just need to be aware that this is expected behavior (it's to do with preventing issues with the browser cache and matching file names).
+You might notice that when images are included when bundling, the output image file in `dist` has a different file name (it will likely be some jumble of numbers and letters). By default, Webpack gives your bundled image files a new name by hashing their contents. You just need to be aware that this is expected behavior (it's to do with preventing issues with the browser cache and matching file names).
```
MaoShizhong, line 347: By setting a custom generator to preserve the original file names, this entire paragraph will no longer make sense.

I purposely left the default hashed asset behaviour because it requires both less manual config and because it's generally more desirable than preserving the original file name to prevent caching issues, as this paragraph says. Since users won't be developing within `dist`, IMHO there's not really a strong reason to encourage learners to change the default asset generator behaviour. They will only be developing within `src` where everything has its original name, of course. For the `dist` assets, they just need to be made aware of the default behaviour, that it's expected (so they don't go "eeeer....I think it's broken", and that they don't need to dig into how the hashing works etc. because it's not relevant to the main learning points.

So personally, I think it's more out of scope to introduce custom generators and assign reading on them. Only my opinion and reasoning when this was written, so we'll see what the team thinks.

raphash: I thought about this suggestion and edited the paragraph because it is just 3 lines of code that will make a significant change and will facilitate students' understanding in my opinion . Preserving the original asset names is very important for **SEO** purposes as the `dist` directory is the directory that will be deployed.


_"For aesthetic and **SEO** purposes this change will be worth it in my opinion"_

MaoShizhong: In my honest opinion, SEO is out of scope of this lesson as well as a lot of the curriculum. Not to say it's not something that shouldn't ever be explored, just the value isn't quite the same in this lesson.

This lesson's goal is to showcase the point of bundlers and expose the bare minimum needed to get things done, as well as any vital things as part of the dev process (like the dev server and source maps), and is why other useful features (like image optimisation) are not included. There is already a lot in this lesson which, arguably, has long-term scope to streamline in terms of the pedagogy and learning experience.

Webpack is only used for the JavaScript course anyway, where later courses use different bundlers and also stay away from teaching about more manual config than is appropriate for the lessons' teaching goals.

raphash: Yes, I respect your opinion, from a learning point of view it would be easier for students to keep their name with the hash. In my honest opinion, the changes would make sense, as it is a desired behavior from my point of view, and which can be easily achieved by configuring a rule, in my opinion this encourages students to seek content also to achieve their own goal.

PR #29411, opened 2025-02-13
What is Node.js?: Rephrase info about leaving out React for later

MaoShizhong:
## Because

Despite the current [advice re: leaving out React for the start of the Node course](https://www.theodinproject.com/lessons/nodejs-introduction-what-is-nodejs#a-word-of-advice), it's still a somewhat common question in the community about why the course starts with EJS instead of React, or asking if it will teach about integrating React for fear of not getting practice. Some even go out of their way to try and use React server components to avoid using EJS.

This implies the current advice is not visible enough and/or its point isn't clear enough. Feedback from such askers have suggested both are valid reasons. The note can be rephrased to be clearer on *why* the course is doing things its way, as well as make more visible via a note box.

## This PR

- Puts info re: React in a note box
- Rephrases note contents

## Issue

N/A

## Pull Request Requirements

### DIFF
--- a/nodeJS/introduction_to_nodeJS/introduction_what_is_nodeJS.md
+++ b/nodeJS/introduction_to_nodeJS/introduction_what_is_nodeJS.md
@@ -19,7 +19,7 @@ This section contains a general overview of topics that you will learn in this l
 
 The [Node.js website](https://nodejs.org/en/about/) declares:
 
-> <span id="what-is-node">As an asynchronous event driven JavaScript runtime, Node is designed to build scalable network applications.</span>
+> As an asynchronous event driven JavaScript runtime, Node is designed to build scalable network applications.
 
 This is a definition that requires a little unpacking.
 
--- a/nodeJS/introduction_to_nodeJS/introduction_what_is_nodeJS.md
+++ b/nodeJS/introduction_to_nodeJS/introduction_what_is_nodeJS.md
@@ -62,9 +62,13 @@ http.createServer(function (req, res) {
 
 This snippet is from the very first lesson in a tutorial that you'll be following very soon. Basically this code is creating a server and saying, "any time we get a network request, run this callback function". This function happens to respond with the text 'Hello World!'. So if you go to a browser and navigate to the correct address and port, you would see that text on your screen.
 
-### A word of advice
+<div class="lesson-note" markdown="1">
+
+#### React frontends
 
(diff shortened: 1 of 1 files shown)


rlmoser99:
I like this idea. However, I have two concerns that I'd like to discuss.

I believe the heading (or title of the block) should summarize this section, something like "Use of React during the Node Course". 

In addition, I think having 3 paragraphs inside this warning block is overwhelming. I wonder if just having the last paragraph in a warning block would stand out and grab people's attention.

MaoShizhong:
> I like this idea. However, I have two concerns that I'd like to discuss.
> 
> I believe the heading (or title of the block) should summarize this section, something like "Use of React during the Node Course".
> 
> In addition, I think having 3 paragraphs inside this warning block is overwhelming. I wonder if just having the last paragraph in a warning block would stand out and grab people's attention.

Fair comment - how does the following sound?

> We will start by focusing on server-side concepts, so client-side things like React will be left for later when you start combining more complex frontends and backends together. You should follow the course as it is written; deviating from the directions can make it more difficult than it needs to be.

I still believe an explanation for why React is left for later (focus on server-side concepts first) should be included as there are those who are aware of the note but not sure why not just use React for the backend/build React front ends for a simple backend etc. Hopefully the above wording is a little easier to digest.

rlmoser99:
That re-wording is great! I really like how you've shortened it, but kept all of the important information. I don't think that will be too overwhelming to be in the warning block.

nodeJS/introduction_to_nodeJS/introduction_what_is_nodeJS.md
```diff
@@ -19,7 +19,7 @@ This section contains a general overview of topics that you will learn in this l
 
 The [Node.js website](https://nodejs.org/en/about/) declares:
 
-> <span id="what-is-node">As an asynchronous event driven JavaScript runtime, Node is designed to build scalable network applications.</span>
+> As an asynchronous event driven JavaScript runtime, Node is designed to build scalable network applications.
```
rlmoser99, line 22: When I look at lines 20 - 24 in the markdown preview tool, I am not sure it displays as you intend.
https://www.theodinproject.com/lessons/preview

MaoShizhong: image

They look the same on my end - the only difference being there's no redundant span with duplicate ID in the former.

rlmoser99: Ah, I see. When I copied & pasted, it did not keep an empty line between the quote and the following line, like it usually does. When I make sure it is there, it looks correct!

rlmoser99 (approved):

PR #29602, opened 2025-04-09
ruby course, project management: fix grammatically incorrect uses of a hyphen (-)

Kezy-d4:
## Because

Please refer to this excerpt from Grammarly's article titled [How to Use a Dash Correctly: En Dash vs. Em Dash](https://www.grammarly.com/blog/punctuation-capitalization/dash/#:~:text=A%20dash%2C%20which%20is%20longer,in%20compound%20terms%20or%20prefixes.):

> **What’s the difference between a dash and a hyphen?**
> A dash, which is longer than a hyphen, is used to indicate ranges, pauses, or emphasis, with common types being the en dash (–) for ranges and the em dash (—) for breaks or emphasis. A hyphen (-) is shorter and primarily used to join words or parts of words, such as in compound terms or prefixes.

Thus, according to the above definition, this PR replaces any grammatically incorrect use of a hyphen (-) throughout the lesson with an em dash (—) to better indicate breaks and emphasis. Any grammatically correct use of a hyphen (-), as well as any use necessary for links, code and bullet lists have been maintained. 

## This PR

Please refer to the diff in order to review the specific changes. I suggest making use of your browser's "find" feature (CTRL/Cmd + F) which should make it easier to differentiate between the hyphens (-) and em dashes (—) while reviewing the diff.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/ruby/object_oriented_programming_basics/managing_ruby_projects.md
+++ b/ruby/object_oriented_programming_basics/managing_ruby_projects.md
@@ -23,7 +23,7 @@ Recall how the projects you worked on in Foundations were structured. The HTML, 
 
 Organizing your projects into different files has many practical benefits. Chief among them is making your code more modular, making it easier to adjust and understand the code as it gets more complex.
 
-Remember that old saying about how to keep your physical environment organized - "A place for everything and everything in its place". The same principle applies to software projects.
+Remember that old saying about how to keep your physical environment organized — "A place for everything and everything in its place". The same principle applies to software projects.
 
 For Ruby projects, the rule of thumbs are:
 
@@ -101,7 +101,7 @@ require 'sort/merge_sort'
 
 ```
 
-Ah. Of course - an error - it can't find `lib/sort`! Those are not relative paths... Fancy schmancy `require_relative` and its implicitly assuming the paths are relative!
+Ah. Of course — an error — it can't find `lib/sort`! Those are not relative paths... Fancy schmancy `require_relative` and its implicitly assuming the paths are relative!
 
 ```ruby
 # You're in the root of the project, the directory that holds main.rb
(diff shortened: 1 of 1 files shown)


JoshDevHub (commented):
I'll have to give this a bit of thought. A couple things that are on my mind:
1. I agree that writing these as em dashes is correct--at least if they're going to be dashes.
1. The curriculum currently only has a few other instances of em dashes, and there also don't seem to be many hyphens as punctuation in this type of context. This makes me a bit inclined to just rewrite these instances in this lesson with different punctuation and avoid the use of em dash at all.
1. But if we do write this with em dash, they are almost always written without spaces on either end of the dash, and your changes still include that spacing.

But yeah I'll think about this some. Feel free to leave any thoughts you may have about any of the points I raised.

Kezy-d4:
Hi, thanks for reviewing. You raise some fair points. I'm not sure which approach would be best. 

If we decide to rewrite those areas, I think it's best if someone more experienced tackles that just to make sure no crucial detail or info is lost in the translation. On the other hand, if we decide to keep the em dashes, I'd be happy to go ahead and remove the spacing around them as you mentioned.

Kezy-d4:
Something else has just become apparent to me. I have actually noticed occurrences of hyphens being used incorrectly throughout some other lessons in the Ruby course, and it may be the case elsewhere in the curriculum as well. Whichever solution we'd go for, I could see why you would want to apply that solution across the board for the sake of consistency, rather than on just one specific lesson. The problem is, that might prove to be a fairly gargantuan task, much larger than was anticipated or considered when making the changes for this lesson in isolation.

JoshDevHub (changes requested):
Sorry for the delay on this @Kezy-d4 

I think I'm fine to allow the use of em dashes. You should eliminate the spacing on either side of the em dashes. You can make another commit onto this PR that does this.

You can also open additional PRs to address other lessons where hyphens are used for sentence breaks.

Kezy-d4:
No worries. I've made the new commit and the spacing should now be removed on either side of each em dash.

PR #29607, opened 2025-04-10
Data Types and Conditionals: Fix grammar

K0R0BKIN:


foundations/javascript_basics/data_types_and_conditionals.md
```diff
@@ -22,7 +22,7 @@ Depending on what kind of work you're doing, you might end up working more with
...
+1. Vocabulary time: a **method** is a bit of functionality built into the language or specific data types. In [the W3Schools lesson on string methods](https://www.w3schools.com/js/js_string_methods.asp), you've learned about a few methods that can be used on strings, such as `replace` and `slice`. There are still many more built-in string methods for which [the MDN documentation provides an exhaustive reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String). You are not expected to memorize these but the documentation will be a very useful reference to revisit, so bookmark it!
```
MaoShizhong, line 25: For accessibility, link text needs to be able to stand by itself. Without "for strings" then the label does not sufficiently describe the link contents (too vague). At the same time, my suggestion (not a maintainer, just a personal one) would tweak the link text further so only the necessary text is part of the link label.
```suggestion
1. Vocabulary time: a **method** is a bit of functionality built into the language or specific data types. In [the W3Schools lesson on string methods](https://www.w3schools.com/js/js_string_methods.asp), you've learned about a few methods that can be used on strings, such as `replace` and `slice`. There are still many more built-in string methods for which the [MDN documentation for strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) provides an exhaustive reference. You are not expected to memorize these but the documentation will be a very useful reference to revisit, so bookmark it!
```

MaoShizhong: I like it :+1:

JoshDevHub (approved):
Thank you for contributing!

In the future, please don't delete the PR template. Even if your change is seemingly very simple, it's valuable to the maintainers that you fill this in. Some maintainers may not review work that hasn't had the template filled in.

PR #29611, opened 2025-04-12
Databases and SQL: Fix grammar and clarify SQL concepts

Maddily:
## Because

The lesson had some grammar issues and unclear phrasing around certain SQL concepts. These edits aim to improve clarity, accuracy, and readability for learners.

## This PR

- Fixes grammar, punctuation, and sentence structure throughout the lesson
- Clarifies when column names must be specified in `INSERT INTO` statements
- Explains limited use of `*` in aggregate functions like `COUNT(*)` vs `MAX(*)`. The "(because how would you `MAX` a column for "name"?)" part is removed because the issue isn’t really about the name column, but that `MAX(*)` doesn't make sense because `MAX` requires a specific type of column
- Adds examples to support conceptual clarifications
- Corrects "whole database" to "whole table"
- Changes incorrect reference to the "`HAVING` function" to the correct term, "`HAVING` clause"

## Issue

Closes #XXXXX

## Pull Request Requirements


databases/databases/databases_and_sql.md
```diff
@@ -34,53 +34,65 @@ This section contains a general overview of topics that you will learn in this l
...
+SQL is the language used to talk to many relational databases.  These databases use lots of tables to store different types of data (e.g., "users" and "posts" tables).  Tables are long lists like spreadsheets where each row is a different record (or object, e.g., a single user) and each column is one of that record's attributes (like name, email, etc.).  The one column that all tables include is an "ID" column, which gives the unique row numbers, and is called the record's "primary key."
 
-You can "link" tables together by making one of the columns in one table point to the ID of another table, for instance a row in the "posts" table might include the author's ID under the column called "user_id".  <span id='foreign-key'>Because the "posts" table has the ID of another table in it, that column is called a "foreign key"</span>.
+You can "link" tables together by making one of the columns in one table point to the ID of another table; for instance, a row in the "posts" table might include the author's ID under the column called "user_id."  <span id='foreign-key'>Because the "posts" table has the ID of another table in it, that column is called a "foreign key."</span>
```
JoshDevHub, line 39:
```suggestion
You can "link" tables together by making one of the columns in one table point to the ID of another table. For instance, a row in the "posts" table might include the author's ID under the column called `user_id`.  <span id='foreign-key'>Because the "posts" table has the ID of another table in it, that column is called a "foreign key."</span>
```
I think all of these examples in the lesson where it's referring to table and column names should be switched to inline code instead of double quotes. For one, this is the standard way people will refer to them, but also it maintains clarity in an instance like this. The column name is `user_id`, not `user_id.`. So feel free to make this change throughout the lesson.

Also instead of semicolon, just make two sentences.

databases/databases/databases_and_sql.md
```diff
@@ -34,53 +34,65 @@ This section contains a general overview of topics that you will learn in this l
...
 
-"Create" queries use `INSERT INTO` and you'll need to specify which columns to insert stuff into and then which values to put in those columns, which looks something like `INSERT INTO users (name, email) VALUES ('foobar','foo@bar.com');`.  This is one of the few queries that you don't need to be careful about which rows you've selected since you're actually just adding new ones into the table.
+"Create" queries use `INSERT INTO` and you'll need to specify which columns to insert stuff into if you're not inserting values for every column or not inserting them in the correct order, followed by the values to insert into those columns. For example, if all columns are filled in the correct order, you can omit the column names:
```
JoshDevHub, line 57:
```suggestion
"Create" queries use `INSERT INTO` and you'll need to specify which columns to insert stuff into if you're not inserting values for every column or not inserting them in the correct order, followed by the values to insert into those columns. For example, if all columns are filled in the correct order, you can omit the column names:
```
Sure this is possible, but inserting without specifying column names is also NOT a good idea. And because of that, I'm a bit reticent to even bring it up.

You can maybe mention it as an aside, but I also wouldn't go as deep into it as you have here, and I'd have a note about how it's considered poor practice.

databases/databases/databases_and_sql.md
```diff
@@ -107,7 +119,7 @@ Now we're getting into the fun stuff.  Aggregate functions like `COUNT` which re
...
 
-The last nifty trick is if you want to only display a subset of your data.  In a normal situation, you'd use a `WHERE` clause to narrow it down.  But if you've used an aggregate function like `COUNT` (say to get the count of posts written for each user in the example above), `WHERE` won't work anymore.  <span id='having-function'>So to conditionally retrieve records based on aggregate functions, you use the `HAVING` function, which is essentially the `WHERE` for aggregates</span>.  So say you only want to display users who have written more than 10 posts:
+The last nifty trick is if you want to only display a subset of your data.  In a normal situation, you'd use a `WHERE` clause to narrow it down.  But if you've used an aggregate function like `COUNT` (say to get the count of posts written for each user in the example above), `WHERE` won't work anymore.  <span id='having-function'>So to conditionally retrieve records based on aggregate functions, you use the `HAVING` clause, which is essentially the `WHERE` for aggregates</span>.  So say you only want to display users who have written more than 10 posts:
```
JoshDevHub, line 122:
```suggestion
The last nifty trick is if you want to only display a subset of your data.  In a normal situation, you'd use a `WHERE` clause to narrow it down.  But if you've used an aggregate function like `COUNT` (say to get the count of posts written for each user in the example above), `WHERE` won't work anymore.  <span id='having-clause'>So to conditionally retrieve records based on aggregate functions, you use the `HAVING` clause, which is essentially the `WHERE` for aggregates</span>.  So say you only want to display users who have written more than 10 posts:
```
"Clause" is definitely correct, but while we're here, we should also change the knowledge check link name to use the correct terminology.

databases/databases/databases_and_sql.md
```diff
@@ -154,7 +166,7 @@ The following questions are an opportunity to reflect on key topics in this less
 - [Which SQL statement is associated with "Read" from the CRUD acronym?](#sql-read)
 - [Which `JOIN` statement keeps only the rows from both tables where they match up?](#inner-join)
 - [How do you use an aggregate function?](#aggregate-function)
-- [In which situation would you use the `HAVING` function?](#having-function)
+- [In which situation would you use the `HAVING` clause?](#having-function)
```
JoshDevHub, line 169:
```suggestion
- [In which situation would you use the `HAVING` clause?](#having-clause)
```
Corresponding update to the knowledge check link on the HAVING clause.

JoshDevHub (changes requested):
Thank you for making all these changes. I have a few suggestions for ways to improve further. Let me know if you have any questions.

databases/databases/databases_and_sql.md
Maddily: Inline code does make more sense. Thanks for the suggestions.

databases/databases/databases_and_sql.md
Maddily: Thanks for this suggestion. I didn’t realize it was considered poor practice.

databases/databases/databases_and_sql.md
Maddily: Appreciate the catch. I’ve made the change.

PR #29666, opened 2025-05-04
Forms and Data Handling: fix: specify that 'empty strings' will be ignored with { value : falsy }

GaneshS288:
## Because

The explaination for optional in express validator left out that ``empty strings`` will also be ignored with ``{value: falsy}``

## This PR

Changes
- This is because ``{ values: "falsy" }`` means values that aren’t ``undefined``, ``null``, ``false``, or ``0`` will still be validated.

To
- This is because ``{ values: "falsy" }`` means values that aren’t ``undefined``, ``null``, ``false``, ``0`` or ``empty strings`` will still be validated.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/nodeJS/express/forms_and_data_handling.md
+++ b/nodeJS/express/forms_and_data_handling.md
@@ -86,7 +86,7 @@ The `body()` function allows you to specify which fields in the request body sho
 ];
 ```
 
-This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy" }` means values that aren't `undefined`, `null`, `false`, or `0` will still be validated.
+This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy" }` means values that aren't `undefined`, `null`, `false`, `0` or `empty strings` will still be validated.
 
 ### Chaining validations
 


nodeJS/express/forms_and_data_handling.md
```diff
@@ -86,7 +86,7 @@ The `body()` function allows you to specify which fields in the request body sho
 ];
 ```
 
-This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy" }` means values that aren't `undefined`, `null`, `false`, or `0` will still be validated.
+This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy" }` means values that aren't `undefined`, `null`, `false`, `0` or `empty strings` will still be validated.
```
wise-king-sullyman, line 89:
```suggestion
This example marks `birthdate` field as optional, but still enforces the ISO8601 date format on inputs. This is because `{ values: "falsy" }` means values that aren't `undefined`, `null`, `false`, `0` or empty strings `""` will still be validated.
```

wise-king-sullyman: Since empty strings is not a literal value here I don't think we should wrap it in backticks, and instead either wrap an empty string in backticks like this or just leave it as empty strings, WDYT?

wise-king-sullyman (changes requested):

GaneshS288:
@wise-king-sullyman yeah your suggestion is reasonable. I committed it since I think "empty strings ``''``" is more explicit than just "empty strings".

wise-king-sullyman (approved):
Thanks!

PR #29743, opened 2025-05-26
Update async_and_await.md

Khaspper:
The update clarifies that we can use await top-level when we are using module scripts

## Because

This PR clarifies the behavior of top-level await in JavaScript, specifically noting that it only works in module script

## This PR

- Added clarification that await is not allowed at the top level in non-module scripts
- Explained how using <script type="module"> in HTML enables top-level await

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/javascript/asynchronous_javascript_and_apis/async_and_await.md
+++ b/javascript/asynchronous_javascript_and_apis/async_and_await.md
@@ -140,7 +140,7 @@ Remember the Giphy API practice project? (If not, you should go back and complet
 </script>
 ```
 
-Since `await` does not work on the global scope, we will have to create an `async` function that wraps our API call to Giphy.
+Since `await` does not work on the global scope (Unless using a module script), we will have to create an `async` function that wraps our API call to Giphy.
 
 ```javascript
 <script>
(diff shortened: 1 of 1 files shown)


javascript/asynchronous_javascript_and_apis/async_and_await.md
```diff
@@ -140,7 +140,7 @@ Remember the Giphy API practice project? (If not, you should go back and complet
 </script>
 ```
 
-Since `await` does not work on the global scope, we will have to create an `async` function that wraps our API call to Giphy.
+Since `await` does not work on the global scope (Unless using a module script), we will have to create an `async` function that wraps our API call to Giphy.
```
mao-sz:
```suggestion
Since `await` does not work in the top level of a non-module script, we will have to create an `async` function that wraps our API call to Giphy.
```
@JustWaveThings opinion on the wording suggestion above? Avoids "global scope" because technically the top level of a module is *not* global. Personally, I feel the Safari top-level await limitation (just researched - really, Safari??) is a non-issue as far as TOP is concerned, and by the time users may end up genuinely facing that, I doubt they'd be incapable of navigating the problem themselves (I doubt that'd likely be in TOP anyway).

JustWaveThings: I agree that the Safari bug is super weird, and is outside the scope of the curriculum. I am happy with your suggestion.

JustWaveThings (approved):
Thanks @Khaspper and @MaoShizhong !

PR #29747, opened 2025-05-27
Node Controllers: Remove express-async-handler content

MaoShizhong:
## Because

Express v5 has been the npm current for a while i.e. what will be installed with `npm install express`, and rejected Promise support for async middleware is now built-in: https://expressjs.com/en/guide/migrating-5.html#rejected-promises. `express-async-handler` is no longer needed.

I've tested this and can confirm the exact same behaviour built in:

Crashes server in v4, catches and calls `next(err)` in v5:
```js
app.get('/', async () => {
    throw new Error('catch me!');
};
```

Required in v4 to catch and call `next(err)` in v4:
```js
const asyncHandler = require('express-async-handler');

app.get('/', asyncHandler(async () => {
    throw new Error('catch me!');
});
```

## This PR

- Removes content on `express-async-handler`
- Rewords relevant sentences

## Issue

N/A

## Pull Request Requirements

### DIFF
--- a/nodeJS/express/controllers.md
+++ b/nodeJS/express/controllers.md
@@ -225,9 +225,9 @@ async function getAuthorById(req, res) {
 
 What if we have many async middleware functions and a lot of them respond the same way just with things like different status codes and response messages? Instead of `try/catch` in each middleware, Express automatically catches any thrown errors and calls `next(error)`, which will pass control over to a special "error middleware function", meaning we wouldn't need to catch and handle in the original controller itself.
 
-#### With a middleware
+#### Error handler middleware
 
-An error middleware function handles all errors in our application that come down from other middleware functions. We want to place this error middleware function at the very end of the application code to ensure it's the last middleware function executed and only handles errors bubbling down from preceding middleware functions.
+An error handler middleware function handles all errors in our application that come down from other middleware functions. We want to place this error middleware function at the very end of the application code to ensure it's the last middleware function executed and only handles errors bubbling down from preceding middleware functions.
 
 Add the following code in `app.js` at the end of all middleware functions in our application:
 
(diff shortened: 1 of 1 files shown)


nodeJS/express/controllers.md
```diff
@@ -218,40 +218,16 @@ async function getAuthorById(req, res) {
 
     // or we can call next(error) instead of sending a response here
     // Using `next(error)` however will only render an error page in the express' default view and respond with the whole html to the client.
-    // So we will need to create a special type of middleware function if we want a different response and we will get to that in a bit.
+    // So we will need to create a special type of middleware function if we want a different response and we will get to that...next.
```
ManonLef: haha pun intended? Does it need a space?
```suggestion
    // So we will need to create a special type of middleware function if we want a different response and we will get to that... next.
```

nodeJS/express/controllers.md
```diff
@@ -218,40 +218,16 @@ async function getAuthorById(req, res) {
...
-```
-
-<div class="lesson-note lesson-note--tip" markdown="1">
-
-The asyncHandler function in the `express-async-handler` library is just 6 lines of code. Try to take a guess on how it's implemented and then do [check out the asyncHandler source code](https://github.com/Abazhenov/express-async-handler/blob/master/index.js) for yourself.
-
-</div>
+What if we have many async middleware functions and a lot of them respond the same way just with things like different status codes and response messages? Instead of `try/catch` in each middleware, Express automatically catches any thrown errors and calls `next(error)`, which will pass control over to a special "error middleware function", meaning we wouldn't need to catch and handle in the original controller itself.
 
 #### With a middleware
```
ManonLef: While we're here, I feel this header could be more descriptive to indicate we are talking about error handling middleware. What do you think?

nodeJS/express/controllers.md
mao-sz: Definitely intended. Ellipses spacing seems to be different with every other English style guide (e.g. space before *and* after, after only, I've seen some even specify "foo . . . bar") but IIRC, most cases of this in the curriculum have a space after if there is a word following.

nodeJS/express/controllers.md
mao-sz: Agreed - probably best to use Express' naming for the heading

ManonLef (approved):

ManonLef:
Thanks again Mao!

PR #29804, opened 2025-06-26
ES6 Modules Lesson: clarify why 'farewellString' is not accessible 

tasmiachow:
## Because

Briefly explained why farewellString is not global — specifically noting that it is not returned, which helps clarify the contrast with greetingString. This edit improves readability for learners who may be confused about why one variable is accessible while the other is not.

## This PR

- Reworded explanation about `farewellString` in `es6_modules.md` to clarify that it is not returned and therefore not accessible globally.

## Issue

Closes #XXXXX
This PR is not tied to a specific open issue.

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/es6_modules.md
+++ b/javascript/organizing_your_javascript_code/es6_modules.md
@@ -61,7 +61,7 @@ const greeting = (() => {
 })();
 ```
 
-Now, the global variable `greeting` will contain `"Hello, Odinite!"` and so our code from `two.js` successfully logs this to the console. However, our private `farewellString` variable is not global, so that cannot be accessed anywhere in `two.js`. Through this, we are able to choose what to expose from one file to be made available to all files that follow it! This is why IIFEs were often called the "module pattern", because they allowed us to write modular code across multiple files before we were given "real modules".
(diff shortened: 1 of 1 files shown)


javascript/organizing_your_javascript_code/es6_modules.md
```diff
@@ -61,7 +61,7 @@ const greeting = (() => {
 })();
 ```
 
-Now, the global variable `greeting` will contain `"Hello, Odinite!"` and so our code from `two.js` successfully logs this to the console. However, our private `farewellString` variable is not global, so that cannot be accessed anywhere in `two.js`. Through this, we are able to choose what to expose from one file to be made available to all files that follow it! This is why IIFEs were often called the "module pattern", because they allowed us to write modular code across multiple files before we were given "real modules".
+Now, the global variable `greeting` will contain `"Hello, Odinite!"` and so our code from `two.js` successfully logs this to the console. However, our private `farewellString` variable is not global and not returned, so it cannot be accessed anywhere in `two.js`. Through this, we are able to choose what to expose from one file to be made available to all files that follow it! This is why IIFEs were often called the "module pattern", because they allowed us to write modular code across multiple files before we were given "real modules".
```
wise-king-sullyman: Hmm, I think this change actually might make things more confusing.

The variable itself is not available outside of the scope of the anonymous function _strictly_ because it isn't global. The same is true for the `greetingString` variable.

The _value_ of `farewellString` isn't available because it isn't returned though, so I do see where your confusion here could've come from.

Can you edit this a bit more to clarify that distinction?

tasmiachow: Okay sure that sounds good

PR #29830, opened 2025-07-04
Codespell: Ignore URIs and emails

MaoShizhong:
## Because

Can't think of a situation where we actually want codespell to check URLs or emails. Example false flag of `cryto` in a URL: #29814

## This PR

- Configures codespell action to ignore words in URLs and emails.

## Issue

Closes #29815 

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project curriculum contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x] If this PR addresses an open issue, it is linked in the `Issue` section
-   [x] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [x] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

PR #29814, opened 2025-06-28
New lesson: Session Management in Practice

mao-sz:
## Because

As part of the Node revamp's 2nd milestone

## This PR

- Adds a new lesson discussing TOP-scoped session management comparisons/necessities
 
## Issue

Closes #29735

## Pull Request Requirements

-   [x] I have thoroughly read and understand [The Odin Project curriculum contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md)
-   [x] The title of this PR follows the `location of change: brief description of change` format, e.g. `Intro to HTML and CSS lesson: Fix link text`
-   [x] The `Because` section summarizes the reason for this PR
-   [x] The `This PR` section has a bullet point list describing the changes in this PR
-   [x] If this PR addresses an open issue, it is linked in the `Issue` section
-   [x] If any lesson files are included in this PR, they have been previewed with the [Markdown preview tool](https://www.theodinproject.com/lessons/preview) to ensure it is formatted correctly
-   [x] If any lesson files are included in this PR, they follow the [Layout Style Guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md)

PR #29834, opened 2025-07-05
Revert "Codespell: Ignore URIs and emails"

MaoShizhong:
Reverts TheOdinProject/curriculum#29830

Codespell broken with ignore params. Example: #29814
Should revert first and I'll explore in a private repo exactly what's wrong, since it seems like it should work going by the action's docs.


rlmoser99 (approved):

mao-sz:
https://github.com/codespell-project/actions-codespell/issues/55

Seems there's a bug with `uri_ignore_words_list`, wildcard and actions, so addressing this in this repo will have to be put on hold.

PR #29849, opened 2025-07-07
Variables and Operators lesson : Fix misleading Filename

xerxes324:
Changed from "no name `.js` " to " `noname.js` " for better clarity, as the former might confuse people into creating a literally empty .js file.

## Because
" no name `.js`  " can be confusing for people as it might lead them to think its an empty .js file which can cause an error. Changing it to `noname.js` can probably help avoid the confusion. 

[ At the very least, it can be changed from the existing " no name `.js` " to " noname `.js` ". ] 

## This PR

* no name `.js` changed to `noname.js` in the variables_and_operators section of the Foundations Course.

## Issue

Closes #XXXXX

## Additional Information

Link to discord discussion : https://discord.com/channels/505093832157691914/505093832157691916/1391852492329062410

## Pull Request Requirements

### DIFF
--- a/foundations/javascript_basics/variables_and_operators.md
+++ b/foundations/javascript_basics/variables_and_operators.md
@@ -65,7 +65,7 @@ Another way to include JavaScript in a webpage is through an external script. Th
 
 JavaScript files have the extension `.js` similar to `.css` for stylesheets. External JavaScript files are used for more complex scripts.
 
-We named our file `javascript.js` but we could have chosen any name like `my-script.js` or even no name `.js`. What is really important is the `.js` extension.
+We named our file `javascript.js` but we could have chosen any name like `my-script.js` or even `noname.js`. What is really important is the `.js` extension.
 
 ### Variables
 


rlmoser99 (changes requested):
I see the confusion, but what it is saying here is that there is no name. The file is just `.js`. I am open to clarifying the sentence, maybe something like this:

```
We named our file `javascript.js` but we could have chosen any name like `my-script.js` or `.js` without a name. What is really important is the `.js` extension.
```

foundations/javascript_basics/variables_and_operators.md
```diff
@@ -65,7 +65,7 @@ Another way to include JavaScript in a webpage is through an external script. Th
 
 JavaScript files have the extension `.js` similar to `.css` for stylesheets. External JavaScript files are used for more complex scripts.
 
-We named our file `javascript.js` but we could have chosen any name like `my-script.js` or even no name `.js`. What is really important is the `.js` extension.
+We named our file `javascript.js` but we could have chosen any name like `my-script.js` or even `noname.js`. What is really important is the `.js` extension.
```
rlmoser99, line 68:
```suggestion
We named our file `javascript.js` but we could have chosen any name like `my-script.js` or `.js` without a name. What is really important is the `.js` extension.
```

pocketradio: https://discord.com/channels/505093832157691914/505093832157691916/1391831840867881131

Well, this was the source of the confusion. someone tried it and it gave them an error ( possibly because of the naming? )
I'm not sure, please do clarify.

mao-sz: It seems if the script src leads to a file name that begins with a dot, the browser interprets the MIME type as `text/html` even if it's explicitly set to `text/javascript` (which would otherwise be implicit). Tested this with various file names inc. `.n.js` and `n.js`, both with and without the leading `./`
`.js` as a plain file name does work when running via node e.g. `node .js` but in this example, it'll be better to use an example other than nameless `.js`

The `.js` also isn't the important part either. Naming and src-ing the file `javascript.rb` will work just fine (with JS code of course), as long as the path indicates a file (just `javascript` no extension will end up with a MIME type of `text/html`, even if explicitly set in the script tag.

sajwan05: I have tried it, and `.js` as a plain file name also does work in browser console and also with `node .js` as clarified by @MaoShizhong. Thankyou so much for clarification.

PR #29896, opened 2025-07-18
Update rendering_techniques.md

Yufan029:
## Because

Based on the context, the wording needs to be changed a bit.

## This PR

Original:
    If none of those checks passed, then we have the data we need to render the list successfully.
Changed:
    If all of those checks passed, then we have the data we need to render the list successfully.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/react/getting_started_with_react/rendering_techniques.md
+++ b/react/getting_started_with_react/rendering_techniques.md
@@ -263,7 +263,7 @@ function App() {
 
 The first `if` statement will now execute and return a `<div>` with the text "Loading..." This is often the case when you are fetching from an API, since it might take some time to actually retrieve the data, it is good practice to show an indicator for that.
 
-If none of those checks passed, then we have the data we need to render the list successfully. Try it out by adding items to the `animals` list and adding the property back.
+If all of those checks passed, then we have the data we need to render the list successfully. Try it out by adding items to the `animals` list and adding the property back.
 
 You can, of course, also accomplish this with just the ternary and `&&` operators.
 


mao-sz:
The original wording is correct. There are 2 checks, first a loading check then an empty check. If either of them pass, the list won't be rendered. The list only renders if neither of the guard checks pass, meaning it is not loading and it is not empty.

Yufan029:
> The original wording is correct. There are 2 checks, first a loading check then an empty check. If either of them pass, the list won't be rendered. The list only renders if neither of the guard checks pass, meaning it is not loading and it is not empty.

Thanks for the prompt reply, I still a bit confused for the working, not my first language, sorry.

Like emtpy check below, if props.animals.length === 0 will return. If props.animals.length !== 0, if check will be passed.
`
  if (props.animals.length === 0) {
    return <div>There are no animals in the list!</div>;
  }`

So I reckon only if both of the loading and empty checks passed, the logic could go further down to the functionality of render list, that's what I understand.

Only if the 'PASS' here means get into the if block, which means that I had a wrong understand of 'PASS', sorry for the cnfusion, please let me know if I made any sence.

mao-sz:
If a check passes that means it's true.

The empty check checks if length is 0. If that pisses, it means the length is 0, meaning the array is empty, therefore no list.

"Pass" here means pass or fail, not "go past it", in case that's what you were confused by?

Yufan029:
> If a check passes that means it's true.
> 
> The empty check checks if length is 0. If that pisses, it means the length is 0, meaning the array is empty, therefore no list.
> 
> "Pass" here means pass or fail, not "go past it", in case that's what you were confused by?

Yes, thanks for the explanation. That make sense, I misunderstand it. I did think like "go and pass it". I'll close this one, thanks.

PR #29987, opened 2025-08-11
fix: Pluralization typo for submit button

AncientNimbus:
## Because
A pluralization issue is found in the [Form Basics](https://www.theodinproject.com/lessons/intermediate-html-and-css-form-basics#submit-buttons) learning material. Since the `submit` type button creates a single button similar to `reset` and `generic`, the sub-heading for submit button should be singular.

## This PR
Update sub-header text to use singular “button” instead of “buttons”

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/intermediate_html_css/forms/form_basics.md
+++ b/intermediate_html_css/forms/form_basics.md
@@ -359,7 +359,7 @@ To create a button, we use the `<button>` element. The content or text we want t
 
 The button element also accepts a `type` attribute that tells the browser which kind of button it is dealing with. There are three types of buttons available to us.
 
-#### Submit buttons
+#### Submit button
 
 Once a user is finished filling in a form, they will need a way to submit it. There is a specialized button for this; the submit button. When a submit button is clicked, it will submit the form it is contained within. The `type` attribute has a value of submit by default, i.e. if the `type` is not specified or the value provided is invalid.
 


mao-sz:
IMO, `Submit buttons` makes more sense grammatically because the heading is for a section talking about submit buttons in general. The heading isn't referring to any individual submit button specifically. 

`The submit button` makes grammatical sense but forms are also not limited to a single submit button. It's perfectly valid to use multiple submit buttons for a single form.

AncientNimbus:
> IMO, `Submit buttons` makes more sense grammatically because the heading is for a section talking about submit buttons in general. The heading isn't referring to any individual submit button specifically.
> 
> `The submit button` makes grammatical sense but forms are also not limited to a single submit button. It's perfectly valid to use multiple submit buttons for a single form.

You are right about that, but I submit the correction mainly due to the context of the paragraph.

Initially, when I first read the heading I thought there are different types of submit button, but there is just one type. 

Also, "The submit button" is used throughout the explanation, the content focuses on the general concept of the button type. 

I hope this clarify my intent here :)

mao-sz:
`Submit button` as a heading wouldn't make sense though, since headings typically should be able to be read like `The following section is about <heading>` e.g. `The following section is about submit buttons` or `The following section is about the submit button`, but not `The following section is about submit button`.

Although in this case you misinterpreted "submit buttons" as "multiple types of submit buttons", I wouldn't say it's obviously ambiguous, in the same way a section heading called `Divs` or `Spans` could technically be interpreted as there being multiple types of divs or spans, but I'd imagine in most cases the more common interpretation would be talking about divs and spans in general.

AncientNimbus:
Since this is more or less a stylistic choice, and not a real bug I shall rest my case after we consider this one last aspect:

Perhaps it is worth considering MDN's reference for the sake of consistency.
- [Submit button - Glossary | MDN](https://developer.mozilla.org/en-US/docs/Glossary/Submit_button)

I mainly find it odd when reading the current learning material, the headers for button type `reset` and `generic` are both singular, while button type `submit` is plural, even though each section is primarily an explanation of the given type. The change here seems to improve the clarity of the subject. In any case, I am happy either way. 

I hope this make sense, thank you. :)

mao-sz:
This is only my opinion I'm offering, and a maint may well hold a different one.

Looking at the rest of the headings, I actually feel instead of changing this one to singular, the other ones should be changed to plural or all changed to `The <type> button`

Personally, `Submit button`, `Reset button` and `Generic button` don't read as sensibly in the context of the lesson flow. Though that is of course nitpicking. Agreed on the consistency issue though, I don't like how one is one way and the others are different.

Edit: Looking at other headings outside of the buttons part, there are plenty others also inconsistent, like `Placeholder attribute` then `The type attribute`, or `Text area` but `Labels`. I don't like this inconsistency and would suggest maints consider if the rest of these headings could also be tidied up with some consistency between them. `Plural` and `The <singular>` work well together because they can be read as `Let's talk about <heading>` but just `<singular>` can't  and seems jarring when mixed with the others IMHO.

AncientNimbus:
> Agreed on the consistency issue though, I don't like how one is one way and the others are different.

Yeah, that's the reason for the PR, as it is somewhat noticeable. Either way will work so long the clarity is established and it supports the lesson's flow.

[type <button>: The Button element - HTML | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#type) 

I think this is a better reference as it is a direct mirror to the subject.

mao-sz:
@AncientNimbus 
Am maint again - let's go with the 3 button subsection headings being pluralised, which should be the least work needing done to make the lesson consistent.

AncientNimbus:
> @AncientNimbus Am maint again - let's go with the 3 button subsection headings being pluralised, which should be the least work needing done to make the lesson consistent.

Thanks for the ping, I have update the file accordingly, and it is ready for review.

mao-sz (approved):
Many thanks!
We now have a new style guide rule that all note boxes must have a level 4 heading. Would you be happy to add a relevant one for the note box that the linter is pointing out?

If not, let me know and I'm happy to merge the PR as it is.

AncientNimbus:
> Many thanks! We now have a new style guide rule that all note boxes must have a level 4 heading. Would you be happy to add a relevant one for the note box that the linter is pointing out?
> 
> If not, let me know and I'm happy to merge the PR as it is.

Yes, I am happy to apply the fix, reviewing the lint requirement now.

intermediate_html_css/forms/form_basics.md
```diff
@@ -391,6 +391,8 @@ To create a generic button, we use the button element with a `type` attribute of
 
 <div class="lesson-note lesson-note--tip" markdown=1>
 
+#### Button Type Defaults in Forms
```
mao-sz: Nit: [Style guide](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#case) requires headings to be in sentence case, not title case

```suggestion
#### Button type defaults in forms
```

mao-sz (changes requested):

intermediate_html_css/forms/form_basics.md
AncientNimbus: Thanks for pointing this out, I have applied your suggestion. :)

PR #30003, opened 2025-08-15
Reducing State : Fix grammatical error

xerxes324:
## Because

Grammatical error fix for better clarity.

## This PR

* Fixed a grammatical error. Changed it from "Similarly to the set function in useState..." to "Similar to the set function in useState...".

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/react/more_react_concepts/reducing_state.md
+++ b/react/more_react_concepts/reducing_state.md
@@ -68,7 +68,7 @@ function handleClick() {
 }
 ```
 
-Similarly to the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.
+Similar to the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.
 Keep in mind that these functions use [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) to determine if the state has changed, if it hasn't, then the component won't re-render.
 
 Whether you use `useState` or `useReducer` is up to you, they're both equivalent and you might use them both in the same component.
--- a/react/more_react_concepts/reducing_state.md
+++ b/react/more_react_concepts/reducing_state.md
@@ -68,7 +68,7 @@ function handleClick() {
 }
 ```
 
-Similarly to how React updates state with the set function in useState, React *only updates the state in the next render* after calling the dispatch function.
+Similarly to how React updates state with the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.
 Keep in mind that these functions use [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) to determine if the state has changed, if it hasn't, then the component won't re-render.
 
 Whether you use `useState` or `useReducer` is up to you, they're both equivalent and you might use them both in the same component.
--- a/react/more_react_concepts/reducing_state.md
+++ b/react/more_react_concepts/reducing_state.md
@@ -68,7 +68,7 @@ function handleClick() {
 }
 ```
 
-Similar to the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.
+Similarly to how React updates state with the set function in useState, React *only updates the state in the next render* after calling the dispatch function.
 Keep in mind that these functions use [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) to determine if the state has changed, if it hasn't, then the component won't re-render.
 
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (commented):
Can you explain why "similarly" is incorrect? It seems proper to me and it flows better when I read it than "similar to" does.

That being said I'm **far** from a grammar expert.

mao-sz:
"Similarly to" would be grammatically correct here since "similarly" is an adverb describing "update" later in the sentence, i.e. "it updates similarly to that".

"Similar to" would be correct if it was being used as an adjective, e.g. "Similar to foo, bar is a function." i.e. "bar is similar to foo"

pocketradio:
okay so in my sentence I’m comparing two things (the set function and the dispatch function), not describing how React updates. That’s why the idiomatic form is “Similar to…”

If I wanted to use “Similarly to”, i’d have to rephrase it so the adverb modifies the verb. for instance : “React updates state similarly to how the set function does”. 

so, comparing actions : `similarly` ; comparing things : `similar to`. here we compare two things ( set and dispatch fn ) , hence my argument.

@mao , your foo- bar example is the same setup as mine. both compare two things. You’ve got foo vs bar, I’ve got set vs dispatch. That’s why “similar to” works in my sentence too. 

That being said, i dont want to continue grammar - policing this for something trivial haha, so keep it if you feel it's correct :p

mao-sz:
Edit: I wonder if the following would be extra extra safely clear about the intent of the sentence (and thus its grammar):

> Similarly to how React updates state with the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.

---

What I mean is that the sentence isn't comparing the set and dispatch functions themselves, which is why you wouldn't use "similar to" here. The sentence is comparing how React updates when the set function is called with how it updates when the dispatch function is called. Compare "Similar, React *only updates the state in the next render* after calling the `dispatch` function." with "Similarly, React *only updates the state in the next render* after calling the `dispatch` function."

The current way is grammatically correct and doesn't *have* to be reworded "React updates state similarly to..." (though that's also a valid wording). To me, the current wording reads the most naturally.

wise-king-sullyman:
I really dig ^ if you think that phrasing is better @xerxes324

ManonLef:
@xerxes324 would you like to change the phrasing according to @mao-sz suggestion? I also prefer that wording over the current wording because it makes it more clear that it's about the way React "updates" and takes away the possible interpretation of it being about the functions.

pocketradio:
@ManonLef Sorry, I just realized I hadn’t opened any of these messages since my last comment. And yes, I see what Mao was trying to say about the focus being on how React updates.
I’m okay with his suggestion now since it makes much more sense. Should I create a new PR with the updated phrasing, or handle it another way?

ManonLef:
@xerxes324 it's okay, no worries! You can commit your change to your branch and it should automatically show up here on the PR

pocketradio:
@ManonLef if you don't mind could you tell me how to do it ? i havent made a change to an existing PR commit before sorry :( 
I tried googling it but it didn't make much sense.

ManonLef:
@xerxes324 if you created this branch locally, you can add the changes locally, then commit them locally, then push them. Then they will automatically show up here

ManonLef:
I think you might have edited directly on Github @xerxes324 . If that's the case you can still [edit the file here](https://github.com/TheOdinProject/curriculum/pull/30003/files)

And then click the 3 dots to pull up the below menu which has an edit option. 

Screenshot 2025-09-04 at 19 52 10

pocketradio:
@ManonLef Thanks a lot for the guidance. I've made the changes necessary as per Mao's earlier feedback. I hope it's accurate ; if not do let me know if i need to change anything else :)

react/more_react_concepts/reducing_state.md
```diff
@@ -68,7 +68,7 @@ function handleClick() {
 }
 ```
 
-Similarly to the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.
+Similarly to how React updates state with the set function in useState, React *only updates the state in the next render* after calling the dispatch function.
```
ManonLef:
```suggestion
Similarly to how React updates state with the `set` function in `useState`, React *only updates the state in the next render* after calling the `dispatch` function.
```

I think we should keep the code formatting for these names

ManonLef (changes requested):

ManonLef:
@xerxes324 and thank you for your patience and willingness to make these changes. I have proposed an edit in my review. You can click `commit suggestion` for it to be committed straight away if you want unless you'd rather edit it yourself.

pocketradio:
@ManonLef Oops my bad, forgot to add the formatting. I hit `commit suggestion`, i hope it works on your end. Sorry for the inconvenience caused.

ManonLef:
@xerxes324 no worries, works like a charm! Thank you again!

PR #30004, opened 2025-08-16
Forms and Data Handling: Fix code snippet

mhndckngyn:
## Because
The added lines of code fix an error that occurs when making a request to the app after running it: `Error: Failed to lookup view "index" in views directory ...`

## This PR

- Adds the line `app.set("views", path.join(__dirname, "views"));`, which set views to the directory where the template files are located.
- Adds the line `const path = require("node:path");`, which imports the necessary module for the aforementioned line.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/nodeJS/express/forms_and_data_handling.md
+++ b/nodeJS/express/forms_and_data_handling.md
@@ -210,8 +210,10 @@ Create folders for `routes`, `views`, `controllers`, `storages`, and an `app.js`
 // app.js
 const express = require("express");
 const app = express();
+const path = require("node:path");
 const usersRouter = require("./routes/usersRouter");
 
+app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "ejs");
 app.use(express.urlencoded({ extended: true }));
 app.use("/", usersRouter);


mhndckngyn:
It looks like the lines I added aren't needed if the views directory is in the same location where the app is run. For example, running `node ./app.js`  and making a request won't give an error if views are placed in `./views`. It is necessary when the views directory is nested further though, like `./src/views`.

This is because the Express app's `views` setting has the value `process.cwd() + '/views'` by default: https://expressjs.com/en/api.html#app.settings.table:~:text=process%2Ecwd%28%29%20%2B%20%27%2Fviews

So the error I encountered above only occurs if the views are placed in a subdirectory (`./src/views` in my case). The lesson does not specifically tell learners to put their code in a subdirectory, so the changes I made might not actually be necessary. It could make the setup more flexible however.

mao-sz:
IMO still sensible to include for explicitness and consistency with the Views lesson.

mao-sz (approved):

PR #30043, opened 2025-08-27
Shopping cart: Separate cart features to separate third page

### LINKED ISSUE #30022: Shopping Cart: Clarify cart page instructions
### Checks

### Describe your suggestion

The current project instructions call for "at least two pages" (home + shop(inc. cart)) and later states
> On the shopping cart page, you should have the same navigation bar that displays the number of items currently in the cart. You should also have a button next to it where you can go to the cart to checkout and pay (however we are not going to implement this logic here).

This is slightly confusing/odd intentions, since this suggests needing a home page (with whatever), and a shop page that contains a non-functional cart link/button that shows the current cart item count i.e. do not implement a cart page, and that the cart button/link appears only on the shop page and is not persistent across the site. That's not often what many shop sites do, and doing it this way potentially misses out on practising a situation where state is needed outside of an outlet, to be shared by multiple outlet children.

In my mind, following how many shop sites do things, the instructions ought to call for 3 pages:
1. Home
2. Shop
3. Cart

On all pages, somewhere in the navbar for example, the link to the cart page should also indicate the current cart item count (e.g. a badge or just part of the link text). The cart page itself does not need to have any payment/checkout behaviour implemented, i.e. it's just a mock cart displaying the items etc.
Doing things this way will also force learners to make use of outlet context, or otherwise state defined outside of an outlet and shared between outlet children.

This is how I've seen most people do things, but a couple of people have been confused as to exactly *what* the project is asking for and trying to specify or clarify. Thinking about it, I agree it's a little oddly worded. I think the above approach would be clearer and also more intuitive with how many shop sites do things.

### Path

Ruby / Rails, Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-react-new-shopping-cart

### (Optional) Discord Name

_No response_

_No response_

mao-sz:
## Because

There has been the odd case of confusion regarding exactly what's expected of pages from the Shopping Cart project and features like the cart button showing the contents quantity. I believe this is partly from overspecifying some instructions (as part of a step for only one page), as well as "combining" the shop and cart features into the same page which is not how many shop sites do things.

A 3-page system (home, shop, cart) is more intuitive and consistent with how many other shop sites do things.

## This PR

- Rewords instructions for 3 pages instead of "at least 2", separating the cart stuff to its own page
- Rephrases cart quantity indicator to be more generic and shared across all pages.
- Converts list of host providers to an unordered list. No semantic reason they should be in an ordered list.

## Issue
Closes #30022

## Pull Request Requirements

### DIFF
--- a/react/the_react_ecosystem/project_shopping_cart.md
+++ b/react/the_react_ecosystem/project_shopping_cart.md
@@ -8,12 +8,12 @@ By now you've come far from your React-baby days. You have tools like routers an
 
 1. Create a new React project.
 1. Think about the component and the folder structure. How could you set up your application? Which components or functionalities do you need? It's a good idea to note this down somewhere you can easily get to and refer back and add to it to keep track.
-1. You should have at least two pages (a home page and a shop page, which includes your shopping cart). Let the user navigate between the pages with a navigation bar, which will be shown on both pages.
+1. You should have three pages: a home page, a shop page, and a cart page. Let the user navigate between the pages with a navigation bar, which will be shown on all pages.
 1. To your homepage, you can add whatever you'd like! A few images or information will be totally fine; it doesn't have to be something fancy - it's to test the concepts taught thus far.
-1. On the shopping cart page, you should have the same navigation bar that displays the number of items currently in the cart. You should also have a button next to it where you can go to the cart to checkout and pay (however we are not going to implement this logic here).
-1. Build individual card elements for each of your products. Display an input field on it, which lets a user manually type in how many items they want to buy. Also, add an increment and decrement button next to it for fine-tuning. You can also display a title for each product as well as an "Add To Cart" button.
+1. On the shop page, build individual card elements for each of your products. Display an input field on it, which lets a user manually type in how many items they want to buy. Also, add an increment and decrement button next to it for fine-tuning. You can also display a title for each product as well as an "Add To Cart" button.
+1. When there are items in your cart, the cart page link in the navbar should indicate how many items are in your cart. This should update in real time as the user adds or removes items from their cart.
+1. On the cart page, you should display the items and their quantities, and allow users to increase/decrease the quantity of items in their cart (including removal if appropriate). Put a fake checkout button in as well. This does not need to do anything (we will not be implementing an actual checkout, nor any form submission etc.).
 1. Fetch your shop items from [FakeStore API](https://fakestoreapi.com) or something similar.
-1. Once a user has submitted their order, the amount on the cart itself should adjust accordingly.
 1. Clear out any `missing in props validation` errors in your app!
 1. Make sure to test your app thoroughly using the React Testing Library. Be careful not to test `react-router-dom` directly, since it is an external library and the developers working on it must have tested the library already.
 1. As usual, style your application so you can show it off! You have a host of options provided already.
--- a/react/the_react_ecosystem/project_shopping_cart.md
+++ b/react/the_react_ecosystem/project_shopping_cart.md
@@ -19,25 +19,25 @@ By now you've come far from your React-baby days. You have tools like routers an
 1. As usual, style your application so you can show it off! You have a host of options provided already.
 1. Lastly, it's time to deploy it! Depending on what hosting solution you're using, you may need some additional configuration so that your routing is handled correctly as a single page application (SPA).
 
-   1. **Netlify**: You need to add a `_redirects` file to the `public/` directory of your project. Copy the following to redirect all routes to the index page and let `react-router-dom` handle the rest. You can read more about this at the [Netlify documentation on redirects](https://docs.netlify.com/routing/redirects/).
+   -. **Netlify**: You need to add a `_redirects` file to the `public/` directory of your project. Copy the following to redirect all routes to the index page and let `react-router-dom` handle the rest. You can read more about this at the [Netlify documentation on redirects](https://docs.netlify.com/routing/redirects/).
 
-      ```text
-      /* /index.html 200
-      ```
+     ```text
+     /* /index.html 200
+     ```
 
-   1. **Vercel**: You need to add a `vercel.json` file at the root of your project and copy the following configuration. Similar to Netlify, this redirects all routes to the index page and lets `react-router-dom` handle the rest. More information can be found here at the [Vercel documentation for SPAs and Vite](https://vercel.com/docs/frameworks/vite#using-vite-to-make-spas).
+   -. **Vercel**: You need to add a `vercel.json` file at the root of your project and copy the following configuration. Similar to Netlify, this redirects all routes to the index page and lets `react-router-dom` handle the rest. More information can be found here at the [Vercel documentation for SPAs and Vite](https://vercel.com/docs/frameworks/vite#using-vite-to-make-spas).
 
-      ```json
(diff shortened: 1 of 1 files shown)


react/the_react_ecosystem/project_shopping_cart.md
```diff
@@ -8,36 +8,36 @@ By now you've come far from your React-baby days. You have tools like routers an
...
-1. Build individual card elements for each of your products. Display an input field on it, which lets a user manually type in how many items they want to buy. Also, add an increment and decrement button next to it for fine-tuning. You can also display a title for each product as well as an "Add To Cart" button.
+1. On the shop page, build individual card elements for each of your products. Display an input field on it, which lets a user manually type in how many items they want to buy. Also, add an increment and decrement button next to it for fine-tuning. You can also display a title for each product as well as an "Add To Cart" button.
+1. When there are items in your cart, the cart page link in the navbar should indicate how many items are in your cart. This should update in real time as the user adds or removes items from their cart.
+1. On the cart page, you should display the items and their quantities, and allow users to increase/decrease the quantity of items in their cart (including removal if appropriate). Put a fake checkout button in as well. This does not need to do anything (we will not be implementing an actual checkout, nor any form submission etc.).
```
ManonLef: I am not sure about a checkout button that doesn't function. People will probably be using this site for their portfolio and it might give the wrong idea something is broken. In the current lesson it states: `go to the cart to checkout and pay (however we are not going to implement this logic here).` which could be freely interpreted to include or not include an actual button, but I think the main point there is that clicking, should lead you to the cart page. 

Maybe I'm overthinking this but I know for my project I didn't want to include a non-functioning button. I would consider it if clicking it brought up a message to state this was a portfolio project on which you couldn't actually checkout (but if they wish to donate/support include my PayPal or something 😆 ). 

Would love your thoughts @mao-sz and maybe team input @TheOdinProject/javascript

react/the_react_ecosystem/project_shopping_cart.md
```diff
@@ -8,36 +8,36 @@ By now you've come far from your React-baby days. You have tools like routers an
...
+         {
+           "source": "/(.*)",
+           "destination": "/index.html"
+         }
+       ]
+     }
+     ```
 
-   1. **Cloudflare Pages**: As of the time of writing, unlike Netlify and Vercel, no additional steps are required as the default behaviour will allow `react-router-dom` to correctly handle redirects for SPAs. You can learn more about this at the [Cloudflare documentation on serving pages](https://developers.cloudflare.com/pages/platform/serving-pages/).
+   -. **Cloudflare Pages**: As of the time of writing, unlike Netlify and Vercel, no additional steps are required as the default behaviour will allow `react-router-dom` to correctly handle redirects for SPAs. You can learn more about this at the [Cloudflare documentation on serving pages](https://developers.cloudflare.com/pages/platform/serving-pages/).
```
ManonLef: image
These non-numbered lists now look a bit morse-y with the dash followed by a dot: `-.` I don't think that was your intention.

ManonLef (changes requested):

ManonLef:
Thanks for this @mao-sz , I think it's a lot clearer. I do think the list should be fixed and I left another comment to ponder.

react/the_react_ecosystem/project_shopping_cart.md
mao-sz: Oh oops, silly me. Must have undid a change by accident some point after checking the markdown preview.

react/the_react_ecosystem/project_shopping_cart.md
mao-sz: The checkout button doesn't have to be operational because we're not building an actual checkout. The button I'm referring to isn't a button that brings you to the cart page, it's the button on the cart page itself that actually makes you check out and go through payment. The instructions are to make clear the user is expected to actually make a cart page, it just doesn't have to go further and make an actual checkout. The original instructions could be interpreted as "you don't need to make a cart page, you just need to make a button that *would* go to the cart", which is what motivated this PR.

I'm also not sure the project instructions should be made to accommodate someone who *may* use the project in their portfolio. We've removed mentions of portfolio in previous projects where they were often interpreted as "you should use this for your portfolio, so you should spend more time going out of the project scope".

I think we should let people decide if they want to use any given project in their portfolio, and if they do, they can decide if they spice it up at all and how. The project spec should ideally focus on keeping the scope about learning for the key concepts, and making an actual checkout with payment/submission falls out of that scope (and always was even with the original instructions).

ManonLef: > The button I'm referring to isn't a button that brings you to the cart page, it's the button on the cart page itself that actually makes you check out and go through payment.

I'm pretty sure we're talking about the same thing. I'm not talking about a "going to the cart page" button either, but about a button on said cart page (that is now part of the scope) which would let you check out/pay (but not really). 

> The instructions are to make clear the user is expected to actually make a cart page, it just doesn't have to go further and make an actual checkout. The original instructions could be interpreted as "you don't need to make a cart page, you just need to make a button that would go to the cart", which is what motivated this PR.

I agree that the prior instructions are confusing and that the intentions of this PR are to make it clear if we should or should not add a cart page. I think this PR covers that very well and it's made clear that 3 pages are expected.

Regarding the possible portfolio project; fair enough regarding people's choice on what to include and or polish, but I would like that to not distract from the actual issue I take with it: Why include a button that doesn't work, and even make it a requirement? In a real life project (I know this isn't one) that also wouldn't make sense to me unless the button shipped with the rest of the feature. It also wouldn't make sense for practice purposes at this point.

mao-sz: Do you think it'll be clearer to just not mention the fake checkout button? And just say something like "we're not actually implementing any proper checkout/payment feature, just displaying the cart"? Since that's my intention, so perhaps the "fake checkout button" is overspecifying that could be misinterpreted.
Then it's up to the user as to exactly how they implement that page in general, as long as they know the key expectation

ManonLef: @mao-sz I am leaning towards that yes. It's been too long since I've checked any of the projects. What's your idea? Do people actually add a dummy button? I know [my own project](https://shopping-cart-manonlef.vercel.app/cart) was pretty much how you updated the lesson (3 pages) but no checkout feature or button on that cart page.

It's definitely not a hill I'm willing to die on honestly. I can see how people might want to add it because we see shopping cart pages everyday in real life and we expect that page to have a (call to) action button. Personally, if it were a requirement, I'd disable it or make it show something on click, but the more we talk about it, the more I'm aware it's a personal preference.

mao-sz: Mine does <https://shtor.netlify.app/>
Just a button called "fake checkout" that says "this does nothing" when clicked. I don't think it's a particularly big issue, and if asked, it's easily explained by "the checkout itself wasn't in scope of the learning for that point".

But I agree on the personal preference thing and it seems it'll be more sensible to just not specify anything beyond what the cart page needs, and that no checkout/payment needs to be implemented. Let people decide how they wish to handle that since a specific approach won't be relevant to the practice points.

ManonLef (approved):
Thanks Mao!

PR #30071, opened 2025-09-04
Factory Functions: Lesson tidyup and content clarifications

### LINKED ISSUE #29212: Factory Functions: Suggested style tidy-up and phrasing tweaks
### Checks

### Describe your suggestion

The [factory function lesson](https://www.theodinproject.com/lessons/node-path-javascript-factory-functions-and-the-module-pattern) is a great lesson but understandably tricky given the wealth of info to be taught.

Recently I've noticed a few questions where confusion stems from not quite following or even missing some comments, or otherwise questions about why the lesson's code style differs from other lessons (in case that is a signficant factor).

I think the following tweaks can help streamline the lesson a little:
- Reposition some comments to be *above* the code they directly refer to rather than below
- Split some code blocks so that some "bare" comments become normal lesson text, aiding in visibility and separating examples addressing separate concepts/steps.
- Adjust code style to be a little more consistent with previous lessons (e.g. using a function declaration for constructor rather than traditional func exp assigned to variable).
- Expand very slightly on opening paragraph for the [section on IIFEs](https://www.theodinproject.com/lessons/node-path-javascript-factory-functions-and-the-module-pattern#the-module-pattern-iifes) to help answer the question "well why can't we just write a normal factory function then call it once instead of wrapping it in an IIFE?"

### Path

Ruby / Rails, Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-javascript-factory-functions-and-the-module-pattern

### (Optional) Discord Name

_No response_

_No response_

mao-sz:
## Because

Due to some inconsistencies with code style and usage of comments, some learners miss some things or get confused in the lesson flow when some parts are in normal text and some parts are in longer comments.

There has also been common confusion regarding the point of IIFEs or what IIFEs actually are (perhaps due to focusing immediately on a practical use of IIFEs without giving a more plain example of what an IIFE really just is - people sometimes overthink it's some special specific construct).

## This PR

- Tweaks code blocks
  - Reorganises comments so that key lesson text is now normal text, with those code blocks split for easier following
  - Uses more consistent formatting across lesson (should better match other lessons too)
- Rephrases ESM note box to more clearly link it with the lesson's section topic
- Adds some clarifying content on definition and purpose of IIFE to cover some gaps in the current explanations

## Issue

Closes #29212

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
+++ b/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
@@ -250,9 +250,11 @@ function createPlayer(name, level) {
 
 ### The module pattern: IIFEs
 
-<div class="lesson-note lesson-note--warning" markdown="1" >
+<div class="lesson-note lesson-note--warning" markdown="1">
 
-ECMAScript 6 introduced a new JavaScript feature called "modules" - which are a set of syntax for importing and exporting code between different JavaScript files. While they are important and powerful, they are covered a bit later in the curriculum. We are not talking about them in this section.
+#### ES6 modules
+
+ECMAScript 6 (released in 2015) introduced a new JavaScript feature called "modules", which are a set of syntax for importing and exporting code between different JavaScript files. For now, we will be talking more generally about the module pattern using IIFEs, which you will still see out in the wild. In a later lesson, we will cover using ES6 modules for similar purposes.
 
 </div>
 
--- a/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
+++ b/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
@@ -28,7 +28,7 @@ In ECMAScript 6, the keywords `let` and `const` were introduced. While `var` var
 let globalAge = 23;
 
 // This is a function - and hey, a curly brace indicating a block
-function printAge (age) {
+function printAge(age) {
   // This is a function scoped variable
   var varAge = 34;
 
@@ -59,11 +59,11 @@ Take a while to brew on that example. In the end, it's not some mind-blowing con
 The best way to approach this would be to start with an example - take a look at this piece of code below.
 
 ```javascript
-function makeAdding (firstNumber) {
+function makeAdding(firstNumber) {
   // "first" is scoped within the makeAdding function
   const first = firstNumber;
 
-  return function resulting (secondNumber) {
+  return function resulting(secondNumber) {
     // "second" is scoped within the resulting function
     const second = secondNumber;
     return first + second;
@@ -75,7 +75,7 @@ But we've not seen an example of a "function" being returned thus far - how do w
 
 ```javascript
 const add5 = makeAdding(5);
-console.log(add5(2)) // logs 7
+console.log(add5(2)); // logs 7
 ```
 
 A lot going on, so let's break it down:
@@ -104,7 +104,7 @@ Because of that, constructors have become unpopular in favor of a pattern that i
 These fancy-sounding functions work very similar to how constructors did, but with one key difference - they levy the power of closures. Instead of using the `new` keyword to create an object, factory functions set up and return the new object when you call the function. They do not use the prototype, which incurs a performance penalty - but as a general rule, this penalty isn’t significant unless you’re creating thousands of objects. Let's take a basic example to compare them to constructor functions.
 
 ```javascript
-const User = function (name) {
+function User(name) {
   this.name = name;
   this.discordName = "@" + name;
 }
@@ -113,7 +113,7 @@ const User = function (name) {
 Hey, this is a constructor... then this can be refactored into a factory!
 
 ```javascript
-function createUser (name) {
+function createUser(name) {
   const discordName = "@" + name;
   return { name, discordName };
 }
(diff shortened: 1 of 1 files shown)


javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
```diff
@@ -57,20 +59,23 @@ Take a while to brew on that example. In the end, it's not some mind-blowing con
...
-// being returned, thus far - how do we use it?
+```
 
+But we've not seen an example of a "function" being returned thus far - how do we use it?
+
+```javascript
 const add5 = makeAdding(5);
-console.log(add5(2)) // logs 7
+console.log(add5(2)); // logs 7
 ```
```
ManonLef: Accidentally came by your PR. It could be me but I'm very not charmed by the function naming here. Since this lesson is being worked on, could we consider something else? For the returned function I think `add` or `addSecondNumber` would work better and is about as descriptive as it gets. `resulting` does not sound like/is a conventional js name for a function. 

For the enclosing `makeAdding ` function I'm not sure yet. Some ideas: `makeClosureAdder`, `makeAddFunction` although I'm not a fan of using the function word inside the name but it's very descriptive, `createAdder`

mao-sz: Glossed over this but I agree re: the function names. I also think there's no need for the extra variables - the params themselves are variables scoped to their own functions.

How does this look:
```js
// firstNumber is scoped to the makeAdder function
function makeAdder(firstNumber) {

  // secondNumber is scoped to addSecondNumber
  // This function doesn't need to be named, we're just
  // making it easier to refer to this function later
  return function addSecondNumber(secondNumber) {
    return firstNumber + secondNumber;
  }
}
```

Alternatively, after a more thorough read of this section on closures, I think the section could be simplified and perhaps given a second example at the end that's more practical. This adding function is more appropriate for a simple rundown of the mechanics.

I think part of the simplification could be making clear this is no different from any other use of functions. If you want to create a string based off some args, you write a function that returns a string. Want an array based off args? Same thing - write a function that returns an array. These are all reusable. Want a function based off some args? Write a function that returns a function.

I've found this perspective pretty successful in the server for demistifying closures. I feel this may be a little out of this PR's scope, so if you agree with the above, I think it'll be best to handle that in a separate PR.

mao-sz: Actually I'll have a little deeper think about this section and repurpose this PR as more content clarification with a side of tidyup, rather than the reverse.

mao-sz:
I don't have a huge amount of time right now but plan to rework the currying/closure section (as discussed above) in a separate PR in the next few days, or after this PR gets merged. 
I think it's out of this PR's scope and I'd rather not block these changes with that section.

javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
```diff
@@ -268,6 +292,10 @@ This is where we encounter the word **encapsulation** - bundling data, code, or
...
 
+#### Why the IIFE?
+
+But then why not just write the factory function then call it once? Why bother with the IIFE? Well without the IIFE, we'd have to give the function a name so we can call it afterwards. Now it has a name, it's both taken a name up in that scope and also means it's reusable in that scope. That may not be desirable - we may purposely want a **singleton**. By wrapping the factory in an IIFE, we achieve the same code flow, except we no longer need to name the function, which also means it can't be referenced later. We are packing the code that creates a calculator into what's effectively a module, then exposing only what needs to be used later in the code: the `calculator` object.
```
ManonLef: I wonder if we should introduce the "singleton" here, a term not covered further but with significant meaning in programming. Perhaps reword it to "we may purposely want it to only be invoked once.

javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
ManonLef: Let me know if you wish to edit this still, or otherwise resolve it and keep as is.

javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
mao-sz: While it is a singleton, I guess it makes sense to avoid a terminology rabbit hole since we're not really teaching the design pattern as a main focus
:+1:

javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
ManonLef: @mao-sz I think this slipped through, or if you don't wish to address this now, you can mark as resolved

mao-sz: @ManonLef ah, missed this. I'll address this in a separate PR, so resolving for now.

mao-sz:
Got through some of my backlog so will look to address the currying stuff in this PR :curry:

mao-sz:
@ManonLef had time to address the closure/curried function stuff. Hopefully the update should have more clear naming as well as lessoning the idea that this is some crazy new concept (using a function to create arrays or strings etc. based on args - no reason why this can't work the same for a function instead).

PR #30146, opened 2025-09-16
Intro to React Testing: Provide background context for UI testing

### LINKED ISSUE #29377: React: Improve introduction to React Testing
### Checks

### Describe your suggestion

The lesson called "Introduction To React Testing" is in dire need of an actual introductory paragraph that explains what is going on in the lesson. We start the lesson with introduction that misses the main point of the lesson, which is that we're here to learn how to test the UI components that React creates. This is unlike how we used jest for example which we only explored in terms of testing the functions/classes/methods that we wrote that didn't directly deal with DOM manipulation/ui testing. So this is a big jump and requires a real introduction.
Without that introduction, we are thrown into the paragraph called "Our First Query" blind. We don't know what a query is yet, we don't know why we need it, we don't even understand the first 2 sentences of this section which are: 
>>> First, we’ll render the component using render. The API will return an object and we’ll use destructuring syntax to obtain a subset of the methods required.

To us, 'render' is what we just learned about in class components (well either that or it is a reference to the life cycle of a component in React which also doesn't fit contextually here). That's the only reference we have to rendering. So is that what we're doing here? And why does the first statement say "render the component" as if it is referring to some specific component that we should have been aware of? Shouldn't it say something more general like "To test a React component, we start by rendering it." Or something like that which is more general?

Then looking at the 2nd sentence we have even more mystery to unravel. It says "The API" as if we should know which API it means? (and no we don't know what it means as we haven't been told to read about an "API" in this lesson yet?) and then it says that "we'll use destructuring syntax" but there is no sign of any such syntax in the example provided immediately following this paragraph? Here's the full example:

```js
// App.jsx

const App = () => <h1>Our First Test</h1>;

export default App;
```
```js
// App.test.jsx

import { describe, it, expect } from "vitest";
import { render, screen } from "@testing-library/react";
import App from "./App";

describe("App component", () => {
  it("renders correct heading", () => {
    render(<App />);
    expect(screen.getByRole("heading").textContent).toMatch(/our first test/i);
  });
});
```

So then we're left again wondering what we've missed...

As an introduction to React testing, this section needs a lot of work!
Thanks for your attention.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-react-new-introduction-to-react-testing

### (Optional) Discord Name

_No response_

Please do not think I opened this so I can personally get clarification on this lesson. I've already done so by asking on discord. I opened the issue because I think the lesson needs to be fixed so others can benefit.

mao-sz:
## Because

This is the first lesson on UI testing which is a bit different to prior testing content.
People often struggle to grasp the value of UI testing due to narrower experience with simpler projects where they can manually test the UI, and don't always make the connection between the value of this and non-UI testing.

Instead of diving straight into setup and syntax, it would benefit to provide a little background context for UI testing and its value.

While we're at it, a small mention of `jsdom` will also provide context for how the tests actually work/what they're doing (otherwise it's a bit of a black box when the setup guide tells people to install it but they don't know why). No need to dive deep - any further can be independently explored.

## This PR

- Adds LO/KC items and intro section about UI testing in general
- Briefly mentions purpose of `jsdom`
- Adds comment pointing out case-insensitive text comparison (using i-regex for cleaner syntax)

## Issue

Closes #29377

## Pull Request Requirements

### DIFF
--- a/react/react_testing/introduction_to_react_testing.md
+++ b/react/react_testing/introduction_to_react_testing.md
@@ -39,7 +39,7 @@ Now that we have everything we need, let's briefly go over what some of those pa
 
 ### Our first query
 
-First, we'll render the component using `render`. The API will return an object and we'll use destructuring syntax to obtain a subset of the methods required. You can read all about what `render` can do in [the React Testing Library API docs about render](https://testing-library.com/docs/react-testing-library/api/#render).
+First, we'll render the component using `render` (you can read all about what the function can do in [the React Testing Library API docs about `render`](https://testing-library.com/docs/react-testing-library/api/#render)). We say "render", but you won't see any screen. Back in the RTL setup article just now, one of the packages installed is `jsdom`, which simulates the DOM (including events) in memory without actually laying anything out visually like in a browser, and that allows us to parse its contents for our tests. This is different to and much less complex than other libraries that actually run a real browser environment for more complex end-to-end tests - that's out of scope here.
 
 ```jsx
 // App.jsx
@@ -62,7 +62,6 @@ describe("App component", () => {
     expect(screen.getByRole("heading").textContent).toMatch(/our first test/i);
   });
 });
-
 ```
 
 <div class="lesson-note" markdown="1">
--- a/react/react_testing/introduction_to_react_testing.md
+++ b/react/react_testing/introduction_to_react_testing.md
@@ -6,9 +6,18 @@ We've learned how to test our vanilla JavaScript applications in a previous sect
 
 This section contains a general overview of topics that you will learn in this lesson.
 
+- Why UI testing is valuable.
 - How to set up a React testing environment.
 - How to test UI elements.
 - Understanding snapshot tests.
+
+### UI testing
+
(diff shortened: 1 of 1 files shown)


mao-sz:
@hbar1st would love to hear your thoughts on this if you have the time, since you were the issue author.

mao-sz:
Thanks for your thoughts @hbar1st 
One thing we are always wary of is trying to become the documentation for stuff and otherwise discouraging independent research as needed.

Everyone will understand things in different ways and may well need to look stuff up further, and that's okay and very much encouraged. Personally, I believe what's been added, in the context of the entire lesson (main contents, links, and assignments), is sufficient to cover the things you mentioned. If someone feels they need a deeper explanation of what the `render` method actually does, the lesson already links to docs and they can continue to browse those docs for more info kr research otherwise.

Same with querying (which is a generic term that means the same thing wherever it's used, whether querying the actual DOM, the jsdom in the test, querying a DB etc.). I feel the lesson as a whole is sufficient in that regard, which takes into account the lesson contents and any links/assignments. Again, if people feel they want a deeper dive into querying specifics, they are always encouraged to research independently.

PR #30156, opened 2025-09-20
Javascript Course:  Working with APIs

karimdevelops:
## Because

It seems extremely unprofessional to just expose the keys like this. I highly recommend adding this plugin for environment variables and storing api key there.

## This PR

Add a webpack plugin for adding env variables

## Issue

Closes #XXXXX

## Pull Request Requirements


GaneshS288:
Not a maintainer but considering the upcoming project "weather app" is frontend only there is no possible way to hide api key when learners try to deploy the site.

The only way would be to ask the user to provide their own key which defeats the purpose of site imo.

goran1010:
Agree with @GaneshS288 here.

Considering that the Weather App Project requires learners to deploy their app using GitHub Pages, that means that they will be pushing their entire `dist` folder to GitHub.
Due to the nature of how `dotenv-webpack` works (by injecting those `.env` variables at build time) this means that those variables will still be exposed on the GitHub anyway.
There is also the problem of needing an introduction to what the `.env` file and variables are - which is taught much later in the curriculum.

I also don't think the fact that it's unprofessional to expose public API keys, is a problem.
For one thing, learners are told very clearly of the dangers of doing this - and warned that this is not how it should be done with paid APIs in a real app.
And just in general, I don't think anyone should strive to create professional apps from TOP projects in the first place - it's not the goal of any of the projects in the curriculum (where it's all about practicing concepts they just read about in the lessons prior to those projects).

Mandatory - not a maintainer.

karimdevelops:
> Agree with @GaneshS288 here.
> 
> Considering that the Weather App Project requires learners to deploy their app using GitHub Pages, that means that they will be pushing their entire `dist` folder to GitHub. Due to the nature of how `dotenv-webpack` works (by injecting those `.env` variables at build time) this means that those variables will still be exposed on the GitHub anyway. There is also the problem of needing an introduction to what the `.env` file and variables are - which is taught much later in the curriculum.
> 
> I also don't think the fact that it's unprofessional to expose public API keys, is a problem. For one thing, learners are told very clearly of the dangers of doing this - and warned that this is not how it should be done with paid APIs in a real app. And just in general, I don't think anyone should strive to create professional apps from TOP projects in the first place - it's not the goal of any of the projects in the curriculum (where it's all about practicing concepts they just read about in the lessons prior to those projects).
> 
> Mandatory - not a maintainer.

Great words!

PR #30255, opened 2025-10-17
Intermediate HTML/CSS: Gives Better Explaination for min and max

### LINKED ISSUE #30020: CSS Functions: Better min() explanation
### Checks

### Describe your suggestion

The following explanation of css function `min()` is very confusing and should be improved.

'Focus on this line width: min(150px, 100%); we can make several observations: If there are 150px available to the image, it will take up all 150px. If there are not 150px available, the image will switch to 100% of the parent’s width. In the first case min() selects 150px, since 150px is the smaller (the minimum) between 150px and 100% of the parent’s width; in the second, it chooses 100%. min() behaves as a boundary for the maximum allowed value, which in this example is 150px.'

I would instead rephrase it to something like this:
```min(150px, 100%) function takes 2 arguments, and returns the smaller value. In this case, it checks whether elements container's width is smaller than 150px. If it is, return this value, else return 150px. The returned value will be used for the css property.```

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-intermediate-html-and-css-css-functions

### (Optional) Discord Name

Whisperer

_No response_

### PR DESCRIPTION
**Pull request by Ansh-Kathil**
## Because

it Provided a better explanation for min and max function 

## This PR

Updated the section including min and max function

## Issue

Closes #30020

## Pull Request Requirements

### DIFF
--- a/intermediate_html_css/intermediate_css_concepts/css_functions.md
+++ b/intermediate_html_css/intermediate_css_concepts/css_functions.md
@@ -103,7 +103,7 @@ You are able to do basic math inside a `min()`. For example, `width: min(80ch, 1
 
 ### max()
 
-`max()` works the same way as `min()`, only in reverse, and is like JavaScript's `Math.max()` and Ruby's `Array#max` methods. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a *minimum* allowed value for a property.
+`max()` works the same way as `min()`, only in reverse, and is like JavaScript's `Math.max()` and Ruby's `Array#max` methods. It will select the largest possible value from within the parentheses.
 
 ```css
 width: max(100px, 4em, 50%);

### CODE REVIEW

**Comment by mao-sz** [comment]
@Ansh-Kathil as per our [general contributing guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md#being-assigned-an-issue), please wait until you have actually been assigned by a maintainer before opening a PR for an open issue. I will wait a little longer for the issue author to respond as it has not been that long since my message, and they may have simply missed the original assignment notification.

If they are inactive after some further time, then I'll reassign and review this PR.

File: intermediate_html_css/intermediate_css_concepts/css_functions.md
Comment (mao-sz) [inline intermediate_html_css/intermediate_css_concepts/css_functions.md:1]: Line 67, there's a linter error flagging due to a note box that's missing an accessible heading, which is now required as per our style guide. It's not something you've done wrong but while you're at it, mind helping us resolve it? Suggestion:
```md
<div class="lesson-note lesson-note--tip" markdown="1">

#### calc() example


```
Note the blank lines as the heading will need a single blank line on either side else the linter won't be happy.

intermediate_html_css/intermediate_css_concepts/css_functions.md
```diff
@@ -93,13 +93,19 @@ The above is just an example of how `calc()` can affect a layout, but keep in mi
...
+
+```css
+width: min(150px, 100%);
+```
+
+This checks whether `100%` of the parent element’s width is smaller than `150px`. If the container is narrower than `150px`, the element will take up the full width of the container `(100%)`. Otherwise, it will be `150px` wide.
 
 You are able to do basic math inside a `min ( )` => for example: `width: min(80ch, 100vw - 2rem);`
 
+This is similar to JavaScript’s `Math.min()` function or Ruby’s `Array#min` method — it simply chooses the smallest value from the given options.
```
mao-sz: A couple of changes:
- We don't need a repeat code block. The original code block already clearly shows `min()` in use.
- Little nit, I prefer for the content to open with the connection to JS/Ruby `min()` - just think it flows in a slightly more streamlined way.

```suggestion
The `min()` function works just like JavaScript's `Math.min()` and Ruby's `Array#min` methods. It takes a list of values separated by commas and returns the smallest one.

This checks whether `100%` of the parent element’s width is smaller than `150px`. If `100%` would be narrower than `150px`, the element will take up the full width of the container (`100%`). Otherwise, it will be `150px` wide.

You are able to do basic math inside a `min()`. For example, `width: min(80ch, 100vw - 2rem);` (you don't even need `calc()` in this case).
```

intermediate_html_css/intermediate_css_concepts/css_functions.md
```diff
@@ -93,13 +93,19 @@ The above is just an example of how `calc()` can affect a layout, but keep in mi
...
+
 ### max()
 
 Max works the same way as min, only in reverse. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a *minimum* allowed value for a property.
```
mao-sz: Similar thing to the previous review comment.
```suggestion
`max()` works the same way as `min()`, only in reverse, and is like JavaScript's `Math.max()` and Ruby's `Array#max` methods. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a *minimum* allowed value for a property.
```

intermediate_html_css/intermediate_css_concepts/css_functions.md
```diff
@@ -110,7 +116,9 @@ Consider the following property of a given element:
 width: max(100px, 4em, 50%);
 ```
 
-From this list of given sizes, `max()` will select the largest one. As long as `4em` or `50%` result in lengths longer than `100px`, `max()` will select (the bigger) one of them. If they are smaller than `100px` (maybe as a cause of user's font size preferences, or their browser's window size or zoom level), then `100px` will win out as the largest. You can think of `100px` in this example as a guard value: `width` here won't ever be set to less than `100px`.
+The `max()` function takes a list of values separated by commas and returns the largest one. For example, `max(100px, 4em, 50%)` compares all three values and sets the element’s width to whichever is largest. If `50%`of the parent container is bigger than `100px` and `4em`, the width will be `50%`. If `4em` is larger than the others, it will use `4em`. This ensures that the element’s width never goes below the largest value specified, which is useful for setting a minimum size in responsive designs.
+
+This behavior is similar to JavaScript’s `Math.max()` function or Ruby’s `Array#max` method.
```
mao-sz: Carrying on from the previous (would combine with that review comment but GitHub doesn't like the diff for some reason...).

Also, just removing some redundant sentences (already covered in the previous paragraph).

I also removed the bit about setting minimum boundaries because in my experience in the community, I actually find that that bit is what causes confusion because people mix up max/min and min/max. I feel it's more sensible to just share how the functions work, then let people play around and read about/figure out good uses for them in their own time elsewhere.

```suggestion
The above compares all three values and sets the element's width to whichever is largest. If `50%` of the parent container is bigger than `100px` and `4em`, the width will be `50%`. If `4em` is larger than the others, it will use `4em`.
```

mao-sz (changes requested):
Many thanks - just a couple of changes detailed below, mainly to wording flow.

Ansh-Kathil:
Thanks for the clarification, the changes are ready for review.

intermediate_html_css/intermediate_css_concepts/css_functions.md
```diff
@@ -93,24 +95,21 @@ The above is just an example of how `calc()` can affect a layout, but keep in mi
...
+This checks whether `100%` of the parent element’s width is smaller than `150px`. If `100%` would be narrower than `150px`, the element will take up the full width of the container (`100%`). Otherwise, it will be `150px` wide.
 
-### max()
+You are able to do basic math inside a `min()`. For example, `width: min(80ch, 100vw - 2rem);` (you don't even need `calc()` in this case).
 
-Max works the same way as min, only in reverse. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a *minimum* allowed value for a property.
+### max()
 
-Consider the following property of a given element:
+`max()` works the same way as `min()`, only in reverse, and is like JavaScript's `Math.max()` and Ruby's `Array#max` methods. It will select the largest possible value from within the parentheses. You can think of `max()` as ensuring a *minimum* allowed value for a property.
```
mao-sz: Oops, I missed this in my previous comment about the whole max/min example thing. Once this has been removed, we should be good to merge.
```suggestion
`max()` works the same way as `min()`, only in reverse, and is like JavaScript's `Math.max()` and Ruby's `Array#max` methods. It will select the largest possible value from within the parentheses.
```

PR #30258, opened 2025-10-18
feat(css-intro): add section explaining CSS initial values

### LINKED ISSUE #30187: Foundations: This course should tell students about the initial values of CSS
### Checks

## TLDR
The [CSS potion](https://www.theodinproject.com/paths/foundations/courses/foundations#css-foundations) of the [Foundations path](https://www.theodinproject.com/paths/foundations/courses/foundations) doesn't explain that every CSS property has a default [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Value_processing#initial_value), which may lead students to a incorrect understanding of CSS.

## Explanation
In [this potion](https://www.theodinproject.com/lessons/foundations-growing-and-shrinking#flex-grow) you state that:
> In the following example the flex shorthand has values for flex-shrink and flex-basis specified with their default values.

This made me think that the `flex: 1 1 0%;` was the default value `flex` value for every flex item. Going forward [here](https://www.theodinproject.com/lessons/foundations-alignment#alignment) I couldn't understand why, without the `flex` property on `.item` it wasn't behaving like: `flex: 1 1 0%;`. 
I realize now that in the statement referred above, you are talking about the `flex: 1` shorthand, and not the default value. Then I searched if without `flex` property it has a default value, which lead me to the mozilla.org link. After reading the Mozilla documentation I know that `flex: 0 1 auto;` is the default value for every flex item and I think it should be mentioned in the course.

### Path

Foundations

### Lesson Url

https://www.theodinproject.com/paths/foundations/courses/foundations#css-foundations

### (Optional) Discord Name

_No response_

_No response_

shahabhilash:
## What changed
- Added a new section at the end of intro-to-css.md explaining CSS initial values.
- Provided examples and a link to MDN reference.

## Why
- Helps learners understand default behavior of CSS properties.

Fixes #30187

### DIFF
--- a/foundations/html_css/css-foundations/intro-to-css.md
+++ b/foundations/html_css/css-foundations/intro-to-css.md
@@ -420,5 +420,5 @@ Every CSS property has an **initial value**, which is the default value used by 
 
 Understanding initial values helps you predict how elements will appear before you apply your own styles.
 
-You can check the initial value of any CSS property on [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference)
+You can check the initial value of any CSS property on [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference).
 


mao-sz:
Please [do not open a PR for an issue if you have not been assigned](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md#being-assigned-an-issue). It avoids duplicate work but also work done in a way that doesn't match what a maintainer may want.

PR #30284, opened 2025-10-23
Intermediate HTML/CSS: Better clamp() explanation and example.

Gritnb:
## Because

The values on the example do not reflect any real world scenario for a h1 (320px). Changed it to something more realistic. The use of term "ideal" does not provide clear insight on what the clamp() function actually does.

## This PR

- Changed the example's values.
- Refined the explanation.

## Pull Request Requirements

### DIFF
--- a/intermediate_html_css/intermediate_css_concepts/css_functions.md
+++ b/intermediate_html_css/intermediate_css_concepts/css_functions.md
@@ -127,8 +127,8 @@ h1 {
 ```
 
 1. the minimum value `1.5rem`
-2. the scaling value `5vw`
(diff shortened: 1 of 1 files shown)


mao-sz (changes requested):
Thanks, I think the proposed wording is clearer and with a more appropriate example (I think the original example's intention was to be a width of something as opposed to a font size, but anyway).

Please restore the original `1.` prefixing for the ordered list items, as our style guide enforces lazy list numbering. You can ignore the lint error about a note box heading as that's being addressed in a different PR.

Issue #26387, opened 2023-09-27
Mocking Callbacks And Components: hard to read article about mocking child components


shahrierNafis:
### Describe your suggestion

The second article in the assignments: [mocking child components](https://medium.com/@taylormclean15/jest-testing-mocking-child-components-to-make-your-unit-tests-more-concise-18691ef6a0c2), uses very complex examples, and the code colour theme is black and white.

But this article from dev.to, [Mocking React components (Jest mocking + React part 2)](https://dev.to/peterlidee/mocking-react-components-jest-mocking-react-part-2-2l8j), and the next one in the series [Returning values from mocks (Jest mocking + React part 3)](https://dev.to/peterlidee/returning-values-from-mocks-jest-mocking-react-part-3-3lfn), does a better job covering component mocking. Both use simple examples and are easier to read than the current one 

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-react-new-mocking-callbacks-and-components

### Checks

- [X] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/.github/blob/main/CONTRIBUTING.md)
- [ ] Would you like to work on this issue?

### (Optional) Discord Name

shahrier_nafis

codyloyd:
looking for input from @TheOdinProject/react on this issue.

wise-king-sullyman:
@01zulfi is this something you're still working on?

01zulfi:
Unfortunately, haven't found time to go over this

mao-sz:
I'll mark it for the team triage meetings so it'll get looked at in due course if someone doesn't get the chance to look at it individually by then

Edit: Actually, related to https://github.com/TheOdinProject/curriculum/pull/30541

mao-sz:
See https://github.com/TheOdinProject/curriculum/pull/30541#pullrequestreview-3631448792

ManonLef:
should this remain on the backlog @mao-sz ?

mao-sz:
Technically kind of since I think the article in question could be improved, but I'm also not really sure what yet and haven't had time to look for a replacement, and am not satisfied with the suggested replacements so far.

Given I haven't seen this crop up much at all recently, let's close this and revisit it if someone ends up opening a new issue on it.

PR #30541, opened 2025-12-18
Update link for mocking React components article

zakingslayerv22:
Taylor Mclean’s article is good but it follows some patterns that should not be encouraged in react.

## This PR
The article contains React anti-patterns that were discouraged earlier in the react course.
1. Using props in state - As taught in previous lessons, this is a React anti-pattern that duplicates data.

```jsx
const GroupForm = ({ initialUsers = [] }) => {
//using props in state
const [users, setUsers] = useState(initialUsers);

//blocks of code
};
```

2. Error in code – He defines a `users` prop in `UserManagement`, then in `Form` he references it as `user`.

```jsx
const Form = () => {
   if (formOpen) {
       if (selectedUser) return <UserForm onSubmit={editUser} user={{}} />;
        //referenced it as user here
       else return <UserForm onSubmit={addUser} user={user[selectedUser]} />;
      }
   return null;
};
```

3. Tests are brittle and inconsistent – The delete and edit tests are based on props while the add test is 
based on closure. I think a good test suite should be more consistent.

I think Peter Jacxsens’s article (although more simplistic) https://dev.to/peterlidee/mocking-react-components-jest-mocking-react-part-2-2l8j does a better job with mocking child components

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/react/react_testing/mocking_callbacks_and_components.md
+++ b/react/react_testing/mocking_callbacks_and_components.md
@@ -149,7 +149,7 @@ The other important thing to note is almost all the tests follow a certain patte
 
 <div class="lesson-content__panel" markdown="1">
 
-1. Read this article about [mocking child components](https://medium.com/@taylormclean15/jest-testing-mocking-child-components-to-make-your-unit-tests-more-concise-18691ef6a0c2). It extensively covers the "how" of mocking child components. (You might need to sign up to read)
+1. Read this article about [mocking react components]([https://medium.com/@taylormclean15/jest-testing-mocking-child-components-to-make-your-unit-tests-more-concise-18691ef6a0c2](https://dev.to/peterlidee/mocking-react-components-jest-mocking-react-part-2-2l8j)). It covers the "how" of mocking child components. (You might need to sign up to read)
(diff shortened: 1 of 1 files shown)


mao-sz (changes requested):
Thanks for the suggestion. There is already an open issue on this https://github.com/TheOdinProject/curriculum/issues/26387 with similar suggestions.

I do agree the current article is a bit tricky to follow, especially with the lack of syntax highlighting. However, I feel like the proposed replacement is a little *too* simplistic, to the point where I can see learners taking what it says as absolute, like the idea that you must only ever unit test components and never actually render any child components. I do see people focusing way too much on the idea of "unit testing" or "integration testing" as categories rather than thinking about what they're testing in the moment and why they're testing it (which may involve more or less isolated tests depending on the situation).

I also don't like how it showcases tests that assert that child components' contents are not rendered, which is definitely not what I'd be writing tests for; those tests are superfluous and only really test that that specific test case successfully mocks a certain component. It says nothing about other tests nor anything about the user-facing behaviour of the site, which is the real point of us testing things.

I'm definitely up for having the first assignment article replaced but I'm not in any rush to do so, and not with the proposed replacement. Happy to review an alternative though, or if you're not up for this then you can close the PR (which is also fine).

PR #30636, opened 2026-01-06
Intro to HTML and CSS lesson: Add CSS Chaining Selectors Example with Type (Element) Selectors

Gabe013:
- Clarify the use of type selectors in a CSS chain selector
- Give erroneous examples with a brief description
- Share a valid example of a properly chained type (element) selector in CSS

## Because

I did not find an example and description of how to actually chain type (element) selectors in a chain of CSS selectors, so I thought other will find it useful when reading the section.

## This PR

* Adds paragraphs of text giving examples of erroneous and correct type (element) selector chaining in CSS
* Provides code snippets for all examples

## Pull Request Requirements

### DIFF
--- a/foundations/html_css/css_foundations/intro_to_css.md
+++ b/foundations/html_css/css_foundations/intro_to_css.md
@@ -215,7 +215,7 @@ For example, this will not work:
 }
 ```
 
-Nor will this if your intention is to chain selectors which include a type (element) selector (as mentioned above, the chaining selector syntax *does not* include white/empty spaces between selectors):
+Nor will the code below if your intention is to chain selectors which include a type (element) selector (as mentioned above, the chaining selector syntax *does not* include white/empty spaces between selectors):
 
 ```css
 .subsection.header div {


mao-sz:
Thanks for the suggested additions. I'm going to close this because given the previous section mentioning what's likely intuitive enough to most people in that you can't chain two type selectors together (because divp is just one word so just means "a divp element", and also how can something be two different elements at the same time anyway?), it should be intuitive enough that there's no way the computer can know that `.headerp` is a class and a type vs just a class called `headerp`. At least in that the lesson doesn't need to go out of its way to teach that, since I would think most who may consider that would quickly work out the problem once they try it for themselves.

It also wouldn't be sensible to talk about the descendent combinator there before we actually introduce it properly, but given the above point, it's not a relevant factor anyway.

Gabe013:
> Thanks for the suggested additions. I'm going to close this because given the previous section mentioning what's likely intuitive enough to most people in that you can't chain two type selectors together (because divp is just one word so just means "a divp element", and also how can something be two different elements at the same time anyway?), it should be intuitive enough that there's no way the computer can know that `.headerp` is a class and a type vs just a class called `headerp`. At least in that the lesson doesn't need to go out of its way to teach that, since I would think most who may consider that would quickly work out the problem once they try it for themselves.
> 
> It also wouldn't be sensible to talk about the descendent combinator there before we actually introduce it properly, but given the above point, it's not a relevant factor anyway.

Hey, mao-sz!

Thank you for your quick reply!

Yes, I agree that my example with botched class selectors written as `.subheadingdiv` or `#headerp` might not be necessary, but I didn't find it intuitive to go ahead and figure it out myself: "Hey, look, I have to write the type (element) selector before a class or ID selector for this to work." Especially if I am reading the lesson for the first time and learning about CSS just then.

I personally believe that adding an example with a properly written chained selector which includes a type (element selector) along with a class and/or ID selector would not extended the lesson too much and that most learners would actually prefer to see it (see my thoughts in the paragraph above)

Yes, I thought about the implications of "throwing out" an example with a descendant combinator during submission without it being previously described directly, but I just couldn't think of a better way in this case and at that moment. At least I did not mention descendant combinators in my example, although I could have and then referenced it as something coming up in the next subheading/section.

Happpy New Year, all the best, thanks for the response! :) 🥳 🌟 🎄 🎅🏻 

Edit: Extended the 2nd paragraph mentioning the other selector types I would add in a chained CSS selector

mao-sz:
To clarify, when I say "intuitive", I don't mean that most will just know it like that, but that if most encounter the thought (of something like `.headerp`), they'd be able to reason it out, exactly like you were able to. Especially following the bit already about chaining multiple type selectors, since it's the exact same principle (how will the browser know it's meant to be two selectors not one?).

For me personally, I'm a little stricter on the criteria for making changes to a lesson in that something should be fixing a clear and unreasonable pedagogical problem in the lesson, like misleading info, paragraph seriously hard to follow, significant missing concept that's core to the topic, etc. 

In honesty, this is the first time I've seen this particular problem raised (whether on here or in the Discord community). Not saying that's abad reflection on you btw. There are always loads of things an individual might not grasp easily or not realise for a while, but if we added notes and clarifications and warnings for every one of those situations, our lessons would be insanely long and bloated. Hence as maints, we have to assess whether a suggestion actually solves an unreasonable pedagogical problem, and IMHO, there's not enough of an unreasonable pedagogical problem here. If the lack of mention of this in the lesson caused many more learners to have unreasonable struggle when going through, and so it came up a lot more and couldn't easily be resolved by a "think about it, try it and see what happens", then sure, that suggests an unreasonable pedagogical problem that needs resolving.

---

None of this is to discourage any potential further contributions, by the way! Just thoughts on this PR alone. Though I would suggest for stuff like this in the future, opening an issue first so things can be discussed with maints before you potentially do any work that won't be accepted.
Direct-to-PR, as per the contributing guide, is of course perfectly fine for small, clearly justifiable changes like typo fixes, broken links etc.

Gabe013:
> To clarify, when I say "intuitive", I don't mean that most will just know it like that, but that if most encounter the thought (of something like `.headerp`), they'd be able to reason it out, exactly like you were able to. Especially following the bit already about chaining multiple type selectors, since it's the exact same principle (how will the browser know it's meant to be two selectors not one?).
> 
> For me personally, I'm a little stricter on the criteria for making changes to a lesson in that something should be fixing a clear and unreasonable pedagogical problem in the lesson, like misleading info, paragraph seriously hard to follow, significant missing concept that's core to the topic, etc.
> 
> In honesty, this is the first time I've seen this particular problem raised (whether on here or in the Discord community). Not saying that's abad reflection on you btw. There are always loads of things an individual might not grasp easily or not realise for a while, but if we added notes and clarifications and warnings for every one of those situations, our lessons would be insanely long and bloated. Hence as maints, we have to assess whether a suggestion actually solves an unreasonable pedagogical problem, and IMHO, there's not enough of an unreasonable pedagogical problem here. If the lack of mention of this in the lesson caused many more learners to have unreasonable struggle when going through, and so it came up a lot more and couldn't easily be resolved by a "think about it, try it and see what happens", then sure, that suggests an unreasonable pedagogical problem that needs resolving.
> 
> None of this is to discourage any potential further contributions, by the way! Just thoughts on this PR alone. Though I would suggest for stuff like this in the future, opening an issue first so things can be discussed with maints before you potentially do any work that won't be accepted. Direct-to-PR, as per the contributing guide, is of course perfectly fine for small, clearly justifiable changes like typo fixes, broken links etc.

Hey, mao-sz!

I didn't really figure it out on my own, I tried this and that, and then I asked ChatGPT for assistance, which provided me with the accurate code/syntax (I know that this is not advised by the Odin Project, sorry! :( )

Thanks for the feedback, I see where you're coming from with this and where the problem lies in my PR request for this change.

I will be more mindful and slower to act for any future contribution ideas (should I have them) and follow the steps you've described.

P.S. Big thanks for clarifying what a small change via PR means in the context of Curriculum updates, I will keep that in mind!

PR #30677, opened 2026-01-18
Update object_basics.md: fix grammar

Goal-max:
Improve clarity and understanding to the reader.

## Because

Paragraph reworded as it was difficult to understand. Help explain concepts to the learner.

## This PR

Reworded the paragraph.

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/foundations/javascript_basics/object_basics.md
+++ b/foundations/javascript_basics/object_basics.md
@@ -60,7 +60,7 @@ const element = document.querySelector("#container");
 element.style.backgroundColor = "red";
 ```
 
-We're changing the variable we declared (`element`), yet the changes affect the corresponding node in the DOM. Why does it happen? It's because the node we have in our code is a **reference** to the same node that our DOM uses. If (`element`) wasn't a reference, but instead was a copy (like primitive data types), our changes would have **no** effect! This is because the changes would be made to the local copy we have only.
+We're mutating the variable we declared (`element`), yet the changes affect the corresponding node in the DOM. Why does it happen? The node we have in our code is a **reference** to the same node that our DOM uses. If `element` wasn't a reference, but instead was a copy of the node itself (like how primitive data types behave), mutating it would have **no** effect because changes would only affect the copy, not the original.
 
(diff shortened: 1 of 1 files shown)


foundations/javascript_basics/object_basics.md
```diff
@@ -60,7 +60,7 @@ const element = document.querySelector("#container");
 element.style.backgroundColor = "red";
 ```
 
-We're mutating the variable we declared (`element`), yet the changes affect the corresponding node in the DOM. Why does it happen? That's because the node we have in our code is a **reference** to the same node that our DOM uses. If that wasn't a reference, but a copy like primitive data types behave, our changes would have **no** effect! Because the changes would be made to the local copy we have.
+We're changing the variable we declared (`element`), yet the changes affect the corresponding node in the DOM. Why does it happen? It's because the node we have in our code is a **reference** to the same node that our DOM uses. If (`element`) wasn't a reference, but instead was a copy (like primitive data types), our changes would have **no** effect! This is because the changes would be made to the local copy we have only.
```
mao-sz: I do think the are some grammar improvements to be made in this paragraph, but there are some counter-suggestions below:
- A few general grammar tweaks to yours
- It's very important we use the verb "mutate" here and not "change", because "change" (as a verb) is often used by people to casually describe both mutation and reassignment, which are completely different mechanisms. e.g. "How come we can push to that const array? I thought const means you can't change it!"
  "Change"/"changes" is fine when used as a noun here because it's already been established that they're referring to changes made specifically by mutation, but they must not be used as a verb here.

```suggestion
We're mutating the variable we declared (`element`), yet the changes affect the corresponding node in the DOM. Why does it happen? The node we have in our code is a **reference** to the same node that our DOM uses. If `element` wasn't a reference, but instead was a copy of the node itself (like how primitive data types behave), mutating it would have **no** effect because changes would only affect the copy, not the original.
```

mao-sz (changes requested):

PR #30704, opened 2026-01-21
Git Basics: Removes outdated information

sussynk:
## Because
Git now uses "main" as default branch name instead of "master", information no longer relevant.

## This PR
* Removes useless information no longer required.

## Issue
Git now uses "main" as default branch instead of "master" making the curriculum information not relevant anymore and redundant command to change the default branch from "master" to "main".

## Additional Information
Source:
https://github.com/github/renaming

## Pull Request Requirements

### DIFF
--- a/git/foundations_git/git_basics.md
+++ b/git/foundations_git/git_basics.md
@@ -18,9 +18,6 @@ This section contains a general overview of topics that you will learn in this l
 
 - Github recently updated the way it names the default branch. This means you need to make sure you are using a recent version of git (2.28 or later). You can check your version by running:
   `git --version`
-- If you haven't already, set your <span id="main-push"></span>local default Git branch to `main`. You can do so by running:
-  `git config --global init.defaultBranch main`
-- For more information on the change from `master` to `main` see [GitHub's Renaming Repository](https://github.com/github/renaming).
 
 #### Create the repository
 


clemedev:
I have git configured before this, perhaps that's why.

Anyways, I find this method would be easier to setup for many people instead of the current SSH key method:
https://cli.github.com/manual/gh_auth_login

Current method might be confusing for many people, the link above is much simpler:
1. sudo apt install gh
2. gh auth login
3. follow on screen instruction and redirect to web browser and login with your GitHub account
4. boom done!

Much simpler and lesser step compared to current method, perhaps I should add this method into the instructions? What do you think.

clemedev:
To follow up on the .gitconfig question, no I don't find any .gitconfig file as well, perhaps it's due to my setting up method is different than the guide

JoshDevHub:
I'm going to close this because git hasn't changed its default yet (even if GitHub has).

The `gh` CLI is a separate issue, although it's one we're probably not interested in accepting.
1. We tell people in many different places to `git clone <ssh_link>`, and just because you're authenticated with the `gh` CLI, that does not magically make that process work. You'd still need an SSH key to clone that way. In order to use the `gh` CLI, we'd have to adjust every place in the curriculum where we tell learners how to interact with remote repositories.
2. Interfacing with GitHub through the `gh` CLI means you now have to learn two tools: `git` for working locally and `gh` for when you want to work with a remote. There's a good case to be made that on a longer timeline, that isn't actually any simpler, even if the initial set-up is simpler.
3. Learning how to use SSH to authenticate with a host and plain `git` to clone, push, pull, etc. from that host means your knowledge is more portable. You're not locked into GitHub's way of doing things, and you could easily work with other source code repository hosts using the same techniques.

PR #30713, opened 2026-01-23
Both Courses: Computer Science: Modify instructions to the BST project

### LINKED ISSUE #30696: Update Binary Search Tree instructions to prepare for testing requirements
Update the Binary Search Tree project requirements for stronger consistency and to prepare for future testing requirements.

JoshDevHub:
## Because

We want to adjust the instructions to make testing easier to add in the future.

Because there are a lot of small, disconnected changes, I expand on my reasoning for them during the next section. Hopefully it's not a huge headache to review. Glad to answer any questions.

### The JS Lesson:
- Use better JavaScript-isms (functions that return `undefined` and 0-arity function names appended with `()`)
- Get rid of the excessive `null` checks in `prettyPrint()`.

### The Ruby Lesson:
- Remove `Comparable` mention in Ruby. You're mostly comparing nodes to plain values, and although you *can* sorta kinda make that work with `Comparable`, I think cross-type comparisons are largely an antipattern (and so do the designers of the Ruby language as expressions like `"1" > 0` will raise an error). It's also difficult to achieve communicative ordering (ie. `Node.new(1) > 0` needs to be the same as `0 < Node.new(1)`) without monkey patching or wading into Ruby's `coerce` protocol. I just don't think this is a good idea.
- Use a keyword argument on the `is_left` flag in `#pretty_print`. Mostly doing this because rubocop doesn't like using positional args for flags (and for good reason).
- In the traversal methods, instruct learners to return an Enumerator any time a block isn't passed in. This is a more Ruby-ish way to handle things, and, given recent changes to the Custom Enumerables project, learners have now been exposed to this concept.

### Both
- Pull some things that were bolded "tips" into noteboxes. One of these I just eliminated because I didn't think it warranted being a tip.
- The previous `find` operation is something I didn't like much. Nodes are an implementation detail in these structures, and aren't meant to be directly exposed. I've replaced it with an `includes()`/`#include?` operation. I've also moved this to the first thing learners build after getting the tree up and running. This is just because I think it's the simplest operation and works better for getting people used to the concepts compared to going straight into insertion and deletion. And because trees can be constructed with nodes upfront, there's no reason to require insertion before this.
- Split the list items in the assignment for insert and delete. I just think this makes things flow better, especially because I add some instructions for what needs to happen if a given value is or isn't in the tree. This has a consequence of the [YouTube video on insert/delete not fitting well](https://www.youtube.com/watch?v=wcIRPqTR3Kc), but I figured the GeeksForGeeks articles would be enough.
- For the traversal methods, instruct passing values rather than nodes into the callback (JS) or block (Ruby)

## Issue

Closes #30696 

## Pull Request Requirements

### DIFF
--- a/javascript/computer_science/project_binary_search_trees.md
+++ b/javascript/computer_science/project_binary_search_trees.md
@@ -14,7 +14,7 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
 1. Build a `Tree` class/factory which accepts an array when initialized. The `Tree` class should have a `root` attribute, which uses the return value of `buildTree()` which you'll write next.
 
-1. Write a `buildTree(array)` function that takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `buildTree()` function should return the level-0 root node.
+1. Write a `buildTree(array)` function that takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `buildTree()` function should return the level-0 root node. You can make this function private (either through using class syntax privacy features or by not including it in your factory's return object), and just call it for initializing the value of the `root` node.
 
    <div class="lesson-note lesson-note--tip" markdown="1">
 
--- a/ruby/computer_science/project_binary_search_trees.md
+++ b/ruby/computer_science/project_binary_search_trees.md
@@ -14,7 +14,7 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
 1. Build a `Tree` class which accepts an array when initialized. The `Tree` class should have a `root` attribute, which uses the return value of `#build_tree` which you'll write next.
 
-1. Write a `#build_tree` method which takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `#build_tree` method should return the level-0 root node.
+1. Write a `#build_tree` method which takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `#build_tree` method should return the level-0 root node. You can make this method `private` and call it in the initializer to set the value of `@root`.
 
    <div class="lesson-note lesson-note--tip" markdown="1">
 


javascript/computer_science/project_binary_search_trees.md
```diff
@@ -12,28 +12,33 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
...
+
+   #### Visually representing your tree
 
-   ```javascript
+   If you would like to visualize your binary search tree, here is a `prettyPrint()` function that will `console.log` your tree in a structured format. This function will expect to receive the root of your tree as the value for the node parameter:
+
+  ```javascript
    const prettyPrint = (node, prefix = '', isLeft = true) => {
-     if (node === null) {
+     if (node === null || node === undefined) {
```
JoshDevHub, line 27: I didn't really want to enforce any particular implementation, and we specifically don't say what an empty tree should be realized as. So that's why I check both `null` or `undefined`.

javascript/computer_science/project_binary_search_trees.md
```diff
@@ -12,28 +12,33 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
...
+1. Write an `includes(value)` function that accepts a value and returns `true` if the given value is in the tree. If the value isn't in the tree, it should return `false`.
+
+1. Write an `insert(value)` function that accepts a value and inserts a new node with that value into the tree. Be sure to insert in a way that preserves the "binary search" property: for each node, every node to its left must have a lower value, and every node to its right must have a greater value. If the function is called with a value that already exists in the tree, the function should do nothing. If you need additional resources, check out the [Geeks for Geeks article on binary search tree insertion](https://wwww.geeksforgeeks.org/insertion-in-binary-search-tree/?ref=lbp).
```
scriabin-enjoyer: there's an extra "w" in the link here (links to "wwww.geeksforgeeks...") lol
:P

javascript/computer_science/project_binary_search_trees.md
```diff
@@ -43,17 +48,25 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
    </div>
 
-1. Write a `find(value)` function that returns the node with the given value.
+1. Write a `deleteItem(value)` function that accepts a value and removes it from the tree. You'll have to deal with multiple cases for this based on how many children the targeted node has. If the given value doesn't exist in the tree, the function should do nothing. If you need additional resources, check out the [Geeks for Geeks article on deleting in a binary search tree](https://wwww.geeksforgeeks.org/binary-search-tree-set-2-delete/?ref=lbp).
```
scriabin-enjoyer: same thing here with the extra "w"

Also in the ruby file, lines 39 and 49

javascript/computer_science/project_binary_search_trees.md
```diff
@@ -12,28 +12,33 @@ You'll build a balanced BST in this assignment. Do not use duplicate values beca
 
 1. Build a `Node` class/factory. It should have an attribute for the data it stores as well as its left and right children.
 
-1. Build a `Tree` class/factory which accepts an array when initialized. The `Tree` class should have a `root` attribute, which uses the return value of `buildTree` which you'll write next.
+1. Build a `Tree` class/factory which accepts an array when initialized. The `Tree` class should have a `root` attribute, which uses the return value of `buildTree()` which you'll write next.
 
-1. Write a `buildTree(array)` function that takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `buildTree` function should return the level-0 root node.
+1. Write a `buildTree(array)` function that takes an array of data (e.g., `[1, 7, 4, 23, 8, 9, 4, 3, 5, 7, 9, 67, 6345, 324]`) and turns it into a balanced binary tree full of `Node` objects appropriately placed (don't forget to sort and remove duplicates!). The `buildTree()` function should return the level-0 root node.
```
mao-sz: Wonder if it'll help learners if we specify this as `` Write a private `buildTree(array)` function ...``? Hasn't come up recently but I recall a fair few times people don't consider this architecture due to unfamiliarity, and then wonder why make a separate standalone `buildTree` to be used in a constructor or similar.

JoshDevHub: Yeah this is probably a decent idea. Especially with testing, it sends the signal that this method doesn't need its own test cases and learners can just focus in on making sure `constructor()`/`#initialize` work right.

I'll add this in.

JoshDevHub:
@mao-sz How does the last commit read to you?

PR #30784, opened 2026-02-10
Fix CSS Tricks flexbox guide link

Yurii214:
## Summary
- update the CSS-Tricks "Guide to Flexbox" URL in the Foundations alignment lesson

## Testing
- not run (docs change)


mao-sz:
Please make sure you check open issues and PRs in case of duplicates, which is one of the checkboxes you'll have had to check but deleted instead. There is already an open issue and PR for this

PR #30788, opened 2026-02-10
Update SOLID presentation link

Yurii214:
## Summary
- replace the SOLID Slideshare link with the PDF link

## Testing
- not run (docs change)


mao-sz:
Since this isn't the first instance, if you wish to contribute to public projects then it would be a good idea to read their [contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md) if they have one (which we do) to ensure to follow the correct process.

There is an open issue on this that has not been approved for work yet, please do not open PRs trying to close whatever issues you can find without having first been assigned by a maintainer.

PR #30792, opened 2026-02-12
Update text_editors.md

amlanmk:
1.There were Issues while installing from the existing method :  " Unable to locate package code_1.109.2-1770755809_amd64.deb"
2. The speed of downloading the package was way too slow.

Installing via snap , solves both the above problems and is fairly simpler and easy

## Issue

Closes #XXXXX

## Pull Request Requirements

### DIFF
--- a/foundations/installations/text_editors.md
+++ b/foundations/installations/text_editors.md
@@ -27,13 +27,14 @@ Choose your Operating System:
 
 <summary class="dropDown-header">Linux</summary>
 
-#### Step 1: Download VSCode
+#### Step 1: Download VSCode via Snap Store 
 
 - Open your **Terminal**.
 - Run the following command to download the latest **VSCode** `.deb` package:
 
 ```bash
-wget -O code-latest.deb 'https://code.visualstudio.com/sha/download?build=stable&os=linux-deb-x64'
+sudo apt update
+sudo apt install snapd
 ```
 
 #### Step 2: Install VSCode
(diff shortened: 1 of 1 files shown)


ric-projects:
> Installing via snap , solves both the above problems and is fairly simpler and easy...

Snap (and other universal package managers/containers like Flatpak or AppImage) are known to not have good system integration, or at least not to the same level as official system packages. Which can cause issues later when having to deal with other tools, and even more so when we're dealing with multiple files (as most projects usually are).

From what I remember all packages in the curriculum are recommended to be installed as system packages, which avoids those possible integration issues.

Probably a good idea to be aware of that, and perhaps even change your snap package to avoid possible future issues.

zachmmeyer:
I wonder if we should instead provide the documentation https://code.visualstudio.com/docs/setup/linux
Or maybe update the instructions via the documentation.
But yeah I don't think the solution is to use Snap, sorry.

zachmmeyer:
Also I've only seen this exact problem show up two times on the Discord, and now this one.

I don't think the majority of users are having this issue. @amlanmk if you can pinpoint the cause of the issue that would be super helpful, I can at least try to replicate on my Xubuntu VM, but in reality I don't see us changing the instructions.

amlanmk:
yeah i figured out the issue , if we use the " sudo apt update && sudo apt upgrade " before trying to install Vscode , it works absolutely fine , if we dont , in some cases it returns with "unable to locate package issue"

zachmmeyer:
That was going to be my guess, so the fix is pretty easy to implement and our instructions are too. Thanks!

zachmmeyer:
@amlanmk Do you want to modify this PR to instead have the `sudo apt update` before the `sudo apt install ./code-latest.deb`? That way you can still contribute to this lesson? Or do you want to leave that for someone else to do.

amlanmk:
@zachmmeyer i did add a comment to do the same , as am quite new to this , i dont quite understand what PR is and how to modify it . I hope to keep contributing to the lesson likewise , in the future

zachmmeyer:
I'll open a new PR that'll close this one.

Issue #30465, opened 2025-11-26
Evaluate value of individual additional resources across curriculum (grouped by section)


mao-sz:
This issue is to help with tracking and audit trail, and is not to be assigned to anyone.

---

Action as a result of discussion started in https://github.com/TheOdinProject/curriculum/discussions/29326. We do not have a clear idea for what constitutes an "additional resource" and thus the sections have often ended up bloated with things like multiple alternative wordings of the exact same content, only tangentially related stuff, or content that goes well beyond the necessary scope of the lesson. The few things that have been genuinely useful have almost always been better served being part of the lesson itself, rather than an additional resource.

### Additional Resources project plan

Team to go through the curriculum section by section. For each lesson that has additional resources, decide what to do with each item.

Possible actions:
- Integrate within main lesson content
- Move to a different lesson and integrate within its main content
- Drop
- Something else that suggests some middle ground where it's valuable enough to mention but not able to be integrated in a lesson's main content in any way

After this has concluded, we will know whether we have a need for an Additional Resources section. If we still have resources not integrated into main lesson content but also not dropped, we'll be able to develop stricter criteria for what constitutes an "Additional Resource".

> [!Important]
> Us deciding to drop a particular resource does not necessarily mean we think the resource is bad/should not be used. We will take into account the whole lesson as well as its place in the entire curriculum. If an additional resource provides something **new *and* significantly valuable** to the pedagogy, then we will prioritise trying to integrate it into the lesson (whether that's through an addition or replacing something).
>
> No matter how good something is by itself, if the lesson already sufficiently covers the learning material, then there is no significant need to integrate that particular additional resource, and it can therefore be dropped. There will always be many great resources for a given topic but we are not a compendium for related resources.
>
> Learners are still always encouraged to do independent research as needed, such as if they're seeking additional examples/rephrasing. We just don't necessarily need to link every possible resource that could be found helpful to an individual.

This issue is the main overview of the current project phase and will be considered complete after all desired work across the entire curriculum has been completed. Sub-issues will be created as we go through the courses.

PR #30814, opened 2026-02-16
Working with APIs: add additional resource

rgb-katana:
## Because
I think this video will be really helpful for beginners, and this lesson doesn't have any additional resources whatsoever. Additionally, the author of the video has created content on other topics that are covered in this section, so this resource may serve as a helpful supplement for readers.

## This PR
Adds an additional resource block for the "Working with APIs" section.

## Issue
No related PR

## Pull Request Requirements

### DIFF
--- a/javascript/asynchronous_javascript_and_apis/working_with_apis.md
+++ b/javascript/asynchronous_javascript_and_apis/working_with_apis.md
@@ -234,3 +234,9 @@ The following questions are an opportunity to reflect on key topics in this less
 - [What is an API?](#apis)
 - [How is access to an API restricted?](#apis)
 - [How do you fetch and extract data from an API?](#fetching-data)
+
+### Additional resources
+
+This section contains helpful links to related content. It isn't required, so consider it supplemental.
+
+- This video on [Fetch API by ColorCode](https://www.youtube.com/watch?v=ubw2hdQIl4E) takes an in-depth look into how Fetch API works.


mao-sz:
Thanks for the suggestion.
Learners are always encouraged to independently research as needed, and we do not want our lessons to become compendiums for every possibly related/helpful alternative resource out there (which is what some of our lessons ended up becoming to a degree); naturally, everyone is going to come across and click better with different resources.

We are actually going through a large-scale sweep on the curriculum to re-evaluate our additional resources (https://github.com/TheOdinProject/curriculum/issues/30465) which has ended up in a lot of them being removed (and any thst weren't removed were integrated into the lesson itself to fill a gap, so the additional resources section always got removed anyway). We're taking a stricter approach to determining if a lesson *needs* a certain resource mentioned or not.

Is there an actual problematic gap in the lesson's pedagogy that this resource will fill?
Or is this more of a thing that you personally found helpful in your independent research and so just wanted to put it out there as an alternative in case others found it helpful too?

rgb-katana:
> We are actually going through a large-scale sweep on the curriculum to re-evaluate our additional resources (#30465) which has ended up in a lot of them being removed (and any thst weren't removed were integrated into the lesson itself to fill a gap, so the additional resources section always got removed anyway). We're taking a stricter approach to determining if a lesson _needs_ a certain resource mentioned or not.

I saw that in the commit history. I actually believe this video link is needed here as an additional resource. I noticed you've kept an article in the Additional Resources section of the [Keys In React](https://www.theodinproject.com/lessons/node-path-react-new-keys-in-react#additional-resources). Personally, I found that article incredibly insightful and doubt I would have discovered it through my own research. That's exactly what inspired me to create this PR - this video made the same impression on me, and I wondered why it isn't shared on The Odin Project.
  
> Is there an actual problematic gap in the lesson's pedagogy that this resource will fill? Or is this more of a thing that you personally found helpful in your independent research and so just wanted to put it out there as an alternative in case others found it helpful too?

In this lesson, there's a section about fetching weather data via Visual Crossing, and the creator in the video does the same thing but with a government API. I think it would be valuable for readers to see this content. Additionally, the author has excellent videos on async/await and promises, which would serve as incredible supplementary material for learners in this section.

To answer your question directly: I don't think there's an actual problematic gap in the lesson's pedagogy. However, I am 100% confident this channel should be mentioned on The Odin Project, and placing it in this lesson is the best spot. Once readers watch this video, they'll discover the author's other related content organically.

mao-sz:
Thanks for elaborating. I'll review the contents when I get more time.

Just a heads up that we have not yet gotten to the React course yet, so any additional resources there haven't been "left there". All lessons that the team have gone through so far have had their additional resources section removed, with only a few resources being integrated in some way. As of yet, we have not found a sensible criteria for what would constitute an additional resource, meaning it's somehow significant enough to the lesson yet not something that can be integrated with the lesson itself (e.g. due to already covering enough ground).

Essentially, consider this a "soft lock" on additional resources for the time being.

Even if a resource is helpful, the same argument can be made for many other resources on the same thing out there. Can this specific resource sensibly be integrated within the lesson instead *without* bloating it (we don't want to assign multiple resources that just talk about mostly the same thing in different ways)? Does it justify replacing one of the assignments? That is our priority for new resources right now, not adding them back to the additional resources section we're removing.


---
TLDR: I'm happy to review this as a proposal to integrate the resource in the lesson/assignment, but not as an additional resource, as we're removing them unless it's clear the resource *cannot* sensibly be integrated into the lesson yet somehow is still important enough that it specifically should get a special mention (and that criteria also prevents the lesson from clogging up with "I also found this resource helpful" in the future, which is exactly where we ended up over so many years)

rgb-katana:
> TLDR: I'm happy to review this as a proposal to integrate the resource in the lesson/assignment, but not as an additional resource, as we're removing them unless it's clear the resource _cannot_ sensibly be integrated into the lesson yet somehow is still important enough that it specifically should get a special mention (and that criteria also prevents the lesson from clogging up with "I also found this resource helpful" in the future, which is exactly where we ended up over so many years)

Thanks for the reply, let me just close the PR

PR #30848, opened 2026-02-26
docs: clarify Step 2 wording from "Gameboard object" to "Gameboard fa…

HemantDahiyaCodes:
…ctory function"

# Confusing wordings in step 2 of the project Tic-tac-toe in full stack javascript path

Link to the project: https://www.theodinproject.com/lessons/node-path-javascript-tic-tac-toe

## Because

The wording "Gameboard object" may confuse learners because they are still getting familiar with the factory functions

The "Gameboard factory function" better reflect what they need to create

## This PR

- Updates the wording in Step 2 from "Gameboard Object" to "Gameboard factory function"
- Improves clarity and consistency with the factory function pattern taught previously
## Issue
N/A

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/project_tic_tac_toe.md
+++ b/javascript/organizing_your_javascript_code/project_tic_tac_toe.md
@@ -7,7 +7,7 @@ We're making a Tic Tac Toe game you can play in your browser!
 <div class="lesson-content__panel" markdown="1">
 
 1. Set up your project with HTML, CSS and JavaScript files and get the Git repo all set up.
-1. You’re going to store the gameboard as an array inside of a Gameboard object, so start there! Your players are also going to be stored in objects, and you’re probably going to want an object to control the flow of the game itself.
+1. You’re going to store the gameboard as an array inside of a Gameboard factory function, so start there! Your players are also going to be stored in objects, and you’re probably going to want an object to control the flow of the game itself.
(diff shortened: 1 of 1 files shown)


goran1010:
I'm not sure about that change - the first substep in the Step 2 clarifies it further:

> Your main goal here is to have as little global code as possible. Try tucking as much as you can inside factories. If you only need a single instance of something (e.g. the gameboard, the displayController etc.) then wrap the factory inside an IIFE (module pattern) so it cannot be reused to create additional instances.

So according to this, gameboard array should probably be within an object, created via factory function inside of an IIFE - but I don't think any of this needs to be mentioned, beyond what is already said.
Learners are told the array should be inside of an object, and are also told to create different objects using IIFEs or factory functions, depending on if they will need only 1 object instance or multiple.

The goal of the project is figuring out how to create those objects, so I'm not sure it would be appropriate to go into details explaining exactly how to create these objects.

Just my 2 cents...

mao-sz:
Thanks for the suggestion. "Object" would make more sense here because the end result is to have an object. That's what other objects would interact with. Factories are a way to create objects, but saying "store inside a Gameboard factory" makes it sound more like the end result is a factory function, but you instead want to use a factory function to make a Gameboard object (which is what the next step clarifies, and also the other object types in that step say: e.g. "player object").

This language is also consistent with previous lessons. Even if you have a constructor or factory function called `Player` and your instances are called `player1` or `player2`, you refer to those instances as "Player objects" because that's what they are.

PR #30864, opened 2026-03-02
Add resource on event capturing and bubbling

tanvirjoy04:
Added a new resource on event capturing vs bubbling.

Purpose: After reading 'Javascript Events' (in assignment), I was not clear about bubbling and capture. But the 'Knowledge Check' section's last question "Explain bubbling vs capture" contains a video that made clearly understand the topic.

Problem Solves/ Benefits: I could be clear about 'bubbling and capture' from the assignment section and could answer the "Explain bubbling vs capture" in Knowledge Section.


mao-sz:
Due to author inactivity, superseding with #30945

PR #30997, opened 2026-04-17
Introduction to RSpec: Fix all top level "describe" blocks to "Rspec.describe"

### LINKED ISSUE #30988: Introduction to RSpec: Describe a class alternative syntax identical to prior syntax
### Checks

### Describe your suggestion

I think there is a typo in the code in the "Basic Syntax" section of this lesson. The following lines are written in the lesson:

> ```
> #spec/calculator_spec.rb
> 
> describe Calculator do
>   describe "#add" do
>     it "returns the sum of two numbers" do
>       calculator = Calculator.new
>       expect(calculator.add(5, 2)).to eql(7)
>     end
>   end
> end
> ```
> ...
> 
> When describing a class, the following syntax is also valid:
> ```
> #spec/calculator_spec.rb
> 
> describe Calculator do
>   #...
> end
> ```
> 

The "also valid" suggests that these two code blocks should have different syntax. But these two blocks appear to have identical syntax.

I think that the second code block, was supposed to say 'context' instead of 'describe' since context is an alias for describe.

### Path

Ruby / Rails

### Lesson Url

https://www.theodinproject.com/lessons/ruby-introduction-to-rspec

### (Optional) Discord Name

_No response_

_No response_

ZackHoang:
## Because

Currently, this lesson is mistakenly introducing the top level `describe` block twices. This PR re-introduces the `Rspec.describe` syntax to be explicit about how RSpec tests should be written.

## This PR

- Fix all top level `describe` blocks to `RSpec.describe`
- Surround a few words in script tags

## Issue

Closes #30988 

## Pull Request Requirements

### DIFF
--- a/ruby/testing_ruby_with_rspec/introduction_to_rspec.md
+++ b/ruby/testing_ruby_with_rspec/introduction_to_rspec.md
@@ -82,17 +82,9 @@ end
 
 Let's go line by line.
 
-<span id='keyword-describe'>First, `Rspec.describe` is an RSpec keyword that defines an "Example Group", or a collection of tests</span>. It takes a class or a string as an argument and is passed a block (`do/end`). `describe` blocks can be nested inside `Rspec.describe`, such as on the second line of our test above. When describing a class, the following syntax is also valid:
+<span id='keyword-describe'>First, `Rspec.describe` is an RSpec keyword that defines an "Example Group", or a collection of tests</span>. It takes a class or a string as an argument and is passed a block (`do/end`). `describe` blocks can be nested inside `Rspec.describe`, such as on the second line of our test above.
 
-```ruby
-#spec/calculator_spec.rb
-
-describe Calculator do
-  #...
-end
-```
-
-While both syntax are valid, stick to `Rspec.describe` for now. Anyway, <span id='keyword-it'>the `it` keyword defines an individual example (aka, test)</span>. `it` takes a string argument and is also passed a block. This block is where our expectations of a method are expressed. In this particular case, when we pass 5 and 2 to the `#add` method, we expect it to return 7. This is concisely expressed in our expectation clause, which uses one of [RSpec's equality matchers](http://rspec.info/features/3-12/rspec-expectations/built-in-matchers/equality/), `eql`:
+<span id='keyword-it'>The `it` keyword defines an individual example (aka, test)</span>. `it` takes a string argument and is also passed a block. This block is where our expectations of a method are expressed. In this particular case, when we pass 5 and 2 to the `#add` method, we expect it to return 7. This is concisely expressed in our expectation clause, which uses one of [RSpec's equality matchers](http://rspec.info/features/3-12/rspec-expectations/built-in-matchers/equality/), `eql`:
 
 ```ruby
   expect(calculator.add(5, 2)).to eql(7)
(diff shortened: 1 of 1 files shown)


ruby/testing_ruby_with_rspec/introduction_to_rspec.md
```diff
@@ -92,16 +92,16 @@ describe Calculator do
 end
 ```
 
-<span id='keyword-it'>The `it` keyword defines an individual example (aka, test)</span>. `it` takes a string argument and is also passed a block. This block is where our expectations of a method are expressed. In this particular case, when we pass 5 and 2 to the `#add` method, we expect it to return 7. This is concisely expressed in our expectation clause, which uses one of [RSpec's equality matchers](http://rspec.info/features/3-12/rspec-expectations/built-in-matchers/equality/), `eql`:
+While both syntax are valid, stick to `Rspec.describe` for now. Anyway, <span id='keyword-it'>the `it` keyword defines an individual example (aka, test)</span>. `it` takes a string argument and is also passed a block. This block is where our expectations of a method are expressed. In this particular case, when we pass 5 and 2 to the `#add` method, we expect it to return 7. This is concisely expressed in our expectation clause, which uses one of [RSpec's equality matchers](http://rspec.info/features/3-12/rspec-expectations/built-in-matchers/equality/), `eql`:
```
KevinMulhern: > While both syntax are valid, stick to `Rspec.describe` for now.

Nit: Because we're recommending that learners don't use it. What do you think about removing the the second example that shows the describe without the RSpec prefix? does it muddy the water and make things harder to understand?

ruby/testing_ruby_with_rspec/introduction_to_rspec.md
```diff
@@ -29,7 +29,7 @@ Finally, `cd` into a project directory that you wish to configure for use with R
 ```bash
 project
   |__lib
-  |   |__script.rb
+  |   |__calculator.rb
```
KevinMulhern, line 32: Nice spot! 💪

ruby/testing_ruby_with_rspec/introduction_to_rspec.md
ZackHoang: @KevinMulhern Now that you mention it, I don't think it does for the following reasons:
1. We did say we want users to use RSpec with its module name without explaining why and some folks might feel strongly to question this, which is a fair concern, and someone will have to explain it on Discord again
2. It really is an either or option, to use or not to use the module name (Hamlet intended yes) would probably not matter too much for the kind of tests learners will be writing within TOP, both on the Ruby and Rails course. I've never worked with Ruby extensively before so I am not sure if there is any point in which learners will have to research about monkey-patching like you mentioned in [this comment](https://github.com/TheOdinProject/curriculum/issues/30988#issuecomment-4260178649) of yours. Anyhow, in the spirit of "conventions over configuration" and trusting learners will be able to research their way out, I think it's okay to remove the second example

KevinMulhern: Those sounds like solid reasons to me! would you like to remove it in this PR?

ZackHoang: @KevinMulhern I'll get that done in a flash!

KevinMulhern (approved):
Thanks for doing this @ZackHoang, this lesson is looking way better 💪

PR #31057, opened 2026-05-07
Testing Routes and Controllers: add a section about resetting the db after each test

### LINKED ISSUE #31027: Testing Routes and Controllers: Add a note about resetting the database after each test
### Checks

### Describe your suggestion

I suggest adding a note that you need to reset the test database after each test. This concerns HTTP methods that act on the database (post, update, patch and delete). For example, if we test a route with a post method and run the test the first time, it would create the record. But, if we run the the test again, it will fail since the record already exists (created in the first test). Thus the need to reset the database state after each test.

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/nodejs-testing-routes-and-controllers

### (Optional) Discord Name

_No response_

_No response_

belmouk:
## Because
It prevents issues when implementing tests on APIs.

## This PR
* removes the last paragraph of the lesson
* adds a section that discusses the importance of resetting dbs after each test
* includes the removed paragraph idea into the new section

## Issue
Closes #31027

## Additional Information
No additional information.

## Pull Request Requirements

### DIFF
--- a/nodeJS/testing_express/testing_database_operations.md
+++ b/nodeJS/testing_express/testing_database_operations.md
@@ -57,27 +57,22 @@ const prisma = new PrismaClient({ adapter });
 
 ### Resetting the database between tests
 
-When testing database operations, we must always isolate our tests. This means, we access the database, interact with it(read, create, update, or delete records) and finally clean it up by resetting it to its initial state.
+When testing database operations, we must always isolate our tests from each other so that no test depends on any other test. For example, we would not want a test to fail if we didn't add or delete a table row in a different test, even if the main behavior being tested actually does work. This means that when we interact with our database in a test (read, create, update, or delete records), we should ideally clean up by resetting the database to its initial state before the next test.
 
-In other words, in:
-
-- `POST` requests, we create new records inside our test, and we delete them at the end.
-- `PUT` requests, we create new records inside our test, update them, check them and then delete them.
-- `DELETE` requests, we create new records inside our test, then delete them.
-
-To avoid repetition and forgetfulness, we recommend using `beforeEach` function, provided by your test runner (Jest or Vitest), to reset your database tables. For example, if we have a database with users and projects:
+This can be conveniently handled in something like a `beforeEach`. For example, if we have a database with users and projects, we could have something like:
 
 ```javascript
-beforeEach(
-  async() => await prisma.$transaction([
-   prisma.user.deleteMany(),
-   prisma.project.deleteMany()
-  ])
+beforeEach(async () => {
+  await prisma.$transaction([
+    prisma.user.deleteMany(),
+    prisma.project.deleteMany(),
+  ]);
+});
 ```
 
 This ensures that the users and projects tables are reset before running any other test.
 
-In addition, if you're splitting your tests into separate files, you need to make sure that test files are run sequentially instead of in parallel to prevent different test database operations from getting mixed up. For Jest, you can add the `--runInBand` flag into your `test` script in `package.json`. If you're using Vitest, you will need to add `fileParallelism : false` to your vitest.config file.
+In addition, if you're splitting your tests into separate files, you need to set up your test runner to run them sequentially (one file after another) instead of in parallel. This prevents different test database operations from getting mixed up. Because Jest defaults to parallel file execution, you can add the [`--runInBand` flag](https://jestjs.io/docs/cli#--runinband) to your `test` script in `package.json` to ensure sequential execution.
 
 Voila, the setup is complete. Now shoo... go get testin'.
 
--- a/nodeJS/testing_express/testing_database_operations.md
+++ b/nodeJS/testing_express/testing_database_operations.md
@@ -55,6 +55,30 @@ const adapter = new PrismaPg({ connectionString });
 const prisma = new PrismaClient({ adapter });
 ```
 
+### Resetting the database between tests
+
+When testing database operations, we must always isolate our tests. This means, we access the database, interact with it(read, create, update, or delete records) and finally clean it up by resetting it to its initial state.
+
+In other words, in:
+
+- `POST` requests, we create new records inside our test, and we delete them at the end.
+- `PUT` requests, we create new records inside our test, update them, check them and then delete them.
+- `DELETE` requests, we create new records inside our test, then delete them.
+
+To avoid repetition and forgetfulness, we recommend using `beforeEach` function, provided by your test runner (Jest or Vitest), to reset your database tables. For example, if we have a database with users and projects:
+
+```javascript
+beforeEach(
+  async() => await prisma.$transaction([
+   prisma.user.deleteMany(),
+   prisma.project.deleteMany()
+  ])
+```
+
+This ensures that the users and projects tables are reset before running any other test.
+
+In addition, if you're splitting your tests into separate files, you need to make sure that test files are run sequentially instead of in parallel to prevent different test database operations from getting mixed up. For Jest, you can add the `--runInBand` flag into your `test` script in `package.json`. If you're using Vitest, you will need to add `fileParallelism : false` to your vitest.config file.
+
 Voila, the setup is complete. Now shoo... go get testin'.
 
 ### Assignment
--- a/nodeJS/testing_express/testing_routes_and_controllers.md
+++ b/nodeJS/testing_express/testing_routes_and_controllers.md
@@ -146,7 +146,23 @@ test("testing route works", done => {
 });
 ```
 
-If we were using a real database here, then we would want to do something similar using either a test or a mock database. We'll talk about setting something like that up in a separate lesson. Suffice it to say for now that you do not want to run test code on your production database!
+### A note about testing routes
+
+In the above example, we have only tested one `POST` request. If we had another `POST` request sending `{ item : "bye"}` before sending the `POST` request `{item : "hey"}`, the above test will fail since the array would be `["bye", "hey"]` instead of `["hey"]`. This makes our tests fragile as the results change without changing our tests!
+
+If we were using a database, the issue will be more pronounced because data is persistent. This means that we would save the same record to our database for as many times as we run the test file. Of course, duplicating records is fine until you get an error for duplicating unique fields (like an email or username).
+
+Thus, it is important that you reset your data storage after each test by deleting any created record using `beforeEach` and `afterEach` methods at the start of your test file. The goal is to isolate each test so that it doesn’t impact others. For example, if we were testing routes of users:
+
+```javascript
+beforeEach(async() => await prisma.user.deleteMany())
+
+afterEach(async() => await prisma.user.deleteMany())
+```
+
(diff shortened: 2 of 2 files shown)


nodeJS/testing_express/testing_routes_and_controllers.md
```diff
@@ -146,7 +146,23 @@ test("testing route works", done => {
 });
 ```
 
-If we were using a real database here, then we would want to do something similar using either a test or a mock database. We'll talk about setting something like that up in a separate lesson. Suffice it to say for now that you do not want to run test code on your production database!
+### A note about testing routes
```
mao-sz: This heading isn't sufficiently descriptive for the section it heads. If I was scanning through the page to look for relevant information, this heading wouldn't help me much. Similarly, the heading doesn't really tell me what to expect from the contents, so navigation-by-heading wouldn't be that helpful either.

Something like "Clearing the database between tests" or "Database data between tests" would be more descriptive.

nodeJS/testing_express/testing_routes_and_controllers.md
```diff
@@ -146,7 +146,23 @@ test("testing route works", done => {
...
+
+```javascript
+beforeEach(async() => await prisma.user.deleteMany())
+
+afterEach(async() => await prisma.user.deleteMany())
+```
+
+Here, we also use `afterEach` to ensure that if you are running multiple test files, data is not leaking between them.
+
+Given the importance of the production database (the source of truth of our application), it is not advisable to modify or query it directly. Instead, we always create a smaller copy, that we call a test database, over which we run our tests without the risk of damaging real production data. We will be talking about this in detail over the next lesson.
```
mao-sz: More just general reviewing of this:

1. This isn't the right lesson for this content to go in. While the point about potential persistant data stands, it's not really the focus of this lesson, which is simply about introducing `supertest` and API route tests in general with some basic contrived examples. Database ops in tests are specifically left for the next lesson (which is what the line you replaced is saying). Mentioning this concept is something that should go in [Testing Database Operations](https://www.theodinproject.com/lessons/node-path-nodejs-testing-database-operations) before the assignment, and would also mean we don't need the last paragraph ("Given the importance of...").
2. Got a source on `beforeEach` **and** `afterEach`? My understanding of Jest and typical JS test runners is that they run all tests in file sequentially, and only run multiple test files in parallel. Having both `beforeEach` and `afterEach` wouldn't help with multiple files in parallel because one file could truncate the DB when a test in another file is halfway through (imagine a route that updated an existing user but between adding the user at the start of the test and updating it, the database got cleared by a different test file).

   In this case, the simplest thing within TOP's scope of projects **if** multiple test files are used (that all hit the database) would be [running tests sequentially](https://jestjs.io/docs/cli#--runinband). A more advanced option could be programmatically creating new test DBs per test file, so all test files are fully isolated. I'm not sure we need to mention that specifically in our lessons though.

   As a result, only `beforeEach` is required.

mao-sz (changes requested):

nodeJS/testing_express/testing_routes_and_controllers.md
belmouk: Thank you for your feedback. I have changed the title accordingly.

nodeJS/testing_express/testing_routes_and_controllers.md
belmouk: Great catch. I have moved the changes to the "Testing Database Operations" lesson and have rewritten it to include your input.

nodeJS/testing_express/testing_database_operations.md
```diff
@@ -55,6 +55,30 @@ const adapter = new PrismaPg({ connectionString });
 const prisma = new PrismaClient({ adapter });
 ```
 
+### Resetting the database between tests
+
+When testing database operations, we must always isolate our tests. This means, we access the database, interact with it(read, create, update, or delete records) and finally clean it up by resetting it to its initial state.
```
mao-sz: What do you think of the following? Just expanding on *why* we isolate tests regarding DB state, and some slight grammar tweaks.

```suggestion
When testing database operations, we must always isolate our tests from each other so that no test depends on any other test. For example, we would not want a test to fail if we didn't add or delete a table row in a different test, even if the main behavior being tested actually does work. This means that when we interact with our database in a test (read, create, update, or delete records), we should ideally clean up by resetting it to its initial state before the next test.
```

nodeJS/testing_express/testing_database_operations.md
```diff
@@ -55,6 +55,30 @@ const adapter = new PrismaPg({ connectionString });
...
+
+In other words, in:
+
+- `POST` requests, we create new records inside our test, and we delete them at the end.
+- `PUT` requests, we create new records inside our test, update them, check them and then delete them.
+- `DELETE` requests, we create new records inside our test, then delete them.
```
mao-sz: I don't think this bit is particularly necessary. Learners should know what POST/PUT/DELETE routes would do at this point, so they should be able to work out what we mean by resetting DB state between tests for each situation.

nodeJS/testing_express/testing_database_operations.md
```diff
@@ -55,6 +55,30 @@ const adapter = new PrismaPg({ connectionString });
...
+
+To avoid repetition and forgetfulness, we recommend using `beforeEach` function, provided by your test runner (Jest or Vitest), to reset your database tables. For example, if we have a database with users and projects:
```
mao-sz: Just streamlining a bit. Jest is implied in both the previous lesson and this one, so no need to mention different test runners (anyone opting to use something else can surely work things out at this point)
```suggestion
This can be conveniently handled in something like a `beforeEach`. For example, if we have a database with users and projects, we could have something like:
```

nodeJS/testing_express/testing_database_operations.md
```diff
@@ -55,6 +55,30 @@ const adapter = new PrismaPg({ connectionString });
...
+
+```javascript
+beforeEach(
+  async() => await prisma.$transaction([
+   prisma.user.deleteMany(),
+   prisma.project.deleteMany()
+  ])
+```
```
mao-sz, line 71: Indentation and missing closing braces, but also a formatting nit just a little more consistent with how snippets in other lessons tend to be formatted.
````suggestion
```javascript
beforeEach(async () => {
  await prisma.$transaction([
    prisma.user.deleteMany(),
    prisma.project.deleteMany(),
  ]);
});
```
````

File: nodeJS/testing_express/testing_database_operations.md [hunk 7]
```diff
@@ -55,6 +55,30 @@ const adapter = new PrismaPg({ connectionString });
...
+
+This ensures that the users and projects tables are reset before running any other test.
+
+In addition, if you're splitting your tests into separate files, you need to make sure that test files are run sequentially instead of in parallel to prevent different test database operations from getting mixed up. For Jest, you can add the `--runInBand` flag into your `test` script in `package.json`. If you're using Vitest, you will need to add `fileParallelism : false` to your vitest.config file.
```
mao-sz: Just a little wording tweak but also adding a doc link, and removing the Vitest bit (similar to earlier comment - if someone's using Vitest, they can look into this themselves). What do you think of the following?
```suggestion
In addition, if you're splitting your tests into separate files, test runners tend to run test files in parallel (with tests in each file run sequentially), so you'll want to ensure they're all run sequentially so database ops in one test file don't interfere with any in another file. For Jest, you can add the [`--runInBand` flag](https://jestjs.io/docs/cli#--runinband) to your `test` script in `package.json`.
```

mao-sz (changes requested):
Thanks for moving this over. Mainly some wording suggestions below

nodeJS/testing_express/testing_database_operations.md
belmouk: I have reformulated the paragraph with your inputs in mind to make it clearer and easy to digest. Let me know if you have any feedbacks.
```suggestion
When testing database operations, we must always isolate our tests from each other so that no test depends on another. This prevents a passing test from failing due to database changes made by a test run. For example, if a test creates a record in the database without cleaning it up, running that same test again will fail because the record already exists. Therefore, every database interaction (reading, creating, updating, or deleting records) must be cleaned up to reset the database to its initial state before the next test runs.
```

nodeJS/testing_express/testing_database_operations.md
belmouk: I have added this paragraph to explain how to test each HTTP method. I wanted to highlight the setup, assert, clean up flow for such tests. But I don't mind removing it.

nodeJS/testing_express/testing_database_operations.md
belmouk: Aligned. Will add this change to the final commit.

nodeJS/testing_express/testing_database_operations.md
belmouk: Thanks, I will add this to the final commit.

nodeJS/testing_express/testing_database_operations.md
belmouk: I have restructured your proposal. Let me know what you think. 
```suggestion
In addition, if you're splitting your tests into separate files, you need to set up your test runner to run them sequentially (one file after another) instead of in parallel. This prevents different test database operations from getting mixed up. Because Jest defaults to parallel file execution, you can add the [`--runInBand` flag](https://jestjs.io/docs/cli#--runinband) to your `test` script in `package.json` to ensure sequential execution. 
```

belmouk:
Thank you @mao-sz for your review and suggestions. I have responded back to each of your comments, I'll wait for your feedback to submit one final commit implementing what we agree on.

nodeJS/testing_express/testing_database_operations.md
mao-sz: I dig it :rocket:

nodeJS/testing_express/testing_database_operations.md
mao-sz: I think at this point I'm the curriculum, accounting for the lesson flow, it feels unnecessary and a little over specific. Not that it's wrong though.

nodeJS/testing_express/testing_database_operations.md
mao-sz: I'm not so sure on the new suggestion, mainly because it starts by saying tests should not depend on other tests but then expands on what can happen even if you only have a single test and just re-run it. Not wrong but confusing in terms of content. We won't need to mention every possible thing that can happen, just provide one example, so in this situation I think I prefer my suggestion over this new one.

belmouk: I have added your version to the last commit.

nodeJS/testing_express/testing_database_operations.md
belmouk: Noted. I have removed this part in the last commit.

mao-sz (approved):
🚀

PR #31070, opened 2026-05-11
Factory Functions and the Module Pattern lesson: Enhance documentation on object destructuring usage

surrelsaga:
## Because
In the `createPlayer` example of the lesson, I found the object destructuring part: "const { getReputation, giveReputation } = createUser(name);" confusing. I initially assumed it behaved like array destructuring which relies on the order rather then relaying on exact property name and it took me a long while to figure this out. I think many other people will also find this confusing, so clarifying this distinction immediately after the code block will help prevent this confusion.

## This PR
- Add a brief explanation clarifying that object destructuring extracts values based on exact property name.
- Add an internal markdone link pointing back to the `### Destructuring` section earlier in this lesson so everyone can go back and review it.

## Issue
N/A

## Additional Information
N/A

## Pull Request Requirements

### DIFF
--- a/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
+++ b/javascript/organizing_your_javascript_code/factory_functions_and_module_pattern.md
@@ -240,7 +240,7 @@ function createPlayer(name, level) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mao-sz:
Thanks for the suggestion. Personally, I'm not convinced adding this is necessary. Individual misunderstandings or misremembering things will always occur and that's of course perfectly fine and expected as you learn. But from a neutral perspective, this addition seems a little out of place and reads with a bit of a "why does this need to be mentioned specifically?" feel.

So I'm going to close this because while it isn't wrong info, we already introduce the syntax and people can always look up docs and experiment independently, same as with any concept or syntax introduced at any point in any other lesson.

PR #31156, opened 2026-06-10
Transforms: Change composite transforms to left-to-right order

mao-sz:
## Because

Composite transform order has long confused learners because the original wording of the MDN Transform docs was
> The transform functions are multiplied in order from left to right, meaning that composite transforms are effectively applied in order from right to left.

This right-to-left order was backed by other doc pages (like for `animation-composition`).

Reading left-to-right, each transform establishes a new coordinate space as a result. Technically, it can be read right-to-left which requires retaining the parent coordinate space across the transforms. According to the CSS spec, both orders are viable depending on which exact part of the underlying process you're dealing with. The MDN docs were confusing because they talk about and switch between both perspectives without saying or explaining, so things seem almost contradictory.

When https://github.com/mdn/content/pull/43496 is merged and the changes are live (~~will undraft this PR once it is~~ now merged), the Transform docs will be clearer about the two orders - left-to-right is, for all practical purposes, the order to use. This is also way more intuitive just in general. Since our lesson currently describes the right-to-left order, we'll need to change it once this goes live. No CodePen changes should be necessary.

## This PR

- Amends composite transform example description to reference left-to-right application order.

## Issue

N/A

## Pull Request Requirements

### DIFF
--- a/advanced_html_css/animation/transforms.md
+++ b/advanced_html_css/animation/transforms.md
@@ -157,9 +157,7 @@ on <a href="https://codepen.io">CodePen</a>.</span>
 
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-If you guessed correctly, congratulations! But this is a tricky concept. MDN's transform docs state that "[composite transforms are effectively applied in order from right to left](https://developer.mozilla.org/en-US/docs/Web/CSS/transform#values)".
-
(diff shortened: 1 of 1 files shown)


mao-sz:
When approved, will hold off on merging until MDN changes are live.

zachmmeyer (approved):

PR #31181, opened 2026-06-20
Object Constructors: Fix numbering and add new resource on Prototypal Inheritance

Ratneshtripathi07:
### Because

The existing resources on prototypal inheritance were useful, but I found an additional article that explains the concept visually and in a more intuitive way. Since prototypal inheritance can be confusing for beginners, I wanted to add another high-quality reference that could improve understanding.

### This PR

- Updated the existing JavaScript.info prototype inheritance article link to the newer and secure HTTPS version and fixed it's numbering.
- Added an additional resource explaining prototypal inheritance visually through examples
- Kept the added resources relevant to the lesson’s objective of understanding prototypes and inheritance in JavaScript

### Issue

N/A

### Additional Information

The added article helped explain prototype chaining and inheritance in a more beginner-friendly way, so I felt it could provide extra learning value for students working through this lesson.


mao-sz (commented):
Thanks for the PR. 2 things:

1. As per our layout style guide, we use [lazy numbering](https://github.com/TheOdinProject/curriculum/blob/main/LAYOUT_STYLE_GUIDE.md#lazy-numbering) for ordered lists, so the `1. 1. 1.` is intentional.
1. I would normally ask for more context about what gap exists in the lesson's pedagogy that would be addressed by this resource, but right now most of the images do not load for me - simply showing "DEV image no longer exists". I'll give it a few days but if by then it's still with broken images, I'll close the PR as we won't want to add that as an assignment then.

Ratneshtripathi07:
Thanks for reviewing the PR.
I wasn’t aware that lazy numbering was part of the project’s layout style guide, so I’ll revert that change.

Regarding the added resource, my reasoning was that while the lesson explains prototypes and prototypal inheritance conceptually, I felt learners may benefit from an additional visual explanation of prototype chaining and how constructor functions, prototype objects, and inheritance relationships interact, since these concepts can feel quite abstract when first encountered.
As for the DEV article issue, I checked the article again from my side and the images are currently loading correctly for me, so it may have been a temporary issue or an intermittent CDN problem. I’ll keep an eye on it, and I understand the concern if the resource proves unreliable over time.

Screenshot from 2026-06-21 03-32-54
Screenshot from 2026-06-21 03-33-16

mao-sz:
I'm UK-based and on multiple networks/devices, most of the images are still unavailable. Another maintainer (US-based) has just confirmed the same on their end.

Given that impacts a very sizeable portion of TOP users, I'm going to reject this specific resource and close this PR.

If you do come across a different resource that you believe fills the same gap sufficiently, you're of course welcome to open a new PR with the details for that. Doesn't necessarily mean it'll be accepted, but we can certainly review it, unlike this one unfortunately.

PR #31244, opened 2026-07-21
Clarify flex-shrink example in documentation

0xSubhan:
## Because

The current explanation assumes readers already understand that shrinking the browser window also reduces the width of the flex container. For beginners, it may not be obvious that the flex container has no explicit width and, as a block-level element, automatically expands or contracts with the available width of its parent. Adding this explanation helps readers understand why resizing the browser causes the flex items to grow or shrink.

## This PR

- Clarifies that the `.flex-container` has no explicit width and automatically fills the available width of its parent.
- Explains that shrinking the browser window also shrinks the flex container.
- Makes it clearer why the flex items shrink when the available space decreases, while `.two` remains at `250px` because of `flex-shrink: 0`.

## Issue

N/A

## Additional Information

This change is intended to improve the learning experience for beginners by explicitly connecting browser resizing with the flex container's changing width. The original explanation assumes this behavior is already understood, which may cause confusion for readers encountering Flexbox for the first time.

## Pull Request Requirements

- [x] I have thoroughly read and understand [The Odin Project curriculum contributing guide](https://github.com/TheOdinProject/curriculum/blob/main/CONTRIBUTING.md)
- [x] The title of this PR follows the `location of change: brief description of change` format
- [x] The `Because` section summarizes the reason for this PR
- [x] The `This PR` section has a bullet point list describing the changes
- [x] If this PR addresses an open issue, it is linked in the `Issue` section
- [x] If any lesson files are included in this PR, they have been previewed with the Markdown preview tool to ensure they are formatted correctly
- [x] If any lesson files are included in this PR, they follow the Layout Style Guide

### DIFF
--- a/foundations/html_css/flexbox/flexbox_growing_and_shrinking.md
+++ b/foundations/html_css/flexbox/flexbox_growing_and_shrinking.md
@@ -44,7 +44,7 @@ In the following example the `flex` shorthand has values for `flex-shrink` and `
 
 The default shrink factor is `flex-shrink: 1`, which means all items will shrink evenly. If you do *not* want an item to shrink then you can specify `flex-shrink: 0;`. You can also specify higher numbers to make certain items shrink at a higher rate than normal.
 
-Here's an example. Note that we've also changed the `flex-basis` for reasons that will be explained shortly. If you shrink your browser window you'll notice that `.two` never gets smaller than the given width of 250px, even though the `flex-grow` rule would otherwise specify that each element should be equally sized.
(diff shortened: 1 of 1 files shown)


mao-sz:
Thanks for the suggestion. I'm going to close this for a couple of reasons, however:
- This is not a confusion I've ever seen occur in the community.
- Users will have been exposed to block behaviour already.
- With the current lesson text, it already sufficiently explains what will happen when the user does a certain thing, so all the user has to do is play with their browser size where they can see things happening. I don't see much room for confusion over potentially thinking the container *should not* automatically size by the available space, to the point of not being able to understand the point of that section.

PR #31264, opened 2026-07-31
Form Validation lesson: Update Styling Validations paragraph

### LINKED ISSUE #31239: Form Validation: weird mistakes
### Checks

### Describe your suggestion

in "[Styling validations](https://www.theodinproject.com/lessons/node-path-intermediate-html-and-css-form-validation#styling-validations)" section, there is this paragraph: "First, we target valid inputs with a green border. Initially, our email and URL fields maintain their default browser styling, representing a neutral state because the user hasn’t interacted with them yet." But if you look at the code, the :user-invalid actually targeted first, so i dont know why the editor would say "First, we target valid inputs with a green border".
just after that paragraph, there is another one: "Once a user enters an invalid value, the border switches to red. You can test this by entering an incorrectly formatted email or URL to see the validation in action."
but "once" i typed something in the email box or url box, it did not turn red unless i clicked somewhere else on the page or unless i clicked the submit button, so i dont know why the editor would say "once".

### Path

Node / JS

### Lesson Url

https://www.theodinproject.com/lessons/node-path-intermediate-html-and-css-form-validation

### (Optional) Discord Name

_No response_

_No response_

ToastedGMS:
## Because
Improve clarity on the Styling Validation paragraphs

## This PR
- Minor clean ups to the order of the text, and clarification on the new expected behavior based on the current usage of `:user-valid` instead of the old `:valid`

## Issue

Closes [#31239](https://github.com/TheOdinProject/curriculum/issues/31239)

## Additional Information
I decided to change the order on the first paragraph to mention the initial state of the inputs first, followed by the valid state and then the invalid state. I feel like that makes it a bit easier to follow since it matches the visual feedback you see (the input isn't valid with a green border before you interact with it), while also matching the order of the CSS as suggested by [tshamsiddin](https://github.com/tshamsiddin).

I also noticed that the input fields remain validated with a green border even if you delete all the text. For example,  if you type in an invalid value, and click away the input gets invalidated (red border). But if you then delete the invalid value and leave it blank, the input then gets validated (green border). That might be worth checking out in the codepen.

Lastly, I wonder if it might be worth it to explain shortly the difference between `:valid :invalid` and `:user-valid :user-invalid`? Since we take the time to explain that the user is expected to click away from the input to trigger it's validation, this might leave the reader wondering why.

## Pull Request Requirements

### DIFF
--- a/intermediate_html_css/forms/form_validations.md
+++ b/intermediate_html_css/forms/form_validations.md
@@ -198,9 +198,10 @@ To see this in action, we will be using our email and website example that we lo
 
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
-First, we target valid inputs with a green border. Initially, our email and URL fields maintain their default browser styling, representing a neutral state because the user hasn't interacted with them yet.
+Initially, our email and URL fields maintain their default browser styling, representing a neutral state because the user hasn't interacted with them yet. We start by targetting valid inputs, setting a green border on the respective input field.
 
-Once a user enters an invalid value, the border switches to red. You can test this by entering an incorrectly formatted email or URL to see the validation in action.
+Once a user enters an invalid value and clicks or tabs away from the input (which would constitute a complete user interaction), the border switches to red. From then on, the border will automatically change between green and red depending on the validity of its contents. Test this out by entering an incorrectly formatted email or URL, clicking away to engage the validation, then editing the value between correctly formatted and incorrectly formatted.
+
 
 ### Conclusion
 
--- a/intermediate_html_css/forms/form_validations.md
+++ b/intermediate_html_css/forms/form_validations.md
@@ -198,10 +198,9 @@ To see this in action, we will be using our email and website example that we lo
 
 <script async src="https://cpwebassets.codepen.io/assets/embed/ei.js"></script>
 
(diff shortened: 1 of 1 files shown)


ToastedGMS:
For the pseudo-class explanation, I was thinking for example of something like:

>You might notice some form input fields execute validation as you type (getting a red border until the value matches the expected validation). This is due to the usage of `:valid/:invalid`. When using these CSS pseudo-classes, the form validation is triggered the moment the page loads, which doesn't require interaction, but can lead to poor UX, by displaying error colors prematurely. It might be preferred to use `:user-valid/:user-invalid` instead, which waits for a completed user interaction (such as typing and leaving the form or clicking away) before attempting to validate.

mao-sz (approved):
Thanks for this. I agree with the first paragraph flipping - does flow a little more intuitively.

Re: valid/invalid, we don't need to include stuff about that. If anyone goes to the `user-*` docs, there will be links to and info about comparison with valid/invalid already. Our lessons are to expose ideas and bring awareness to things but we should be wary about trying to become documentation.

Re: empty fields, non-required inputs are always valid when empty. This is something that's easily found via googling if someone sees that behaviour and goes "huh", so I'm not concerned about it in the context of this lesson. When I'm able to, I might see if the codepen could do with `required` attributes too but I'm not too worried about them not having them really (they've never had them and I don't think I've ever seen this be a problem before, at least not one that someone can't resolve after a quick google).

PR #31274, opened 2026-08-04
markdownlint test utils: Fix test execution on Windows

Abhishek371681:
## Because
Running the custom markdownlint rule tests on Windows environments previously failed due to two issues:
1. `spawnSync` calls `npm` directly without a shell, which throws `ENOENT` on Windows as `npm` is a batch script (`npm.cmd`).
2. `markdownlint-cli2` returns lint outputs using forward slash (`/`) path separators for all platforms, causing assertion mismatches with expected paths constructed via `path.join()`, which resolves to backslashes (`\\`) on Windows.
These fixes make the test suite fully compatible with Windows.

## This PR
* Modified `markdownlint/test_utils/fix.js` to run `spawnSync` with `{ shell: true }` so that `npm` resolves correctly on Windows.
* Modified `markdownlint/test_utils/lint.js` to normalize the forward slash path separators in linter error reports to platform-specific path separators on Windows, aligning them with the expected outputs constructed by tests.

## Issue

Closes #XXXXX

## Additional Information
Tests were verified locally on Windows, with all 67 tests successfully passing after these changes.

## Pull Request Requirements

### DIFF
--- a/foundations/installations/setting_up_git.md
+++ b/foundations/installations/setting_up_git.md
@@ -96,7 +96,7 @@ This section contains a general overview of topics that you will learn in this l
 
    #### Multiple SSH keys
 
-   If you have already setup an ssh key pair with GitHub on a different machine prior to starting The Odin Project, GitHub allows you to have multiple key pairs associated with your account. You can just follow these instructions again to set up another key pair and register it with GitHub.
+   If you have already setup an SSH key pair with GitHub on a different machine prior to starting The Odin Project, GitHub allows you to have multiple key pairs associated with your account. You can just follow these instructions again to set up another key pair and register it with GitHub.
 
    </div>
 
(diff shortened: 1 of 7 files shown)


mao-sz:
Thanks for opening this but 2 things:

1. You created this branch off a different feature branch that already had commits in it, so those unrelated commits have contaminated this branch and PR. When you work on a new feature, you should branch off whatever would be your target branch, so in this case, you checkout back to `main` (because that's your target), pull from upstream main to ensure you've got the latest `main` changes, *then* create your new feature branch off `main`. That keeps your different features isolated and no unrelated commits contaminating.
2. Had a chat with others on the team and we agree that Windows compatibility is not something we need to implement.
   The team all use (or have access to) Linux or macOS, and many don't do development on Windows. Almost every contributor here will also be using Linux or macOS; the majority of contributors here making meaningful contributions will be those following or having followed the curriculum, and Windows is not supported as part of it (VirtualBox, dual-booting and WSL are all supported options for running Linux as a Windows user).
   Therefore, there isn't a lot lost by not making our internal tooling Windows-compatible, but there is a large maintenance burden introduced if we do make it Windows-compatible, because we'd have to ensure it stays compatible going forward. That's not something the team feels is necessary in context.

Therefore, I'll be closing this PR.

PR #31276, opened 2026-08-05
Warn if npm scripts ran with non-linux/macOS

mao-sz:
## Because

Our internal tooling is not made to ensure Windows compatibility and that isn't something we're interesting in ensuring due to lack of benefit and increased maintenance burden.
Currently, the only thing affected by this is our custom lint rules' test suites, though this may extend to other things in the future. The linter itself still works with Windows.

Providing a compatibility warning should be sufficient documentation of OS support - will not block npm package installation or scripts (so people can still use the linter) but should no longer be a surprise if other things (like the custom rule tests) don't work.

## This PR

- Adds npm warnings if used with OSes other than Linux or macOS (available platform names as per [Node `process.platform` docs](https://nodejs.org/docs/latest-v24.x/api/process.html#processplatform))

## Issue

N/A

## Additional Information

For QA:

With Linux
image

With windows
image

## Pull Request Requirements

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,5 +10,17 @@
   "license": "CC BY-NC-SA 4.0",
   "devDependencies": {
     "markdownlint-cli2": "^0.20.0"
+  },
+  "devEngines": {
+    "os": [
+      {
+        "name": "linux",
+        "onFail": "warn"
+      },
+      {
+        "name": "darwin",
+        "onFail": "warn"
+      }
+    ]
   }
 }


mao-sz:
@bycdiaz tagged you for review since I know you use macOS. Would you be able to checkout to this PR and test that `npm i`/`npm install` *does not* show any `devEngines.os` warning? i.e. should just work fine like it does for me in Linux

bycdiaz:
Didn't see any `devEngines.os` warnings.

I ran this through Claude and it said there could be Windows users on older versions of Node (18, 20) may not get the warning. But I imagine that's not a huge deal here? I think we're good but figured I'd relay in case you feel that's useful.

bycdiaz (approved):
This is so luxurious. ✨

mao-sz:
> Didn't see any `devEngines.os` warnings.
> 
> I ran this through Claude and it said there could be Windows users on older versions of Node (18, 20) may not get the warning. But I imagine that's not a huge deal here? I think we're good but figured I'd relay in case you feel that's useful.

`devEngines` is only available in npm v10.9.0+, so Node v20 or lower will not ship with that by default and will therefore not run the checks. But I'm not concerned about that as it'll be a very small percentage of people hit by that and they'd likely have other compatibility issues (or can easily install a later Node and npm version).

PR #31278, opened 2026-08-05
docs: add note about Vite linter selection

rey-1e:
## Because

Recent versions of Vite prompt users to choose a linter during project creation, but the lesson does not mention this. This can be confusing for learners creating their first React application.

## This PR

- Adds a lesson note explaining what a linter is.
- Mentions that recent versions of Vite may prompt users to choose between `ESLint` and `Oxlint`.
- Recommends `ESLint` for learners who are unsure which option to choose.

## Issue

N/A

## Additional Information

N/A

### DIFF
--- a/react/introduction/setting_up_a_react_environment.md
+++ b/react/introduction/setting_up_a_react_environment.md
@@ -59,6 +59,14 @@ npm create vite@latest my-first-react-app -- --template react
 
 You may be asked if you want to install the `create-vite` package, in which case you should accept by typing `y` then hitting <kbd>Enter</kbd>. If it asks you any questions about using experimental features, you can answer no to them. Finally, answer yes to `Install with npm and start now?`.
 
+<div class="lesson-note" markdown=1>
+
+#### Choosing a linter
+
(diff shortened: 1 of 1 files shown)


ZackHoang:
Not sure if we need to specify this as in the JavaScript course learners are already exposed to linters and we already introduced them to ESlint specifically. At this point, I doubt learners need to be told which linter to use because if they are so inclined to use Oxlint, documentation is always there if needed.

rey-1e:
> Not sure if we need to specify this as in the JavaScript course learners are already exposed to linters and we already introduced them to ESlint specifically. At this point, I doubt learners need to be told which linter to use because if they are so inclined to use Oxlint, documentation is always there if needed.

yup I agree, sounds totally valid to me.

mao-sz:
Thanks for the PR but closing this because I would like to think learners at this point in the curriculum can navigate that question themselves, and won't be frozen in place unless we tell them what to do.

rey-1e:
> Thanks for the PR but closing this because I would like to think learners at this point in the curriculum can navigate that question themselves, and won't be frozen in place unless we tell them what to do.

yupp

PR #31375, opened 2026-09-08
Project HashMap: Tweak specs for upcoming testing instructions

### LINKED ISSUE #30513: Audit the project requirements in the computer science section to make sure they're fit for having learners write tests while working through them.
Going to work through the projects in both paths with tests and take notes on different problems that I think should be addressed to make the projects more test friendly.

### TODO

Will probably be more here as I work through some things.

### LINKED ISSUE #31374: Project HashMap: Tweak specs for upcoming testing instructions
Some general phrasing and grammar tweaks needed, along with the odd return value tweak (e.g. JS `get()` should return `undefined` rather than `null` - more consistent with JS-like behaviour such as `Map.prototype.get()`).

mao-sz:
## Because
Some general phrasing and grammar tweaks needed, along with the odd return value tweak (e.g. JS `get()` should return `undefined` rather than `null` on missing key - more idiomatic for JS).

## This PR

- Tweaks phrasing/grammar for consistency and general flow.
- Changes JS' `get()` return value for missing keys to `undefined`

## Issue

Closes #31374
Closes #30513


goran1010 (approved):
🚀

